PR #10, opened 2015-12-07
Make ReviewClasses less confusing.

stasszczesniak:
Previous version of CritiqueBrainz used review classes for various features within the website. The problem is that these classes are not very useful and confusing. Now review_classes from critiquebrainz/data/constants.py and its' usages are removed.
My task on GCI : https://codein.withgoogle.com/dashboard/task-instances/5507485452468224/

Also git was missing in scripts/bootstrap.sh


gentlecat:
LGTM. :+1: Thank you!

PR #11, opened 2015-12-12
Updated nullable for is_blocked

adeshp96:
Set nullable for is_blocked to True so that the existing dump works.


gentlecat:
It's shouldn't be nullable. Instead, `is_blocked` should be included into data dumps. See https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/data/export_manager.py#L250.

adeshp96:
Yup okay got it

PR #14, opened 2015-12-13
Added note for drafts not being private and fixed ISE on reviewing event without place

adeshp96:
Added the note and used "fab extract_strings" to create the portable object template file(messages.pot)
Also fixed bug reported at http://tickets.musicbrainz.org/browse/CB-194

### DIFF
--- a/critiquebrainz/frontend/templates/review/modify/event.html
+++ b/critiquebrainz/frontend/templates/review/modify/event.html
@@ -11,7 +11,11 @@
       {% endif %}
     </dd>
     <dt>{{ _('Place') }}</dt>
-    <dd>{{ entity['place-relation-list'][0]['place']['name'] or '-' }}</dd>
+    {% if entity['place-relation-list'] is not defined %}
+      <dd>{{ _('Unknown place') }}</dd>
+    {% else %}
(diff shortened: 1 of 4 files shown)


gentlecat:
It might be better to extract strings separately from pull requests. First of all, it prevents conflicts by doing string updates only on *master* branch. Second, it keeps diff in the pull request cleaner and easier to review.

Don't worry about it now. Just something to keep in mind for the future. Also, good to see that you think about localization when making changes like this one. :+1:

critiquebrainz/frontend/templates/review/modify/event.html
```diff
@@ -11,7 +11,11 @@
       {% endif %}
     </dd>
     <dt>{{ _('Place') }}</dt>
-    <dd>{{ entity['place-relation-list'][0]['place']['name'] or '-' }}</dd>
+    {% if entity['place-relation-list'] is not defined %}
+      <dd>{{ _('Unknown place') }}</dd>
```
gentlecat: Can you make this stand out from styling of regular name somehow? Maybe add `text-muted` class?

adeshp96: Done.

PR #15, opened 2015-12-14
Review only search implemented

adeshp96:
I have implemented the reviews only search for release groups for now. Tested too. Doesn't works slow on my PC. A checkbox is also implemented for this. If you check the checkbox for artist or event nothing changes. Whereas for release groups it will filter for results with reviews only. Try it once. I don't think it is too slow. Also, I think it's a crucial feature because, if someone searches on critiquebrainz they want critiques too show up too and not just tracks(in case of release groups), so the user can simply filter the ones with reviews. For those who think otherwise, the checkbox will help :D


gentlecat:
Same comment about strings from another PR applies here as well. See https://github.com/metabrainz/critiquebrainz/pull/14#issuecomment-164618006.

critiquebrainz/frontend/templates/review/modify/base.html
```diff
@@ -56,6 +56,7 @@
     {% block license_agreement_input %} {% endblock %}
   </form>
   {% block buttons %} {% endblock %}
+  <br /><small class="text-warning"><em>{{ _('Note: Drafts are not private!') }}</em></small>
```
gentlecat: Is this from another pull request? You should keep each in separate branches based on *master*.

adeshp96: yeah i realized this problem a bit late :/  so should I submit another pull request? I can do this for the other pull requests too, perhaps making everything look a lot more cleaner? Either way, I will note this in future. Thank you

adeshp96:
I have implemented this feature for search for any search query. Please let me know if there is any problem.

adeshp96:
Merge conflict resolved.

critiquebrainz/frontend/views/search.py
```diff
@@ -1,13 +1,15 @@
...
 from critiquebrainz.frontend.external import musicbrainz
 
 search_bp = Blueprint('search', __name__)
 
 RESULTS_LIMIT = 10
 
 
-def search_wrapper(query, type, offset=None):
-    if query:
+def search_wrapper(query, type, offset=None,review_only=False):
```
gentlecat: There should be a space after `,`.

critiquebrainz/frontend/views/search.py
```diff
@@ -16,17 +18,49 @@ def search_wrapper(query, type, offset=None):
...
+        elif type == "event":
+            count, results = musicbrainz.search_events(query)
+        elif type == "release-group":
+            count, results = musicbrainz.search_release_groups(query)
+        else:
+            count, results = 0, []
     else:
         count, results = 0, []
+    if review_only is True :
+        if type== "artist":
```
gentlecat: Space before `==`.

critiquebrainz/frontend/views/search.py
```diff
@@ -16,17 +18,49 @@ def search_wrapper(query, type, offset=None):
...
+            count, results = musicbrainz.search_artists(query)
+        elif type == "event":
+            count, results = musicbrainz.search_events(query)
+        elif type == "release-group":
+            count, results = musicbrainz.search_release_groups(query)
+        else:
+            count, results = 0, []
     else:
         count, results = 0, []
+    if review_only is True :
```
gentlecat: No space before `:`.

critiquebrainz/frontend/views/search.py
```diff
@@ -16,17 +18,49 @@ def search_wrapper(query, type, offset=None):
...
+        if type== "artist":
+             fresults=[]
```
gentlecat: Spaces around `=`.

critiquebrainz/frontend/views/search.py
```diff
@@ -16,17 +18,49 @@ def search_wrapper(query, type, offset=None):
...
+             for group in results:
+                 count, release_groups = musicbrainz.browse_release_groups(artist_id=group['id'], release_types=["album"])
+                 for release in release_groups:
+                     if(Review.list(entity_id=release['id'])[0]):
```
gentlecat: You don't need to use parentheses with `if` statement.

gentlecat:
Please clean up your code.

gentlecat:
ok to test

critiquebrainz/frontend/views/search.py
```diff
@@ -18,17 +19,50 @@ def search_wrapper(query, type, offset=None):
...
 
 
 @search_bp.route('/')
 def index():
     query = request.args.get('query')
     type = request.args.get('type')
-    count, results = search_wrapper(query, type)
-    return render_template('search/index.html', query=query, type=type, results=results, count=count, limit=RESULTS_LIMIT)
+    if(request.args.get('review-only')=="on"):
+        review_only=True;
```
gentlecat: You don't need to put a semicolon after statements in Python.

critiquebrainz/frontend/views/search.py
```diff
@@ -18,17 +19,50 @@ def search_wrapper(query, type, offset=None):
...
     return count, results
 
 
 @search_bp.route('/')
 def index():
     query = request.args.get('query')
     type = request.args.get('type')
-    count, results = search_wrapper(query, type)
-    return render_template('search/index.html', query=query, type=type, results=results, count=count, limit=RESULTS_LIMIT)
+    if(request.args.get('review-only')=="on"):
```
gentlecat: Parentheses are not needed.

gentlecat:
I can't get it to work for some reason. I pass `review-only=on` but still see all results.

gentlecat:
Thanks for those fixes! Since we now have an ability to connect to the MusicBrainz database directly, I'd rather use that method. Would definitely be much faster and efficient.

PR #17, opened 2015-12-17
Added apt links to search results during review

adeshp96:
Now the text can be clicked to open the respective links.


gentlecat:
I think these should open in new tab/window, since these just link to additional info about results people will be searching for. It's important not to loose the context.

adeshp96:
New tab fix done.

gentlecat:
Looks good. Thanks!

PR #21, opened 2015-12-22
Exclude drafts from public dumps

YatharthROCK:
Drafts reviews and revisions are included in the public data dumps. This should not be the case if users did not know that their drafts were not private.  #14 resolved [jira:CB-160](http://tickets.musicbrainz.org/browse/CB-160) by adding a note to warn users that drafts weren't private.

This PR resolves [jira:CB-208](http://tickets.musicbrainz.org/browse/CB-208), which attemps to solve the problem better by excluding draft reviews and revisions from the public data dumps: 

The commit messages contain detailed information about the implementation.

### DIFF
--- a/critiquebrainz/data/export_manager.py
+++ b/critiquebrainz/data/export_manager.py
@@ -171,7 +171,8 @@ def public(location=os.path.join(os.getcwd(), 'export', 'public'), rotate=False)
         reviews_combined_tables_dir = os.path.join(temp_dir, 'cbdump-reviews-all')
         create_path(reviews_combined_tables_dir)
         with open(os.path.join(reviews_combined_tables_dir, 'review'), 'w') as f:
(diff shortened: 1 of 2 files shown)


YatharthROCK:
@gentlecat This work was done in response to the GCI task “Warn users drafts are not private.” The task consisted of either implementing the quick fix in #14 or this PR's solution. The task was removed because the earlier fix was implemented outside of GCI. However, it would be nice if the task could be put back so this PR's distinct work can be submitted for that task.

uklauer:
+1 to the structure of the patch/commit series – each commit appears to make sense for itself. :-)

I’d drop the merge commit, though; either completely if there are no conflicts, or else I’d rebase the branch.

YatharthROCK:
@uklauer Thanks for the note! Is the detail of the git commit messages good? Also, I force pushed with my commits rebased on top of upstream's master.

gentlecat:
LGTM

PR #22, opened 2015-12-22
CB-207 Replace "Edit on MusicBrainz" icon

opatel99:
Replaced the old MusicBrainz icon file with the new MusicBrainz logo (16x16 SVG). I then adjusted all old references of the .ico image -> .svg image using the grep tool.

http://tickets.musicbrainz.org/browse/CB-207


critiquebrainz/frontend/templates/user/info.html
```diff
@@ -5,7 +5,7 @@
 {% block profile_content %}
   {% if user.musicbrainz_id %}
     <div>
-      <img src="{{ url_for('static', filename='favicons/musicbrainz-16.ico') }}" />
+      <img src="{{ url_for('static', filename='favicons/musicbrainz-16.png') }}" />
```
Freso: If this exact same file(name) is used in all these different places, maybe it would make sense to make it a variable defined somewhere higher up? I'm also thinking ahead to once (in the future) when we move to use [StaticBrainz](http://tickets.musicbrainz.org/browse/MBH-413) and will want to point to that instead - it will be easier to update it in one place, and be less likely to miss somewhere.

gentlecat:
Is there an SVG version of this icon? It'll probably be better to use that one instead for the same reasons we try to keep most of our other images in vector form.

gentlecat:
You forgot to remove PNG version of the icon.

gentlecat:
Looks good. Thank you.

PR #23, opened 2015-12-22
Display review language

YatharthROCK:
This PR resolves the part of http://tickets.musicbrainz.org/browse/CB-123 about simply displaying reviews' languages. The work is broken up into two commits.

### DIFF
--- a/critiquebrainz/frontend/__init__.py
+++ b/critiquebrainz/frontend/__init__.py
@@ -1,4 +1,6 @@
 from flask import Flask
+from Babel import Locale
+
 import os
 
 
@@ -70,6 +72,7 @@ def create_app(debug=None):
     app.jinja_env.filters['datetime'] = reformat_datetime
     app.jinja_env.filters['track_length'] = track_length
     app.jinja_env.filters['entity_details'] = musicbrainz.get_entity_by_id
+    app.jinja_env.filters['language_name'] = lambda language_code: Locale(language_code).language_name
 
     # Blueprints
     from critiquebrainz.frontend.views import frontend_bp
--- a/critiquebrainz/frontend/__init__.py
+++ b/critiquebrainz/frontend/__init__.py
@@ -1,5 +1,5 @@
 from flask import Flask
-from flask_babel import Locale
+from flask_babel import Locale, get_locale
 
 import os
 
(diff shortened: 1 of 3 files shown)


gentlecat:
Can we show flags? Is there a modern set of flag icons that will be suitable for this?

gentlecat:
Also, next time assign yourself to the task on JIRA and mark it as "In Review".

critiquebrainz/frontend/__init__.py
```diff
@@ -70,6 +72,7 @@ def create_app(debug=None):
     app.jinja_env.filters['datetime'] = reformat_datetime
     app.jinja_env.filters['track_length'] = track_length
     app.jinja_env.filters['entity_details'] = musicbrainz.get_entity_by_id
+    app.jinja_env.filters['language_name'] = lambda language_code: Locale(language_code).language_name
```
gentlecat: Can we display translated name of the language (in currently selected language)? This doesn't look very helpful:

YatharthROCK:
@LordSputnik Babel would have been installed automatically too, but I chose in c2e22de8 to simply use Flask-Babel instead of Babel.

critiquebrainz/frontend/__init__.py
YatharthROCK: Implemented in c372586

YatharthROCK:
@gentlecat I worked on implementing this a little bit, and I could do it, but I don't think we should.

Languages don't have a one-to-one correspondence with countries and thus flags. English or Portuguese both have two good candidates for flags, the choice of either of which is exclusionary to the other country. Arabic is worse. Then there are countries like India and Canada, not all of whose residents think "Hindi" and "English"/"French" when they see their flag.

YatharthROCK:
> Also, next time assign yourself to the task on JIRA and mark it as "In Review".

Should I have done this considering that my PR only addresses a part of CB-123? I suppose the correct solution here is to split up CB-123.

gentlecat:
CB-123 is for adding language indication. You basically created exactly the same ticket.

YatharthROCK:
> You basically created exactly the same ticket.

@gentlecat Yeah, it's a (strict) subset! I created it as a sub-ticket because there is another, distinct part of the ticket: To translate reviews. Perhaps _that_ should have just been forked of, and CB-123 restricted to displaying the language? I'm not sure as there was already discussion in CB-123 about translating reviews.

critiquebrainz/frontend/__init__.py
gentlecat: Definitely better now.

Freso:
Yeah, the automated review translation is something that came to the ticket after the ticket was made, so that was never a part of CB-123 to begin with, so that should be the thing getting its own ticket if anything. :)

Freso:
> Can we show flags? Is there a modern set of flag icons that will be suitable for this?

It is generally a bad a idea to use country flags for language indication. If we don't *have* to do this, I would prefer that we didn't.
http://www.worldaccent.com/blog/2012/09/flags-multilingual-website-navigation.html
https://www.cs.tut.fi/~jkorpela/flags.html

gentlecat:
Right, that makes sense. :+1:

PR #25, opened 2015-12-24
Show appropriate placeholder ‘cover art’ image for events

YatharthROCK:
This PR resolves [CB-186](http://tickets.musicbrainz.org/browse/CB-186). The commit summaries and full messages detail the implementation.

PR #26, opened 2015-12-24
Indicate type of entity of review where mixed

YatharthROCK:
This PR resolves [CB-184](http://tickets.musicbrainz.org/browse/CB-184). It builds upon #25 and so should only be merged after/if #25 is merged. The first commit's message reads:

> A macro was defined and used in the three places reviews of mixed-type entities are displayed.
> 
> It is up to the translators to translate 'release_group' in English as 'release group' (and similarly for other languages).


YatharthROCK:
This branch was rebased on top of #25; the older discussion on GitHub is [here](https://github.com/YatharthROCK/critiquebrainz/commit/4986e8d8d08fd0c38ae17a48528e654df4ba83e1).

gentlecat:
Try not to rebase branches like that. We don't want to loose discussion.

critiquebrainz/frontend/templates/macros.html
```diff
@@ -35,3 +35,27 @@
...
+            $(this).attr("src", "/static/img/missing-art.png");
+          })
+          .attr("src", "//coverartarchive.org/release-group/{{ entity_id }}/front-250");
+    </script>
+  {% endif %}
+{% endmacro %}
+
+{% macro review_byline(review) %}
+{{ _(
+  '%(entitytype)s review by %(user)s',
```
mwiencek: That may not be ideal for translators. The strings `_('Event')` and `_('Music')` below have no context, and may be translated differently alone than they would be in this sentence. I'd use separate strings for each entity type.

YatharthROCK: @mwiencek That makes sense; thanks! Resolved in d9e5144

YatharthROCK:
@gentlecat Ok; I'll merge in the future. ‘Nested’ PRs are complicated…

critiquebrainz/frontend/templates/macros.html
```diff
@@ -51,3 +51,11 @@
     </script>
   {% endif %}
 {% endmacro %}
+
+{% macro review_credit(review) %}
+{{ _(
+  'Event review by %(user)s' if review.entity_type == 'event' else 'Music review by %(user)s' if review.entity_type == 'release_group' else 'Review by %(user)s',
```
gentlecat: `Music` is way too general. It's probably better to just put `Release group` there.

critiquebrainz/frontend/templates/macros.html
```diff
@@ -51,3 +51,11 @@
...
+  user = '<img class="avatar" src="%s&s=16" /> '|safe % review.user.avatar + '<a href="%s">%s</a>'|safe % (url_for('user.reviews', user_id=review.user.id), review.user.display_name),
+  entitytype = _('Event')
+) }}
```
gentlecat: Also, can you update this macro a bit so it doesn't take so much space horizontally? It's hard to read and understand. I'd just replace shortcut conditionals with full versions.

critiquebrainz/frontend/templates/macros.html
YatharthROCK: @gentlecat From a0b68ac92's commit message:

> http://i.imgur.com/UTbDvxj.png
> Previously, the review byline would sometimes get cut off. It has been
shortened.

I think that experienced users will not be confused by the `Music` because there is only one musical entity type that reviews apply too. As for new users, they would be as if not less confused by `Music` than by `Release Group`, not being familiar with the latter term.

Yet another alternative might be to move the entity type indication to the _end_ of the review credits; however, this too results in ugly chopping off.

critiquebrainz/frontend/templates/macros.html
YatharthROCK: @gentlecat I wrapped the other line too; see aabdd31

critiquebrainz/frontend/templates/macros.html
gentlecat: We need to discuss the alternatives then. I don't think `Music` is a good one.

critiquebrainz/frontend/templates/macros.html
```diff
@@ -51,3 +51,17 @@
...
+
+{% macro review_credit(review) %}
+  {% if review.entity_type == 'event' %}
+    {% set credit_string = 'Event review by %(user)s' %}
+  {% elif review.entity_type == 'release_group' %}
+    {% set credit_string = 'Music review by %(user)s' %}
+  {% else %}
+    {% set credit_string = 'Review by %(user)s' %}
+  {% endif %}
+  {{ _(credit_string,
```
gentlecat: Are strings extracted correctly when specified like that?

YatharthROCK: @gentlecat Oh, right; I hadn't realized that, not having extracted the strings myself. Tested and resolved in 7876cd72814

YatharthROCK:
There is more discussion in [CB-184](http://tickets.musicbrainz.org/browse/CB-184).

For now, "Release Group review by user" is shown on the landing page, and "Review by user" followed by the date and then the entity type on the Browse page.

gentlecat:
Regarding your prototype on JSFiddle: http://chatlogs.metabrainz.org/brainzbot/metabrainz/2015-12-28/?msg=3455270&page=2.

I don't like how we have two ways to display basically the same thing in listings now. It should be consistent and easy to understand just by glancing over.

YatharthROCK:
@gentlecat Would you want text overload on the cover art on the review detail page too, where there is ample space for saying "Event review by <user>" but where overlaying text on the cover art might look ugly? If not, would the inconsistency between the review detail page and everywhere else reviews are listed be acceptable to you?

I've proposed many solutions until now, some of which would be fine with me and Freso, but I'm not sure what it is that you think would be good solution (in large part because our schedules make it difficult to have synchronous conversation in IRC).

gentlecat:
This is primarily for lists. On review pages we clearly reference which entity is being reviewed.

YatharthROCK:
> This is primarily for lists.

Okay; I'll go with sth like the JSFiddle but without needing to hover to see the text.

> On review pages we clearly reference which entity is being reviewed.

@gentlecat I don't see the words `Release Group` or `Event` explicitly anywhere on the review detail page; is it not worth adding?

YatharthROCK:
@gentlecat It now looks like this. How else would you like it to look like?
screen shot 2015-12-29 at 5 02 29 pm
]

gentlecat:
> Events always use the location pin; do they need text?
> RG Reviews with no cover art show a cd image; would text be appropriate for them? I know that CB plans to have reviews for not just RGs and events in the future, so it might be appropriate.

Yes, I think it should be displayed on every item.

This is what I had in mind:


And here's the change to your branch: 1261a187a276e7cac6a21a4789796d8902cf4fee. Though it's a bit buggy in Firefox and Safari:


So what you can do is cherry-pick my commit and fix problems there.

> I don't see the words Release Group or Event explicitly anywhere on the review detail page. Did you not mention the following in the ticket yourself?
>> Individual review pages could use some improvements related to that. I think it's better to let user know what it is explicitly.

Right, I forgot that is still to be added. Name of the entity type in the title should be enough there.

gentlecat:
I merged this in 9c9add3c914fc4a427ff537ae8f9a065f1fc9a9c after making some adjustments that I mentioned before.

PR #31, opened 2015-12-29
Untranslatable string in revision list

Debeet:


critiquebrainz/frontend/templates/review/revision_results.html
```diff
@@ -1,6 +1,6 @@
 {% for result in results %}
   <tr>
-    <td><a href="{{ url_for('review.entity', id=review.id, rev=result[0]+1) }}">Revision {{ result[0] + 1 }}</a></td>
+    <td><a href="{{ url_for('review.entity', id=review.id, rev=result[0]+1) }}">{{ _('Revision') }} {{ result[0] + 1 }}</a></td>
```
gentlecat: It's probably better to make it more explicit:

    {{ _('Revision %(number)s', number=result[0]+1) }}

Debeet:
Thanks for the input. Yeah, it's better that way. I updated the fix.

gentlecat:
Perfect. Thank you!

PR #33, opened 2016-01-02
CB-146 embed spotify player in review editor

caroline-gschwend:
Also adds spotify macro to macros.html and updates all usages of spotify player to use macro.

### DIFF
--- a/critiquebrainz/frontend/templates/macros.html
+++ b/critiquebrainz/frontend/templates/macros.html
@@ -51,3 +51,17 @@
     </script>
   {% endif %}
 {% endmacro %}
+
+{% macro spotify_player(spotify_mappings) -%}
+    {% if spotify_mappings %}
+        <iframe src="https://embed.spotify.com/?uri={{ spotify_mappings[0] }}&theme=white"
+                width="100%" height="80" frameborder="0" allowtransparency="true"></iframe>
+        <em><small><a class="text-muted" href="{{ url_for('mapping.spotify_list', release_group_id=id) }}">{{ _('View Spotify mapping information') }}</a></small></em>
+      {% else %}
+        <span class="text-muted">
+          {{ _('We don\'t have a mapping between this release group and Spotify.
+          Please help us find the right album.') }}
+        </span>
+        <div style="text-align:center;font-weight:bold;"><a style="font-size:16px;" href="{{ url_for('mapping.spotify', release_group_id=id) }}" rel="nofollow">{{ _('Match this!') }}</a></div>
+      {% endif %}
+{%- endmacro %}
--- a/critiquebrainz/frontend/templates/macros.html
+++ b/critiquebrainz/frontend/templates/macros.html
@@ -52,16 +52,14 @@
   {% endif %}
 {% endmacro %}
 
-{% macro spotify_player(spotify_mappings) -%}
-    {% if spotify_mappings %}
-        <iframe src="https://embed.spotify.com/?uri={{ spotify_mappings[0] }}&theme=white"
-                width="100%" height="80" frameborder="0" allowtransparency="true"></iframe>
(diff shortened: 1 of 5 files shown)


critiquebrainz/frontend/templates/macros.html
```diff
@@ -51,3 +51,17 @@
...
+    {% if spotify_mappings %}
+        <iframe src="https://embed.spotify.com/?uri={{ spotify_mappings[0] }}&theme=white"
+                width="100%" height="80" frameborder="0" allowtransparency="true"></iframe>
+        <em><small><a class="text-muted" href="{{ url_for('mapping.spotify_list', release_group_id=id) }}">{{ _('View Spotify mapping information') }}</a></small></em>
+      {% else %}
+        <span class="text-muted">
+          {{ _('We don\'t have a mapping between this release group and Spotify.
+          Please help us find the right album.') }}
+        </span>
+        <div style="text-align:center;font-weight:bold;"><a style="font-size:16px;" href="{{ url_for('mapping.spotify', release_group_id=id) }}" rel="nofollow">{{ _('Match this!') }}</a></div>
```
gentlecat: It means that this block offering to add a mapping will now show up on all pages. It might be better to make it optional by having another argument in that macro.

critiquebrainz/frontend/templates/review/modify/release_group.html
```diff
@@ -20,4 +21,8 @@
 </div>
 <div class="col-md-2">
   {{ cover_art(entity['id'], 'release_group', 'id=cover-art class=img-responsive style=max-height:120px;max-width:120px;') }}
+  
+  <div id="spotify-play-button">
+	 {{ spotify_player(spotify_mappings) }}
```
gentlecat: Player will never show up here because `spotify_mappings` is undefined here.

gentlecat:
You should try to keep indentation consistent with the rest of the files (no tabs, 4 or 2 spaces depending on whether it is Python or JavaScript).

critiquebrainz/frontend/views/review.py
```diff
@@ -197,6 +197,7 @@ def create():
         flash(gettext("You can only write a review for an entity that exists on MusicBrainz!"), 'error')
         return redirect(url_for('search.selector', next=url_for('.create')))
     return render_template('review/write.html', form=form, entity_type=entity_type, entity=entity)
+    return render_template('review/entity/release_group.html' % review.entity_type, review=review, spotify_mappings=spotify_mappings, vote=vote)
```
gentlecat: This part of code will never be reached. It's a good idea to test your changes beforehand. To connect to server with Spotify mappings you can update config file to have this (just make sure that you overwrite old variables):
```python
# Server with Spotify mappings
# https://github.com/metabrainz/mbspotify
MBSPOTIFY_BASE_URI = "http://mbspotify.musicbrainz.org/"
#MBSPOTIFY_ACCESS_KEY = ""
```

gentlecat:
Player will not show up when editing a review. You probably need to make similar modifications in view that is responsible for editing in *frontend/views/review.py*.

critiquebrainz/frontend/views/review.py
```diff
@@ -198,6 +198,10 @@ def create():
         return redirect(url_for('search.selector', next=url_for('.create')))
     return render_template('review/write.html', form=form, entity_type=entity_type, entity=entity)
 
+    if review.entity_type == 'release_group':
```
gentlecat: This part of code will never be reached because there's a `return` right above it.

gentlecat: Also you should use `entity_type` there, not `review.entity_type`. Please test your changes and make sure they work before submitting.

gentlecat:
Squashed and merged in 53f92eb16a77a1e69be5df243cfd1833d4dc5d84.

PR #34, opened 2016-01-10
CB-200: Use 403 instead of 404 for reviews that are hidden

dufferzafar:
http://tickets.musicbrainz.org/browse/CB-200


gentlecat:
Thanks.

PR #38, opened 2016-01-18
CB-134 More information about Spotify albums on mapping page is shown now.

stasszczesniak:
CritiqueBrainz provides an embedded Spotify player on pages that are mapped to it in mbspotify. Mappings are created by users on CritiqueBrainz using a search interface that we have. Search gets data from Spotify using their Web API, but the only parts we use are cover art and name of the album. Now users can also see the artists that made some particular album and the release year.

[Related ticket] (http://tickets.musicbrainz.org/browse/CB-134)

Prototypes: 

1. **Artists from tracks if unavailable, date in the same line as artists and in brackets.**
data_w_nawiasie_przy_artystach_artysci_ze_sciezek_male

2. **Artists from default artists, date in the same line as artists and in brackets.**
data_w_nawiasie_przy_artystach_male

3. **Artists from tracks if unavailable, date in the same line as type and in brackets.**
data_w_nawiasie_przy_typie_artysci_ze_sciezek

4. **Artists from default artists, date in the same line as type and in brackets.**
data_w_nawiasie_przy_typie

5. **Artists from default artists, date in the next line.**
data_na_dole_various

6. **Artists from tracks if unavailable, date in the next line.**
data_na_dole_sciezki

7. **All artists from tracks if unavailable, date in the next line**
all_artists_from_tracks

8. **Freso's idea**
freso

### DIFF
--- a/critiquebrainz/frontend/templates/mapping/spotify.html
+++ b/critiquebrainz/frontend/templates/mapping/spotify.html
@@ -29,9 +29,15 @@
           </a>
         </div>
         <div class="caption">
-          <div class="title">{{ item.name }}</div>
-          <span class="text-muted">{{ item.album_type }}</span><br />
-          <em><a href="{{ item.external_urls.spotify }}">{{ _('Listen on Spotify') }}</a></em><br />
+            <div class="title">
+                {{ item.name }}
+                <br>
+                <span style="font-weight:normal">by </span>
+                {{ artists(item) }}
+                {{ release_date_string(item) }}
+            </div>
+            <span class="text-muted">{{ item.album_type }}</span><br />
+            <em><a href="{{ item.external_urls.spotify }}">{{ _('Listen on Spotify') }}</a></em><br />
         </div>
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


Freso:
Okay, this is crazy, but how about something like
"**Album Name**
by **Artist Name**
**2009** single
*Listen on Spotify*"
?

stasszczesniak:
Added the #8 option regarding Freso's proposition.

Freso:
I actually like that. It both visually expresses the difference, but at the same time also makes sense (e.g., »that's Iggy Pop's 2006 album«).

gentlecat:
Yeah, I agree, that looks good. Is it implemented here or do you need to make changes?

critiquebrainz/frontend/templates/mapping/spotify.html
```diff
@@ -29,9 +29,16 @@
           </a>
         </div>
         <div class="caption">
-          <div class="title">{{ item.name }}</div>
-          <span class="text-muted">{{ item.album_type }}</span><br />
-          <em><a href="{{ item.external_urls.spotify }}">{{ _('Listen on Spotify') }}</a></em><br />
+            <div class="title">
+                {{ item.name }}
+                <br>
```
gentlecat: <br />

critiquebrainz/frontend/templates/mapping/spotify.html
```diff
@@ -29,9 +29,16 @@
...
+                <span style="font-weight:normal">by </span>
+                {{ artists(item) }}
+                <br>
```
gentlecat: Same here.

critiquebrainz/frontend/templates/mapping/spotify.html
```diff
@@ -29,9 +29,16 @@
...
+            </div>
+            <b> {{ release_date_string(item) }} </b>
```
gentlecat: There should be no spaces after opening and before closing tags.

gentlecat:
You need to fix indentation in the template. It needs to be 2 spaces.

stasszczesniak:
Should be fine now.

PR #39, opened 2016-01-19
Fixed the results from Spotify to show in the right order.

stasszczesniak:
Fixed the results from Spotify to show in the right order.


gentlecat:
:+1:

PR #41, opened 2016-01-20
CB-196 Artist pages show the same member multiple times

stasszczesniak:
[CB-196](http://tickets.musicbrainz.org/browse/CB-196) [GCI](https://codein.withgoogle.com/dashboard/task-instances/6192451874193408/?sp-page=1) 
Artist pages showed the same member multiple times, so I've added merging of these duplicate members and showing the list of attributes, periods for each member. Former members attributes are not shown, because periods list might be too long and it wouldn't look nicely together. 
**Current prototype screens:** (click the artist name to see the current view on CB)
  
Note: attribute names lists may have spaces before commas, it is solved now but it left on the screenshots.
  
[**Foo fighters**](https://critiquebrainz.org/artist/67f66c07-6e61-4026-ade5-7e782fad3a5d)
foo fighters critiquebrainzb

[**Drifters**](https://critiquebrainz.org/artist/9e2d3f58-0653-4007-bcb7-1746fcdd9363)
driftersb
(note: the image is cut on purpose and the list is not cut in the realty, I've not changed the rest of the code)

[**Tiny Hoedown**](https://critiquebrainz.org/artist/af57fdcb-a3a4-493d-a3a9-028c591a70ed)
tiny hoedown critiquebrainzb

[**Arcade Fire**](https://critiquebrainz.org/artist/52074ba6-e495-4ef3-9bb4-0703888a9f68)
arcade fire critiquebrainzb

[**AC/DC**](https://critiquebrainz.org/artist/66c662b6-6e2f-4930-8610-912e24c63ed1)
ac dc critiquebrainzb

[**AC/DC** with disambiguation](https://critiquebrainz.org/artist/66c662b6-6e2f-4930-8610-912e24c63ed1)
ac dc critiquebrainz_with_dis

### DIFF
--- a/critiquebrainz/frontend/templates/artist.html
+++ b/critiquebrainz/frontend/templates/artist.html
@@ -115,7 +115,7 @@
     <h4>{{ _('Artist information') }}</h4>
     {% if artist.type %}<p class="text-muted">{{ artist.type }}</p>{% endif %}
 
-    {% if artist.current_members %}
+    {% if band_members.current_members %}
       <b>
         {% if artist.type and artist.type == 'Person' %}
           {{ _('Member of') }}
@@ -124,17 +124,17 @@
         {% endif %}
       </b>
       <ul class="list-unstyled">
-        {{ print_members(artist.current_members[:15]) }}
-        {% if artist.current_members | length > 15 %}
+        {{ print_members(band_members.current_members[:15]) }}
+        {% if band_members.current_members | length > 15 %}
           <div id="more-members" class="collapse">
-            {{ print_members(artist.current_members[15:]) }}
+            {{ print_members(band_members.current_members[15:]) }}
           </div>
           <a data-toggle="collapse" data-target="#more-members" style="cursor: pointer;"><strong>{{ _('View all...') }}</strong></a>
         {% endif %}
       </ul>
     {% endif %}
 
-    {% if artist.former_members %}
+    {% if band_members.former_members %}
       <b>
         {% if artist.type and artist.type == 'Person' %}
           {{ _('Former member of') }}
@@ -143,10 +143,10 @@
         {% endif %}
       </b>
       <ul class="list-unstyled">
-        {{ print_members(artist.former_members[:15]) }}
-        {% if artist.former_members | length > 15 %}
+        {{ print_members(band_members.former_members[:15]) }}
+        {% if band_members.former_members | length > 15 %}
           <div id="more-former-members" class="collapse">
-            {{ print_members(artist.former_members[15:]) }}
+            {{ print_members(band_members.former_members[15:]) }}
           </div>
           <a data-toggle="collapse" data-target="#more-former-members" style="cursor: pointer;"><strong>{{ _('View all...') }}</strong></a>
         {% endif %}
--- a/critiquebrainz/frontend/templates/artist.html
+++ b/critiquebrainz/frontend/templates/artist.html
@@ -10,24 +10,24 @@
          {% if member.begin %}title="{{ _('Member since %(date)s', date=member.begin) }}"{% endif %}>
         {{ member.artist.name }}
       </a>
-      {% if member.periods -%}
-        <span class="text-muted">(
-        {%- for period in member.periods -%}
-            {{ period[0] + '-' + period[1] }}
-            {%- if not loop.last -%}
-              {{ ', ' }}
-            {%- endif -%}
-        {%- endfor -%}
-        )</span>
+      {% if member.periods %}
+        {% if not member.ended and member.periods and member.periods[0][0] %}
+          <span class="text-muted">
+            ({{ 'since ' + member.periods[0][0] }})
+          </span>
+        {% else %}
+          <span class="text-muted">
+          ({% for period in member.periods -%}
+            {{ period[0] + '-' + period[1] + ', ' * (not loop.last) }}
+          {%- endfor %})
+          </span>
+        {%- endif %}
       {% endif %}
       {% if member['attribute-list'] and not member.ended %}
         <br />
         <span class="text-muted">
-          {% for attr in member['attribute-list']%}
-            {{ attr }}
-            {% if not loop.last %}
-              {{ ', ' }}
-            {% endif %}
+          {% for attr in member['attribute-list'] -%}
+            {{ attr + ', ' * (not loop.last) }}
           {% endfor %}
         </span>
       {% endif %}
--- a/critiquebrainz/frontend/templates/artist.html
+++ b/critiquebrainz/frontend/templates/artist.html
@@ -10,10 +10,26 @@
          {% if member.begin %}title="{{ _('Member since %(date)s', date=member.begin) }}"{% endif %}>
         {{ member.artist.name }}
       </a>
-      {% if member.end %}
-        <span class="text-muted">({{ member.begin }} - {{ member.end }})</span>
-      {% elif member.artist.disambiguation %}
-        <span class="text-muted">({{ member.artist.disambiguation }})</span>
+      {% if member.periods -%}
+        <span class="text-muted">(
+        {%- for period in member.periods -%}
+            {{ period[0] + '-' + period[1] }}
+            {%- if not loop.last -%}
+              {{ ', ' }}
+            {%- endif -%}
+        {%- endfor -%}
+        )</span>
+      {% endif %}
+      {% if member['attribute-list'] and not member.ended %}
+        <br />
+        <span class="text-muted">
+          {% for attr in member['attribute-list']%}
+            {{ attr }}
+            {% if not loop.last %}
+              {{ ', ' }}
+            {% endif %}
+          {% endfor %}
+        </span>
       {% endif %}
     </li>
   {% endfor %}
--- a/critiquebrainz/frontend/templates/artist.html
+++ b/critiquebrainz/frontend/templates/artist.html
@@ -6,9 +6,8 @@
 {% macro print_members(members) %}
   {% for member in members %}
     <li>
-      <a href="{{ url_for('artist.entity', id=member.artist.id) }}"
-         {% if member.begin %}title="{{ _('Member since %(date)s', date=member.begin) }}"{% endif %}>
-        {{ member.artist.name }}
+      <a href="{{ url_for('artist.entity', id=member.artist_id) }}">
+        {{ member.name }}
       </a>
       {% if member.periods %}
         {% if not member.ended and member.periods and member.periods[0][0] %}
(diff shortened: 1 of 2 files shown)


Freso:
That's a very neat artist you've found there. How will it handles people who stepped out of the group and later rejoined? Do the members have disambiguation (if so, why isn't it shown)? How about former members with an unknown join or part date?

critiquebrainz/frontend/templates/artist.html
```diff
@@ -10,10 +10,26 @@
...
-      {% elif member.artist.disambiguation %}
-        <span class="text-muted">({{ member.artist.disambiguation }})</span>
+      {% if member.periods -%}
+        <span class="text-muted">(
+        {%- for period in member.periods -%}
+            {{ period[0] + '-' + period[1] }}
+            {%- if not loop.last -%}
+              {{ ', ' }}
+            {%- endif -%}
+        {%- endfor -%}
```
gentlecat: Why are you inserting `-` before and/or after sometimes and leave it off in other places? Either be consistent and use it in all your code, or drop it completely.

critiquebrainz/frontend/templates/artist.html
```diff
@@ -10,10 +10,26 @@
...
+        )</span>
+      {% endif %}
+      {% if member['attribute-list'] and not member.ended %}
+        <br />
+        <span class="text-muted">
+          {% for attr in member['attribute-list']%}
+            {{ attr }}
+            {% if not loop.last %}
+              {{ ', ' }}
+            {% endif %}
```
gentlecat: Can you reduce this to `', ' if ...`?

critiquebrainz/frontend/templates/artist.html
```diff
@@ -10,10 +10,26 @@
...
+        <span class="text-muted">(
+        {%- for period in member.periods -%}
+            {{ period[0] + '-' + period[1] }}
+            {%- if not loop.last -%}
+              {{ ', ' }}
+            {%- endif -%}
+        {%- endfor -%}
+        )</span>
+      {% endif %}
+      {% if member['attribute-list'] and not member.ended %}
```
gentlecat: Again, reference values in a dictionary consistently. Either use `member['whatever']` or `member.whatever`. Not both.

critiquebrainz/frontend/views/artist.py
```diff
@@ -46,3 +48,50 @@ def entity(id):
...
+
+
+def squash_members_attributes(members):
+    members = [member.copy() for member in members]
+    ids_order = dict([(member['target'], index) for index, member in enumerate(members)])
+    
+    members.sort(key=lambda x: x['target'])
+    result_members = []
+
+    for i in range(0, len(members)):
```
gentlecat: This is not how you iterate over a list in Python. See http://learnpythonthehardway.org/book/ex32.html for example.

critiquebrainz/frontend/views/artist.py
```diff
@@ -46,3 +48,50 @@ def entity(id):
...
+        return (begin_date, end_date)
+    else:
+        return None
+
+
+def get_year_from_date(date):
+    if date:
+        return date.split('-')[0]
+    else:
+        return ''
```
gentlecat: All these functions you added are supposed to be used only within this module, so it's better to prefix them with `_` to indicate that.

gentlecat: I'd even move `get_year_from_date` function inside `get_period` (but don't add the prefix then, since it would be in a different scope).

critiquebrainz/frontend/views/artist.py
```diff
@@ -46,3 +48,50 @@ def entity(id):
...
+                members[i]['attribute-list'].extend(members[j].get('attribute-list', []))
+                period = get_period(members[i])
+                if(period):
+                    members[i]['periods'].append(period)
+                j += 1
+            result_members.append(members[i])
+            members[i]['periods'].sort(key=lambda x: x[1], reverse=True)
+
+    result_members.sort(key=lambda x: ids_order[x['target']])
+    return result_members
```
gentlecat: Can you explain what you are doing in this function?

critiquebrainz/frontend/templates/artist.html
stasszczesniak: Because I don't want any whitespace in some places. I can use it in all my code but I don't see any value in changing my whole code.

critiquebrainz/frontend/templates/artist.html
stasszczesniak: What do you mean ?

critiquebrainz/frontend/templates/artist.html
stasszczesniak: member.attribute-list shows warning in my editor

critiquebrainz/frontend/views/artist.py
stasszczesniak: But it works ... ? Doesn't it ? Which line do you have in mind particularly ?

critiquebrainz/frontend/views/artist.py
stasszczesniak: I am sorting the elements firstly by 'target' key.
Than I am squashing the elements with the same 'target'
by finding the first element with some target and then iterating over all the elements that have the same target and then squashing all of them.

critiquebrainz/frontend/views/artist.py
stasszczesniak: Done.

critiquebrainz/frontend/templates/artist.html
gentlecat: {{ ', ' if ... }}

critiquebrainz/frontend/templates/artist.html
gentlecat: Warning probably because variable names can't have `-` character in them. So it's probably better to use the first option. I know some of existing code is inconsistent either, but there's no reason to make it worse.

critiquebrainz/frontend/templates/artist.html
stasszczesniak: I am guessing that 
```html
{% if not loop.last %}
   {{ ', ' }}
{% endif %}
```
means the same as the thing you posted in jinja2, right ?

critiquebrainz/frontend/templates/artist.html
stasszczesniak: Yeah, I remember - it was because of this. Ok, I will change it then.

stasszczesniak: But the second option is used often in the rest of the file. Do you think it would be worth changing ? Maybe in another PR ?

critiquebrainz/frontend/templates/artist.html
gentlecat: Yes, but don't remember exactly how it works.

critiquebrainz/frontend/views/artist.py
gentlecat: One that I left a comment on (60). "It works" doesn't mean it's good enough.

stasszczesniak: Yeah, that's right. I am not familiar with python etiquette other than PEP8

stasszczesniak: I would be using 
```python
  for member in members:
```
if i don't use i in the next few lines

critiquebrainz/frontend/templates/artist.html
stasszczesniak: Ok. At first I didn't know what do you mean, but now I do fortunately :) Changed.

critiquebrainz/frontend/templates/artist.html
stasszczesniak: I mean consistency is an important factor, but I think it would be like changing all fors to whiles to be consistent. They are not doing the same so I see no point in using only one version. Prove me wrong if I am not thinking clearly ;)

critiquebrainz/frontend/views/artist.py
gentlecat: This seems too complicated for what you are trying to do. What if you just build a list with a dictionary for each member:
```
{
  "name": ...,
  "periods": [..],
}
```

critiquebrainz/frontend/templates/artist.html
gentlecat: I don't remember introducing these hyphens in our codebase, they are unnecessary and distracting. http://jinja.pocoo.org/docs/dev/templates/#whitespace-control

critiquebrainz/frontend/views/artist.py
stasszczesniak: This way it is done in O(n log n), where n is the number of members (and let's assume that the number of periods and attributes is constant). I could easily do it in O(n^2) and I am currently thinking how to do that fastly and simply at the same time.

stasszczesniak: This way it is done in O(n log n), where n is the number of members (and let's assume that the number of periods and attributes is constant). I could easily do it in O(n^2) and I am currently thinking how to do that fastly and simply at the same time.

stasszczesniak: Ok, I could do what you've just written.

gentlecat: This doesn't matter when you are working with a list of, let's say, 4 items on average.

critiquebrainz/frontend/templates/artist.html
stasszczesniak: I've removed the unnecessary hyphens.

stasszczesniak:
Disambiguations are removed for now, because I don't see any place to show them. I can do it but it wouldn't just look neat.

critiquebrainz/frontend/views/artist.py
```diff
@@ -46,3 +48,56 @@ def entity(id):
         release_group['review_count'] = review_count
     return render_template('artist.html', id=id, artist=artist, release_type=release_type,
                            release_groups=release_groups, page=page, limit=limit, count=count)
+
+
+def squash_duplicated_members(members):
```
gentlecat: I would add a docstring here, describe what's going on and what potentially confusing stuff like `target` is.

gentlecat: You also forgot to prefix it with `_`.

critiquebrainz/frontend/views/artist.py
```diff
@@ -46,3 +39,55 @@ def entity(id):
...
+        if artist_id in members_by_artist_id:
+            members_by_artist_id[artist_id]['attributes'].extend(member.get('attribute-list', []))
+        else:
+            members_by_artist_id[artist_id] = {
+                'name': member.get('artist', {}).get('name', ''),
+                'periods': [],
+                'attributes': member.get('attribute-list', []),
+                'disambiguation': member.get('disambiguation', ''),
+                'artist_id': member.get('artist', {}).get('id', None),
+                'ended': member.get('ended', False)
```
gentlecat: It's incorrect to assume value of `ended` if it's undefined.

critiquebrainz/frontend/views/artist.py
```diff
@@ -18,16 +19,8 @@ def entity(id):
...
-        artist['current_members'] = []
-        artist['former_members'] = []
-        for member in artist['band-members']:
-            if 'ended' in member and member['ended'] == 'true':
-                artist['former_members'].append(member)
-            else:
-                artist['current_members'].append(member)
-
+    # Note that some artists might not have a list of members because they are not a band
+    artist.update(_get_band_members(artist))
```
gentlecat: Just create a new variable `band_members` and pass it to the template.

PR #43, opened 2016-02-14
Use server_default for User.is_blocked

dufferzafar:
I am not really sure why, but I had to do this to get the database 
working properly.


gentlecat:
Do you need to change the schema as well?

gentlecat:
This might actually be caused by outdated data dumps. Need to check.

gentlecat:
@brainzbot, ok to build.

gentlecat:
@brainzbot, ok to test.

PR #44, opened 2016-02-14
CB-214: /user/name/{username} redirects to userpage

dufferzafar:
This is a cheap way to 'fix' [CB-214](http://tickets.musicbrainz.org/browse/CB-214)

The 'correct' way would involve modifying a lot of places in the source.

I'm not really sure if this makes any sense, though :|

### DIFF
--- a/critiquebrainz/frontend/views/user.py
+++ b/critiquebrainz/frontend/views/user.py
@@ -2,6 +2,7 @@ from flask import Blueprint, render_template, request, redirect, url_for, flash
 from flask_babel import gettext
 from flask_login import login_required, current_user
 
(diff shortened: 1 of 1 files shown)


critiquebrainz/frontend/views/user.py
```diff
@@ -11,6 +12,14 @@
 user_bp = Blueprint('user', __name__)
 
 
+@user_bp.route('/name/<string:username>')
+def name(username):
+    user = User.get(display_name=username)
```
gentlecat: `display_name` is not unique.

gentlecat:
I commented on the ticket as well.

PR #46, opened 2016-02-14
CB-172: Remove header from flash messages

dufferzafar:
The [JIRA ticket](http://tickets.musicbrainz.org/browse/CB-172) talks about a different way of implementing the flash message system, but I felt that a lot of the places in the code will need to be changed for it, and maybe it wouldn't be worth it.


gentlecat:
This looks good, but I think it would still be nice to have it updated to use the wrapper.

PR #47, opened 2016-02-15
CB-89: Fix language selector behaviour in presence of query string

dufferzafar:
[JIRA Ticket](http://tickets.musicbrainz.org/browse/CB-89)


gentlecat:
This looks good. You forgot about selector in the footer.

PR #50, opened 2016-03-31
CB-201: Better missing art icon

caroline-gschwend:
Just an idea for the missing art disc placeholder image. 

http://tickets.musicbrainz.org/browse/CB-201


gentlecat:
I don't think that it's better. Current image is used specifically to indicate a missing release group, which is vaguely resembled by the disc/recording icon. It is another way to distinguish types of entities in CritiqueBrainz when cover art (or other type of image) is missing.

From the ticket:
> The present missing-art icon is a grey circle which is tooo boring.

As I said in IRC, it's not meant to be fun.

gentlecat:
Thanks for sketching that though. :+1:

I'll close the ticket for now.

PR #54, opened 2016-06-06
Docker support

gentlecat:
*Documentation still needs to be written.*

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,5 @@
+.git
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
@@ -8,7 +10,7 @@ venv
 /build
 
 # Node
-/node_modules
+node_modules
 
 # Logs
 *.log
@@ -20,6 +22,9 @@ pip-delete-this-directory.txt
 
 # Backups and dumps
 /export/
+
+# Static builds
+/critiquebrainz/frontend/static/build
 
 # Test results
 htmlcov
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,37 @@
+version: '2'
+services:
+
+  db:
+    build:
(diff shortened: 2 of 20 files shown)


gentlecat:
Also need to figure out how to do configuration properly.

.dockerignore
```diff
@@ -0,0 +1,33 @@
...
+# Node
+node_modules
+
+# Logs
+*.log
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Application data
+/data
```
mayhem: is this supposed to be absolute?

docker/nginx_custom/nginx.conf
```diff
@@ -0,0 +1,24 @@
+worker_processes 4;
```
mayhem: Any thoughts on how to do docker configs for dev vs deployment?

mayhem:
Overall this looks pretty good to me. Clearly we still need to learn more about different configurations (dev vs production), but this certainly looks good enough to try and do a test deploy.

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+      - .:/code
+      - ./data/app:/data
+    ports:
+      - "3031:3031"
+    depends_on:
+      - db
+      - memcached
+
+  nginx:
+    restart: always
```
mayhem: Why are the other modules not marked at restart?

mayhem:
Aside from the comments already posted and the gulp/static stuff, this is looking VERY GOOD.

.dockerignore
gentlecat: Yes. It's in the root of the context (git repo). https://docs.docker.com/engine/reference/builder/#dockerignore-file

gentlecat:
This branch has been merged into `beta`.

PR #65, opened 2016-12-07
CB-125: Added Bandcamp Player Integration

anshuman73:
Added support for Bandcamp Player.


anshuman73:
screenshot from 2016-12-08 00-19-17
screenshot from 2016-12-08 00-19-55
Screenshots attached

critiquebrainz/frontend/external/external_players.py
```diff
@@ -0,0 +1,29 @@
...
+            elif re.match(r'^(http|https)://[A-Za-z0-9]+.bandcamp.com/?', url_rel['target']):
+                streaming_url_rels['bandcamp'] = bandcamp_parser(url_rel['target'])
+    return streaming_url_rels
+
+
+def bandcamp_parser(bandcamp_link):
+    """Gives the unique bandcamp id of the entered album/release
+    Required for embedding the player.
+    """
+    html = requests.get(bandcamp_link).text
```
gentlecat: I don't think I like making requests to other services on page loads like that.

gentlecat: In that case we'll have to skip this until something changes.

anshuman73: Alright !

Freso: > I don't think I like making requests to other services on page loads like that.

Why not if that's the way Bandcamp wants it to be done?

anshuman73: I think because it slows down the page load. When I was monitoring my network, i saw it takes upto 3s for the initial page load the initial html of the page(which probably includes this request)

PR #67, opened 2016-12-08
CB-117: Using identicon as default Gravatar

anshuman73:
Fixed default gravatar to "indenticon" from "mm"

### DIFF
--- a/critiquebrainz/data/model/user.py
+++ b/critiquebrainz/data/model/user.py
@@ -95,9 +95,9 @@ class User(db.Model, AdminMixin, DeleteMixin):
     def avatar(self):
         """Link to user's avatar image."""
         if self.show_gravatar and self.email:
-            return "https://gravatar.com/avatar/" + hashlib.md5(self.email.encode("ascii")).hexdigest() + "?d=mm&r=pg"
+            return "https://gravatar.com/avatar/" + hashlib.md5(self.email.encode("ascii")).hexdigest() + "?d=identicon&r=pg"
         else:
-            return "https://gravatar.com/avatar/placeholder?d=mm"
+            return "https://gravatar.com/avatar/placeholder?d=identicon"
 
     @property
     def stats(self):
--- a/critiquebrainz/data/model/user.py
+++ b/critiquebrainz/data/model/user.py
@@ -97,7 +97,7 @@ class User(db.Model, AdminMixin, DeleteMixin):
         if self.show_gravatar and self.email:
             return "https://gravatar.com/avatar/" + hashlib.md5(self.email.encode("ascii")).hexdigest() + "?d=identicon&r=pg"
         else:
-            return "https://gravatar.com/avatar/" + hashlib.md5(self.id.replace("-", "").encode("ascii")).hexdigest() + "?d=identicon"
+            return "https://gravatar.com/avatar/" + hashlib.md5(self.id.encode("ascii")).hexdigest() + "?d=identicon"
 
     @property
     def stats(self):
--- a/critiquebrainz/data/model/user.py
+++ b/critiquebrainz/data/model/user.py
@@ -97,7 +97,7 @@ class User(db.Model, AdminMixin, DeleteMixin):
         if self.show_gravatar and self.email:
             return "https://gravatar.com/avatar/" + hashlib.md5(self.email.encode("ascii")).hexdigest() + "?d=identicon&r=pg"
         else:
-            return "https://gravatar.com/avatar/placeholder?d=identicon"
+            return "https://gravatar.com/avatar/" + hashlib.md5(self.id.replace("-", "").encode("ascii")).hexdigest() + "?d=identicon"
 
     @property
     def stats(self):
--- a/critiquebrainz/data/model/user.py
+++ b/critiquebrainz/data/model/user.py
@@ -95,9 +95,9 @@ class User(db.Model, AdminMixin, DeleteMixin):
     def avatar(self):
         """Link to user's avatar image."""
         if self.show_gravatar and self.email:
-            return "https://gravatar.com/avatar/" + hashlib.md5(self.email.encode("ascii")).hexdigest() + "?d=identicon&r=pg"
+            return "https://gravatar.com/avatar/" + hashlib.md5(self.email.encode("utf-8")).hexdigest() + "?d=identicon&r=pg"
         else:
-            return "https://gravatar.com/avatar/" + hashlib.md5(self.id.encode("ascii")).hexdigest() + "?d=identicon"
+            return "https://gravatar.com/avatar/" + hashlib.md5(self.id.encode("utf-8")).hexdigest() + "?d=identicon"
 
     @property
     def stats(self):
--- a/critiquebrainz/data/model/user_test.py
+++ b/critiquebrainz/data/model/user_test.py
@@ -148,14 +148,14 @@ class UserTestCase(DataTestCase):
         self.assertEqual(len(user.reviews), 0)
 
     def test_avatar_property(self):
-        user = User(display_name=u'Tester', email=u"example@example.org")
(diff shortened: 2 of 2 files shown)


anshuman73:
Screenshot for reference:

screenshot from 2016-12-08 23-29-58

critiquebrainz/data/model/user.py
```diff
@@ -95,9 +95,9 @@ def reviews(self):
     def avatar(self):
         """Link to user's avatar image."""
         if self.show_gravatar and self.email:
-            return "https://gravatar.com/avatar/" + hashlib.md5(self.email.encode("ascii")).hexdigest() + "?d=mm&r=pg"
+            return "https://gravatar.com/avatar/" + hashlib.md5(self.email.encode("ascii")).hexdigest() + "?d=identicon&r=pg"
         else:
-            return "https://gravatar.com/avatar/placeholder?d=mm"
+            return "https://gravatar.com/avatar/placeholder?d=identicon"
```
gentlecat: Are those going to be the same for all users without email address specified?

anshuman73: yes.

gentlecat: Well, the whole point of using those is to have unique avatars for each user even if they don't have an email address.

anshuman73: It seems none of the available types of images in Gravatar support that. 
Just tried with all those here - https://en.gravatar.com/site/implement/images/
We could however generate random hashes for those who don't have an email id to get unique avatars

gentlecat: Except we want exactly opposite of random. Avatar needs to be unique for each user, but not different every time.

anshuman73: well, it seems it doesn't always work.

gentlecat: That's the hash they use: https://en.wikipedia.org/wiki/MD5

anshuman73: Well, the hack does seem to generate a gravatar that I know isn't the default one

screenshot from 2016-12-09 02-50-13

anshuman73: It works fine alright. I just tested it on mine and yours user id.
This means each user will now have their own gravatar incase they don't have one associated to their email and will get a unique one for them only till the time their user id remains the same (which will)

anshuman73:
@gentlecat I think I need your help in amending the test. How are the user.id stored in the db ? It's showing they are stored as int type ?

critiquebrainz/data/model/user_test.py
```diff
@@ -148,21 +148,21 @@ def test_reviews_property(self):
         self.assertEqual(len(user.reviews), 0)
 
     def test_avatar_property(self):
-        user = User(display_name=u'Tester', email=u"example@example.org")
+        user = User(display_name=u'Tester', email=u"example@example.org", id=u"ea1c2bf3-f246-d47c-f6a4-e5f6e7b8d9e0")
```
gentlecat: Not sure I understand why you specify an `id` there. It's generated when a user is created: https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/data/model/user.py#L16.

gentlecat: If it's to make sure that MD5 hash produced below is correct, you can retrieve it from a `user` after it's added to a database.

critiquebrainz/data/model/user_test.py
```diff
@@ -148,21 +148,21 @@ def test_reviews_property(self):
...
         db.session.add(user)
         db.session.commit()
 
         # By default show_gravatar attribute should be set to False
         self.assertFalse(user.show_gravatar)
         # so avatar property returns generic avatar
-        self.assertEqual(user.avatar, "https://gravatar.com/avatar/placeholder?d=mm")
+        self.assertEqual(user.avatar, "https://gravatar.com/avatar/476c357ea2c081b90bf2a6659d0fbd26?d=identicon")
```
gentlecat: Rather than hard-coding MD5 hash here I'd rather see it calculated based on the user ID.

critiquebrainz/data/model/user_test.py
```diff
@@ -148,21 +148,21 @@ def test_reviews_property(self):
...
 
         # Let's allow to show avatar of this user.
         user.show_gravatar = True
         db.session.commit()
 
         self.assertTrue(user.show_gravatar)
-        self.assertEqual(user.avatar, "https://gravatar.com/avatar/f72c502e0d657f363b5f2dc79dd8ceea?d=mm&r=pg")
+        self.assertEqual(user.avatar, "https://gravatar.com/avatar/f72c502e0d657f363b5f2dc79dd8ceea?d=identicon&r=pg")
```
gentlecat: Same here. I see it was like that before, but let's improve it a bit.

critiquebrainz/data/model/user.py
```diff
@@ -95,9 +95,9 @@ def reviews(self):
     def avatar(self):
         """Link to user's avatar image."""
         if self.show_gravatar and self.email:
-            return "https://gravatar.com/avatar/" + hashlib.md5(self.email.encode("ascii")).hexdigest() + "?d=mm&r=pg"
+            return "https://gravatar.com/avatar/" + hashlib.md5(self.email.encode("ascii")).hexdigest() + "?d=identicon&r=pg"
         else:
-            return "https://gravatar.com/avatar/placeholder?d=mm"
+            return "https://gravatar.com/avatar/" + hashlib.md5(self.id.replace("-", "").encode("ascii")).hexdigest() + "?d=identicon"
```
anshuman73: hmm, no, it works with `-`s as well
I'll change it.

critiquebrainz/data/model/user_test.py
anshuman73: Hmm, didn't realise that. I did it mostly to make sure the tests are hard-coded, just like they were mostly before.
I'll patch it :)

critiquebrainz/data/model/user_test.py
anshuman73: But isn't this hash generated on the basis of the test email id provided here
https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/data/model/user_test.py#L151

anshuman73: What we could do is perhaps generate random email id and then assign that to user and check if the correct hash for it is there in the gravatar url

gentlecat: When you say "random" you probably mean something else. MD5 hash can be calculated from the email address specified when user is created. You can get it from `user.email`.

critiquebrainz/data/model/user.py
```diff
@@ -97,7 +97,7 @@ def avatar(self):
         if self.show_gravatar and self.email:
             return "https://gravatar.com/avatar/" + hashlib.md5(self.email.encode("ascii")).hexdigest() + "?d=identicon&r=pg"
         else:
-            return "https://gravatar.com/avatar/" + hashlib.md5(self.id.replace("-", "").encode("ascii")).hexdigest() + "?d=identicon"
+            return "https://gravatar.com/avatar/" + hashlib.md5(self.id.encode("ascii")).hexdigest() + "?d=identicon"
```
anshuman73: No, it was ascii earlier, so I thought it was best to keep it that way.
I'll change this.

critiquebrainz/data/model/user_test.py
anshuman73: No, i meant generating a random email address.
As far as I understood the code (I probably am wrong), the email address was only assigned to the test user's instance in the line (https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/data/model/user_test.py#L151).
So I thought that we wanted the tests to run on assigned data (like most of the test data is, for example all names, and any sort of test data attributes are assigned to the instance and then tested on their basis, eg name="Tester", display_name="test", etc)

I assume doing user.email will work in the above test ? So should I change it to that ?

PR #70, opened 2016-12-19
CB-128: API usage examples

anshuman73:
Usage examples have only been created for endpoints without Oauth, due to few problems with it

### DIFF
--- a/critiquebrainz/ws/review/views.py
+++ b/critiquebrainz/ws/review/views.py
@@ -23,6 +23,15 @@ def review_entity_handler(review_id):
     :statuscode 404: review not found
 
     :resheader Content-Type: *application/json*
+
+    .. code-block:: bash
+
+       $ curl -X GET https://critiquebrainz.org/ws/1/review/b7575c23-13d5-4adc-ac09-2f55a647d3de
+
+    .. code-block:: json
+
+       {"review":{"created":"Tue, 10 Aug 2010 00:00:00 GMT","edits":0,"entity_id":"03e0a99c-3530-4e64-8f50-6592325c2082","entity_type":"release_group","id":"b7575c23-13d5-4adc-ac09-2f55a647d3de","language":"en","last_updated":"Tue, 10 Aug 2010 00:00:00 GMT","license":{"full_name":"Creative Commons Attribution-NonCommercial-ShareAlike 3.0 Unported","id":"CC BY-NC-SA 3.0","info_url":"https://creativecommons.org/licenses/by-nc-sa/3.0/"},"rating":0,"source":"BBC","source_url":"http://www.bbc.co.uk/music/reviews/3vfd","text":"REVIEW GOES HERE","user":{"created":"Wed, 07 May 2014 14:55:23 GMT","display_name":"Paul Clarke","id":"f5857a65-1eb1-4574-8843-ae6195de16fa","karma":0,"user_type":"Noob"},"votes_negative":0,"votes_positive":0}}
+
     """
     review = Review.query.get_or_404(str(review_id))
     if review.is_hidden:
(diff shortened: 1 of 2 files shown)


critiquebrainz/ws/review/views.py
```diff
@@ -23,6 +23,15 @@ def review_entity_handler(review_id):
     :statuscode 404: review not found
 
     :resheader Content-Type: *application/json*
+
+    .. code-block:: bash
```
gentlecat: You should put these before status codes and error. It should be part of the description.

gentlecat: Also put a subtitle that clearly marks them as examples.

critiquebrainz/ws/review/views.py
```diff
@@ -23,6 +23,15 @@ def review_entity_handler(review_id):
...
+
+       $ curl -X GET https://critiquebrainz.org/ws/1/review/b7575c23-13d5-4adc-ac09-2f55a647d3de
+
+    .. code-block:: json
+
+       {"review":{"created":"Tue, 10 Aug 2010 00:00:00 GMT","edits":0,"entity_id":"03e0a99c-3530-4e64-8f50-6592325c2082","entity_type":"release_group","id":"b7575c23-13d5-4adc-ac09-2f55a647d3de","language":"en","last_updated":"Tue, 10 Aug 2010 00:00:00 GMT","license":{"full_name":"Creative Commons Attribution-NonCommercial-ShareAlike 3.0 Unported","id":"CC BY-NC-SA 3.0","info_url":"https://creativecommons.org/licenses/by-nc-sa/3.0/"},"rating":0,"source":"BBC","source_url":"http://www.bbc.co.uk/music/reviews/3vfd","text":"REVIEW GOES HERE","user":{"created":"Wed, 07 May 2014 14:55:23 GMT","display_name":"Paul Clarke","id":"f5857a65-1eb1-4574-8843-ae6195de16fa","karma":0,"user_type":"Noob"},"votes_negative":0,"votes_positive":0}}
```
gentlecat: This is currently unreadable and needs to be formatted. See https://acousticbrainz.readthedocs.io/api.html#post--api-v1-datasets- for example.

anshuman73:
@gentlecat It seems that even `GET /user/me/reviews` doesn't work. The paste https://dpaste.de/bHic was apprently correct

anshuman73:
Also, as I said earlier, none of the POST requests work right

critiquebrainz/ws/review/views.py
```diff
@@ -196,7 +339,7 @@ def review_post_handler(user):
 
     :reqheader Content-Type: *application/json*
 
-    :json uuid entity_id: UUID of the release group that is being reviewed
+    :json uuid entity_id: UUID of the entity that is being reviewed
```
anshuman73: Wanted confirmation on this as well.
I believe it should be entity as the ID can be for any of the entities (release_group' or 'event')

gentlecat: That is correct.

critiquebrainz/ws/review/views.py
```diff
@@ -267,6 +450,20 @@ def review_vote_put_handler(review_id, user):
 
     **OAuth scope:** vote
 
+    **Request Example:**
+
+    .. code-block:: bash
+
+        $ curl -X PUT -H "Content-type: application/json" -H "Authorization: Bearer <your Oauth token>" -d '{"vote":true}' "https://critiquebrainz.org/ws/1/review/9cb11424-d070-4ac1-8771-a8703ae5cccd/vote"
```
gentlecat: Instead of `your OAuth token` you can be more specific and say `access token`.

gentlecat: You can also split long rows so that they are fully visible and easier to read:

```bash
$ curl "https://critiquebrainz.org/ws/1/review/9cb11424-d070-4ac1-8771-a8703ae5cccd/vote" \
       -X PUT -H "Content-Type: application/json" -d '{"vote":true}' \
       -H "Authorization: Bearer <access token>"
```

anshuman73: Sure, I'll change both of these

PR #71, opened 2016-12-22
CB-211: Fix styling for Incorrect button

anshuman73:
Fix for https://tickets.metabrainz.org/browse/CB-211

### DIFF
--- a/critiquebrainz/frontend/templates/mapping/list.html
+++ b/critiquebrainz/frontend/templates/mapping/list.html
@@ -58,7 +58,7 @@
               {{ artists|join(', ') }}
             </td>
             <td>{{ album.release_date }}</td>
-            <td><a class="btn btn-danger btn-xs" href="{{ url_for('mapping.spotify_report', release_group_id=release_group.id, spotify_id=album.id) }}">{{ _('Incorrect') }}</a></td>
+            <td><a class="btn btn-danger btn-xs" href="{{ url_for('mapping.spotify_report', release_group_id=release_group.id, spotify_id=album.id) }}">{{ _('Report as Incorrect') }}</a></td>
           </tr>
         {% endfor %}
         </tbody>
--- a/critiquebrainz/frontend/static/styles/main.less
+++ b/critiquebrainz/frontend/static/styles/main.less
@@ -37,6 +37,7 @@ body {
     .btn {
       background-color: @turquoise;
       margin-right: 0px;
+      color: @white
     }
   }
 }
--- a/critiquebrainz/frontend/static/styles/main.less
+++ b/critiquebrainz/frontend/static/styles/main.less
@@ -37,7 +37,8 @@ body {
     .btn {
       background-color: @turquoise;
       margin-right: 0px;
-      color: @white
+      color: @white;
+      border-radius: 0;
     }
   }
 }
--- a/critiquebrainz/frontend/static/styles/theme/buttons.less
+++ b/critiquebrainz/frontend/static/styles/theme/buttons.less
@@ -1,5 +1,4 @@
 .btn {
-  border: 0;
   border-radius: 0;
   transition: all 0.2s;
   margin: 0.15em;
@@ -7,8 +6,10 @@
 }
 
 .button-variant(@color; @background:transparent; @border:@background) {
-  color: @color;
+  color: @background;
   background: @background;
+  border-color: @background;
+  background-color: transparent;
   &:hover, &:focus, &:active, &.active, .open .dropdown-toggle& {
     background: lighten(@background, 7%);
   }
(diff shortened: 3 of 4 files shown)


anshuman73:
Current looks - 

screenshot from 2016-12-22 23-41-55

anshuman73:
Alright. I'll mock up changes for them all

uklauer:
I don’t think “Report Incorrect“ is a good phrase. “Report as incorrect”, perhaps.

anshuman73:
@gentlecat 
Would you suggest adopting Bootstrap v4 for incorporating that design ?

gentlecat:
Should be easy to port those buttons from v4 to v3. We have a custom theme files, which you can put those in.

anshuman73:
Alright, I'll just copy the styles then
Anyway porting to v4 would have been difficult. They've shifted to sass instead of less, so that would need ruby :(

anshuman73:
Current looks -

http://imgur.com/a/6q0lk

anshuman73:
Known issues:

1) The buttons which used to have class btn-default (grey coloured buttons) when ported to btn-outline-secondary do not give the same colour and depth.
This sometimes makes it less visible than it should be in the situation

**Possible fix** - Change the colour scheme of the class btn-outline-secondary class. _This means changing the original Bootstrap's css, which may not be suggestible, **thus, permission for the same is required**_

2) The buttons which used to have class btn-primary (orange coloured buttons), when changed to btn-outline-primary turn blue in colour.

**Possible fix** - Change the class btn-outline-primary to btn-outline-warning (which is the same shade of orange as original). _But this means the buttons's classes may not retain their semantic meaning, i.e. describe the button's functionality properly, **thus again, permission for the same is required**_

anshuman73:
Having the base styling of btn made all buttons look extremely small for some reason. Another btn class (https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/frontend/static/styles/theme/boostrap/buttons.less#L9) took care of it better. It also seems to be the default for bootstrap. (just like in the link you mentioned).

Also, are you suggesting to port the new bootstrap's default btn class into ours ?

I'll use more of the Less features to make it easier to read.

I got all the new classes from the bootstrap v4's css files

gentlecat:
> ^ Are you suggesting to port the new bootstrap's default btn class into ours ?

No other way to make them look the same.

anshuman73:
Ported the new btn class to bootstrap v4 using less

Issues mentioned in `https://github.com/metabrainz/critiquebrainz/pull/71/#issuecomment-269098472` still exist. @gentlecat  Any suggestions for the same ?

critiquebrainz/frontend/templates/mapping/list.html
```diff
@@ -58,11 +58,11 @@
               {{ artists|join(', ') }}
             </td>
             <td>{{ album.release_date }}</td>
-            <td><a class="btn btn-danger btn-xs" href="{{ url_for('mapping.spotify_report', release_group_id=release_group.id, spotify_id=album.id) }}">{{ _('Incorrect') }}</a></td>
+            <td><a class="btn btn-outline-danger" href="{{ url_for('mapping.spotify_report', release_group_id=release_group.id, spotify_id=album.id) }}">{{ _('Report as Incorrect') }}</a></td>
```
gentlecat: `btn-xs` is not needed anymore?

gentlecat:
> Issues mentioned in https://github.com/metabrainz/critiquebrainz/pull/71/#issuecomment-269098472 still exist.

Colors can be modified later, if needed.

critiquebrainz/frontend/templates/mapping/list.html
anshuman73: nope, that's the one that gave it the small padding, which needs to be removed

anshuman73:
> Another change that needs to be made (or reverted) is related to -outline modifiers in classes. If we are making all of the buttons on the website with an outline style then its definition becomes redundant and it can be just be integrated with the default btn

wouldn't that make us change the /bootstrap/buttons.less file ? or I should try overriding it with the theme/buttons.less

gentlecat:
As I said in the same comment, you can override values in the default `btn` and other related classes.

anshuman73:
but then all of this porting to bootstrap v4 doesn't really make sense. the only reason we did so was so that we could use the default classes provided by it, and not make our own classes.

anshuman73:
I squashed the commits (and their reverts) for a better commit history

anshuman73:
@gentlecat All good.
The buttons have successfully got the new look from v4.
commit history is now also clean

anshuman73:
Known issue - the Search button on right top corner gets a border which makes it look awkward.

screenshot from 2017-01-01 15-35-48

_**Possible Fix:**_ Remove the `btn-default` class for the search button. It still maintains it's looks. Also fix the `color` attribute in main.less which will change when the `btn-default` class is removed

anshuman73:
@gentlecat done

anshuman73:
I couldn't find any other property that changes the border-radius except for the two that you mentioned

PR #72, opened 2016-12-23
Fix rendering issues in docs

anshuman73:
There are some mergability issues as my local history is all messed up.


anshuman73:
@gentlecat this is good to go from my side. The wrongly rendered docs are live, so you may want to merge this. Let me know if I need to change anything

PR #73, opened 2017-01-04
CB-222: List of moderators

tigranl:


gentlecat:
@brainzbot, ok to test.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,11 @@
+from flask import Blueprint, render_template
+from critiquebrainz.data.model.user import User
+from custom_config import ADMINS
```
gentlecat: There may be ([and are](https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/frontend/__init__.py#L14-L28)) other config files. Some might exist, some might not. So the right thing to do here would be to use the `config` object from [`flask.current_app`](http://flask.pocoo.org/docs/0.12/api/#flask.current_app). You should be able to find examples throughout other CritiqueBrainz modules.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,11 @@
...
+
+moderators_bp = Blueprint('moderators', __name__)
+
+
+@moderators_bp.route('/moderators')
+def page():
+    users = User.query.filter(User.display_name.in_(ADMINS)).all()
```
gentlecat: Database queries shouldn't be built in modules related to views. I know that we still do that in a lot of places, but that was a mistake which makes some things confusing. Apart from this, it would be better not to write any ORM code since we are trying to migrate away from that too.

There is now a new package `critiquebrainz.db` where database access code should go.

tigranl:
Maybe I should use default config?

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,11 @@
+from flask import Blueprint, render_template
+from critiquebrainz.data.model.user import User
+from custom_config import ADMINS
+
+moderators_bp = Blueprint('moderators', __name__)
+
+
+@moderators_bp.route('/moderators')
```
gentlecat: Perhaps this should be just `/` instead since you already add a prefix in the `create_app` function. Otherwise path to this view is `/moderators/moderators`, which is unnecessary.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,11 @@
...
+def page():
```
gentlecat: `list` might be a better name for this function. `page` is not very specific.

critiquebrainz/frontend/templates/moderators/moderators.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'base.html' %}
+
+{% block title %}{{ _('Moderators') }} - CritiqueBrainz{% endblock %}
+
+{% block content %}
+<div>
```
gentlecat: This `div` should be indented. Also keep in mind that we use 2 spaces for indentation in templates.

critiquebrainz/frontend/templates/moderators/moderators.html
```diff
@@ -0,0 +1,11 @@
...
+    {% for item in users %}
+        <a href="{{ url_for('user', user_id=item.musicbrainz_id) }}">{{ item.display_name }}</a>
```
gentlecat: Did you test this?

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/werkzeug/wsgi.py", line 659, in __call__
    return app(environ, start_response)
  File "/usr/local/lib/python3.5/site-packages/flask/app.py", line 2000, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/local/lib/python3.5/site-packages/flask/app.py", line 1991, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/usr/local/lib/python3.5/site-packages/flask/app.py", line 1567, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/usr/local/lib/python3.5/site-packages/flask/app.py", line 1988, in wsgi_app
    response = self.full_dispatch_request()
 ...
  File "/code/critiquebrainz/frontend/templates/base.html", line 46, in block "wrapper"
    {% block content %}
  File "/code/critiquebrainz/frontend/templates/moderators/moderators.html", line 8, in block "content"
    <a href="{{ url_for('user', user_id=item.musicbrainz_id) }}">{{ item.display_name }}</a>
  File "/usr/local/lib/python3.5/site-packages/flask/helpers.py", line 332, in url_for
    return appctx.app.handle_url_build_error(error, endpoint, values)
  File "/usr/local/lib/python3.5/site-packages/flask/app.py", line 1811, in handle_url_build_error
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/usr/local/lib/python3.5/site-packages/flask/helpers.py", line 322, in url_for
    force_external=external)
  File "/usr/local/lib/python3.5/site-packages/werkzeug/routing.py", line 1758, in build
    raise BuildError(endpoint, values, method, self)
werkzeug.routing.BuildError: Could not build url for endpoint 'user' with values ['user_id']. Did you mean 'user.info' instead?
```

gentlecat:
Also, can you assign yourself to a ticket that you are working on (https://tickets.metabrainz.org/browse/CB-222)? This is to make sure that someone else doesn't start working on it not knowing that there's already your work.

critiquebrainz/frontend/views/moderators.py
tigranl: Why are you migrating from ORM?

gentlecat: https://tickets.metabrainz.org/browse/CB-230

critiquebrainz/frontend/views/moderators.py
tigranl: I think, `moderators` would be a better choice, since `list` is built-in name.

gentlecat: Good point. 👍

tigranl:
With this code https://gist.github.com/tigranl/faea820a44b8425bf215bb57b4edf2fb I am getting `sqlalchemy.exc.ProgrammingError: (ProgrammingError) column "musicbrainz_id" does not exist LINE 1: SELECT musicbrainz_id FROM user WHERE display_name IN ('tigr... ^ "SELECT musicbrainz_id FROM user WHERE display_name IN ('tigrank', 'test')" {}` It seems like db isn't initialized, but with ORM approach it actually works, so I am confused.

anshuman73:
@tigranl 
Did you initialise the db after you started the docker ?
https://critiquebrainz.readthedocs.io/intro.html#startup

tigranl:
@anshuman73 yes

gentlecat:
Try putting `user` in quotes: `"user"`. It's one of the keywords https://www.postgresql.org/docs/current/static/sql-keywords-appendix.html.

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,17 @@
+from critiquebrainz import db
+from critiquebrainz.db import exceptions as db_exceptions
+import sqlalchemy
+
+
+def getusers_id(usernames):
```
gentlecat: Name and documentation of this function are misleading. You don't return just an ID, but all the columns.

I would suggest to alter this function slightly. We can have more general `get_many_by_mb_username`, which would return a list of users (as dictionaries).

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,17 @@
...
+    """Get MusicBrainz ids of users
+    Args:
+        usernames (list): MusicBrainz usernames.
+    """
+    with db.engine.connect() as connection:
+        req = sqlalchemy.text("""SELECT * FROM "user" WHERE display_name IN :users""")
+        result = connection.execute(req, users=tuple(usernames))
```
gentlecat: Minor thing, but it would be better to keep all queries defined consistently:

```python
result = connection.execute("""
    SELECT *
      FROM "user"
     WHERE display_name IN :users
""", {"users": tuple(usernames)})
```

See https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/db/vote.py#L69-L75 for example.

gentlecat: Also make sure to explicitly define which columns are returned there instead of just `SELECT *`. Refer to the example I linked.

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,17 @@
...
+        if not result:
+            return db_exceptions.NoDataFoundException("Can't find users")
```
gentlecat: I don't see much point in having this exception. What if we want to find users that exist, given a list of MusicBrainz usernames, and just ignore missing ones? This is probably what we want for the moderator list feature, considering that it's possible that moderator might not have an associated user in the CritiqueBrainz database.

I would suggest two options:

1. Add an `ignore_missing` argument. If it's set to True then return a list with users that were found and skip ones that are missing. If it's set to `False`, then check if all users passed in the `usernames` argument were found, and if not then raise an exception (ideally with a list of missing users).
2. Ignore missing users by default, but document this behaviour.

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,17 @@
...
+        row = result.fetchall()
+        return row
```
gentlecat: Two rows above can be combined, though you'll probably end up rewriting this part anyway.

critiquebrainz/frontend/templates/moderators/moderators.html
```diff
@@ -0,0 +1,18 @@
+{% extends 'base.html' %}
+
+{% block title %}{{ _('Moderators') }} - CritiqueBrainz{% endblock %}
+
+{% block content %}
+<div class="container">
+  <h1>CritiqueBrainz Moderators</h1>
```
gentlecat: `CritiqueBrainz` can be dropped, since it's pretty obvious where we are.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,19 @@
...
+from critiquebrainz.db import users as db_users
+
+moderators_bp = Blueprint('moderators', __name__)
+
+
+@moderators_bp.route('/')
+def moderators():
+    admins = current_app.config['ADMINS']
+    db_admins = db_users.getusers_id(admins)
+    db_admins = [dict(_) for _ in db_admins]
```
gentlecat: I would suggest to do this conversion in the database function and return dictionaries from there. See my other comment in that part.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,19 @@
...
+    user_names = [_['display_name'] for _ in db_admins]
+    no_auth = list(set(admins) - set(user_names))
```
gentlecat: What does `no_auth` mean?

critiquebrainz/db/users.py
tigranl: Yeah, at the beginning it was returning only ids, I changed it and forgot to update.

tigranl:
Also, please, have a look at my PEP8 (https://github.com/metabrainz/critiquebrainz/pull/75)  PR.

tigranl:
By the way, how to access users' avatars?

gentlecat:
`User` model has this method https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/data/model/user.py#L95-L100. You can create a similar function and add an `avatar_url` as a part of returned dictionaries.

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,35 @@
...
+        and avatar_url (Gravatar url)
+
+    Ignores missing users
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+          SELECT "id", "display_name", "email", "created",
+                  "musicbrainz_id", "show_gravatar"
+          FROM "user"
+          WHERE display_name IN :users"""), users=tuple(usernames))
```
gentlecat: It would also be better to pass arguments in a dictionary, for consistency sake.

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,35 @@
...
+        row = result.fetchall()
+        users = [dict(_) for _ in row]
```
gentlecat: Row above can be combined with this one. Also, with list comprehensions we typically use first letter of a variable (`r` for `row`) instead of `_`.

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,35 @@
...
+        for item in users:
+            item.update({'avatar_url': avatar(item)})
```
gentlecat: This can be just `item["avatar_url"] = avatar(item)`.

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+          SELECT "id", "display_name", "email", "created",
+                  "musicbrainz_id", "show_gravatar"
+          FROM "user"
+          WHERE display_name IN :users"""), users=tuple(usernames))
+        row = result.fetchall()
+        users = [dict(_) for _ in row]
+        for item in users:
```
gentlecat: `item` -> `user`

critiquebrainz/db/users_test.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class UserTestCase(DataTestCase):
+    def setUp(self):
+        super(UserTestCase, self).setUp()
+        self.user1 = User.get_or_create("test", musicbrainz_id='0')
+        self.user2 = User.get_or_create("test1", musicbrainz_id='1')
+        self.admins = [self.user1.display_name, self.user2.display_name]
+
+    def test_response_type(self):
```
gentlecat: Typically in a test case each method tests a particular set of features in a given function. The way to name them is the following: `test_get_many_by_mb_username`. Can also append clarification of particular feature the test is for, but isn't necessary if there's just one test for a function.

critiquebrainz/db/users_test.py
```diff
@@ -0,0 +1,18 @@
...
+        self.assertEqual(type(users.get_many_by_mb_username(self.admins)), list)
```
gentlecat: This can be replace by assertion that list with values returned from `get_many_by_mb_username` exactly matches another. Checking type in this case is not enough, values need to be checked too.

Make sure to check with different sets of arguments. For example, with `show_gravatar` set to `True` and `False` (you'll need to modify users that are created in `setUp` to make sure that one of the doesn't allow their Gravatar to be displayed).

critiquebrainz/db/users_test.py
```diff
@@ -0,0 +1,18 @@
...
+from critiquebrainz.data.model.user import User
+from critiquebrainz.db import users
+
+
+class UserTestCase(DataTestCase):
+    def setUp(self):
+        super(UserTestCase, self).setUp()
+        self.user1 = User.get_or_create("test", musicbrainz_id='0')
+        self.user2 = User.get_or_create("test1", musicbrainz_id='1')
+        self.admins = [self.user1.display_name, self.user2.display_name]
```
gentlecat: Admins list probably shouldn't be in the `setUp` method since it's not going to be relevant to all function that will be in the `critiquebrainz.db.users` module. It just happens to be the case now. So I'd move that into the method that tests admin-related functions.

User definitions can be left there.

critiquebrainz/db/users_test.py
gentlecat: Also, why `admins` and not just `users`. It shouldn't matter here.

critiquebrainz/db/users_test.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def test_response_type(self):
+        self.assertEqual(type(users.get_many_by_mb_username(self.admins)), list)
+
+    def test_exception(self):
```
gentlecat: This name is also too general.

critiquebrainz/frontend/templates/moderators/moderators.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'base.html' %}
+
+{% block title %}{{ _('Moderators') }} - CritiqueBrainz{% endblock %}
+
+{% block content %}
+<div class="container">
+  <h1>Moderators</h1>
```
gentlecat: This string needs to be wrapped into `_(...)` to be translatable.

critiquebrainz/frontend/templates/moderators/moderators.html
```diff
@@ -0,0 +1,20 @@
...
+  <div id="list">
+    {% for item in db_admins %}
+      {% if item.id %}
+        <img src="{{ item.avatar_url }}" width="25px" height="25px">
+        <a href="{{ url_for('user.info', user_id=item.id) }}" style="margin-right: 3px"> {{ item.display_name }} </a>
```
gentlecat: `> {{ item.display_name }} <` -> `>{{ item.display_name }}<`

critiquebrainz/frontend/views/test/test_moderators.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class ModeratorsTestCase(FrontendTestCase):
+    def test_template_render(self):
+        try:
+            response = self.client.get('/moderators/')
+            self.assert200(response)
+            for admin in current_app.config['ADMINS']:
+                self.assertIn(admin, str(response.data.decode('utf-8')))
+        except:
```
gentlecat: This makes the test pointless. :)

critiquebrainz/frontend/views/test/test_moderators.py
```diff
@@ -0,0 +1,13 @@
+from flask import current_app
+from critiquebrainz.frontend.testing import FrontendTestCase
+
+
+class ModeratorsTestCase(FrontendTestCase):
+    def test_template_render(self):
```
gentlecat: `test_template_render` -> `test_moderators`

So `test_<name_of_the_view_function>`.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,20 @@
...
+from critiquebrainz.db import users as db_users
+
+moderators_bp = Blueprint('moderators', __name__)
+
+
+@moderators_bp.route('/')
+def moderators():
+    admins = current_app.config['ADMINS']
+    db_admins = db_users.get_many_by_mb_username(admins)
+    user_names = [item['display_name'] for item in db_admins]
```
gentlecat: `item` -> `admin`

gentlecat: It would be easier to read this function if the same thing is named consistently. In this case `admin` and `user` mean the same.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,20 @@
...
+    # list of MB users without CritiqueBrainz account
+    no_auth = list(set(admins) - set(user_names))
```
gentlecat: `no_auth` is a bit unclear. Perhaps `missing_admins`?

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,20 @@
...
+    if user_names != admins:
```
gentlecat: What does this do?

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,20 @@
...
+        for item in no_auth:  # Add them to the main list of admins
+            user = {'display_name': item, 'musicbrainz_id': None,
+                    'avatar_url': 'https://gravatar.com/avatar/'}
+            db_admins.append(user)
+    return render_template('moderators/moderators.html', db_admins=db_admins)
```
gentlecat: Why `db_admins` and not just `admins` or better`moderators`. In the template it shouldn't matter where the data comes from.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,20 @@
...
+def moderators():
+    admins = current_app.config['ADMINS']
+    db_admins = db_users.get_many_by_mb_username(admins)
+    user_names = [item['display_name'] for item in db_admins]
+    # list of MB users without CritiqueBrainz account
+    no_auth = list(set(admins) - set(user_names))
+    if user_names != admins:
+        for item in no_auth:  # Add them to the main list of admins
+            user = {'display_name': item, 'musicbrainz_id': None,
+                    'avatar_url': 'https://gravatar.com/avatar/'}
```
gentlecat:
```python
user = {
    'display_name': item,
    'musicbrainz_id': None,
    'avatar_url': 'https://gravatar.com/avatar/',  # Note the trailing comma
}
```
or
```python
user = {'display_name': item, 'musicbrainz_id': None, 'avatar_url': 'https://gravatar.com/avatar/'}
```

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,20 @@
...
+            db_admins.append(user)
```
gentlecat: This row can be combined with the definition above.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,20 @@
...
+
+@moderators_bp.route('/')
+def moderators():
+    admins = current_app.config['ADMINS']
+    db_admins = db_users.get_many_by_mb_username(admins)
+    user_names = [item['display_name'] for item in db_admins]
+    # list of MB users without CritiqueBrainz account
+    no_auth = list(set(admins) - set(user_names))
+    if user_names != admins:
+        for item in no_auth:  # Add them to the main list of admins
```
gentlecat: Looking at this it's completely unclear what this loop iterates over. Something this is better:
```python
for user in missing_users:
    ...
```

critiquebrainz/frontend/views/moderators.py
gentlecat: Instead of referencing `https://gravatar.com/avatar/` this can be more consistent with the rest of the website, which uses `identicon` as the default avatar type. Hash can be generated from a MusicBrainz username.

critiquebrainz/frontend/templates/moderators/moderators.html
gentlecat: And MusicBrainz username should be displayed instead.

critiquebrainz/frontend/templates/moderators/moderators.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'base.html' %}
+
+{% block title %}{{ _('Moderators') }} - CritiqueBrainz{% endblock %}
+
+{% block content %}
+<div class="container">
+  <h1>Moderators</h1>
+  <div id="list">
+    {% for item in db_admins %}
```
gentlecat: `{% for admin in db_admins %}`

gentlecat: Better settle for one way of calling users in a template to avoid confusion. I'd suggest `moderator`/`moderators` since that's what they are called publicly.

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,35 @@
...
+from critiquebrainz import db
+
+
+def avatar(item):
+    url = "https://gravatar.com/avatar/{0}{1}"
+    email = item['email']
+    if email and item['show_gravatar']:
+        return url.format(md5(email.encode('utf-8')).hexdigest(), "?")
+    else:
+        return url
```
tigranl: @gentlecat why is it hashing id? Also, I have some changes for that function, see my previous message.

critiquebrainz/frontend/views/test/test_moderators.py
tigranl: I have a problem with this one. If `ADMINS` list is empty it will raise an exception, how do I handle this?

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,35 @@
...
+        All columns of "user" table (list of dictionaries)
+        and avatar_url (Gravatar url)
+
+    Ignores missing users
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+          SELECT "id", "display_name", "email", "created",
+                  "musicbrainz_id", "show_gravatar"
+          FROM "user"
```
tigranl: What about `SELECT *`? You may implement new values in `user` model, and it would be much easier to access in future.

gentlecat: It's more about avoiding implicit kind of query. If someone looks just at that function they might not know what columns are in the `user` table. And in some cases not all columns are necessary.

critiquebrainz/frontend/views/test/test_moderators.py
gentlecat: What exception and where?

tigranl: Since I pass empty list to `get_many_by_mb_username` it raises `sqlalchemy.exc.ProgrammingError`

critiquebrainz/db/users.py
anshuman73: @tigranl , 
gravatars generates images based on md5 hashes provided to it. http://en.gravatar.com/site/implement/images/

And we required a fix where people who don't provide email ids/ don't allow their gravatar to be shown should also get an image displayed. As their IDs were unique, we used it to generate the new hashes.
https://github.com/metabrainz/critiquebrainz/pull/67#discussion_r91604203

tigranl: @anshuman73, I got it, thanks.

critiquebrainz/db/users_test.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def test_avatar(self):
+        user3 = {'display_name': 'test2',
+                 'id': '3',
+                 'email': None,
+                 'show_gravatar:': False,
+                 }
+        gravatar_url = "https://gravatar.com/avatar/{0}{1}"
+        test_link = gravatar_url.format(md5(user3['id'].encode('utf-8'))
+                                        .hexdigest(), "?d=identicon")
```
gentlecat: URL shouldn't be generated here. It should just be a hardcoded string that you get as a result when passing a particular set of parameters.

critiquebrainz/db/users_test.py
```diff
@@ -0,0 +1,30 @@
...
+        self.user1 = User.get_or_create("test", musicbrainz_id='0')
+        self.user2 = User.get_or_create("test1", musicbrainz_id='1')
+
+    def test_get_many_by_mb_username(self):
+        self.users = [self.user1.display_name, self.user2.display_name]
+        self.assertEqual(type(users.get_many_by_mb_username(self.users)), list)
+        self.assertRaises(DatabaseError, users.get_many_by_mb_username, [])
+
+    def test_avatar(self):
+        user3 = {'display_name': 'test2',
```
gentlecat: Why `3`? That seems unnecessary at this scope.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,25 @@
...
+def moderators():
+    admins = current_app.config['ADMINS']
+    moderators = db_users.get_many_by_mb_username(admins)
+    config_list = [moderator['display_name'] for moderator in moderators]
+    # list of MB users without CritiqueBrainz account
+    missing_users = list(set(admins) - set(config_list))
+    if config_list != admins:
+        for user in missing_users:  # Add them to the main list of admins
+            url = "https://gravatar.com/avatar/{0}{1}" \
+                .format(md5(user.encode('utf-8')).hexdigest(), "?d=identicon")
```
gentlecat: The problem is that now the way avatar URLs are generated is defined in multiple places. This means any modifications to that functionality will now become more complex. Can you reuse the function that generates avatars in the `critiquebrainz.db.user` module?

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,40 @@
...
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+          SELECT id,
+                  display_name,
+                  email,
+                  created,
+                  musicbrainz_id,
+                  show_gravatar
+          FROM "user"
+          WHERE display_name IN :users"""), {'users': tuple(usernames)})
```
gentlecat: https://github.com/metabrainz/critiquebrainz/pull/73#discussion_r95170374. And please adjust the indentation of column names (they are off by one).

critiquebrainz/db/users_test.py
```diff
@@ -0,0 +1,30 @@
...
+
+class UserTestCase(DataTestCase):
+    def setUp(self):
+        super(UserTestCase, self).setUp()
+        self.user1 = User.get_or_create("test", musicbrainz_id='0')
+        self.user2 = User.get_or_create("test1", musicbrainz_id='1')
+
+    def test_get_many_by_mb_username(self):
+        self.users = [self.user1.display_name, self.user2.display_name]
+        self.assertEqual(type(users.get_many_by_mb_username(self.users)), list)
```
gentlecat: https://github.com/metabrainz/critiquebrainz/pull/73#discussion_r95174533

critiquebrainz/frontend/views/test/test_moderators.py
gentlecat: This needs to be fixed then. What output do you expect when the list is empty?

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,25 @@
...
+moderators_bp = Blueprint('moderators', __name__)
+
+
+@moderators_bp.route('/')
+def moderators():
+    admins = current_app.config['ADMINS']
+    moderators = db_users.get_many_by_mb_username(admins)
+    config_list = [moderator['display_name'] for moderator in moderators]
+    # list of MB users without CritiqueBrainz account
+    missing_users = list(set(admins) - set(config_list))
```
gentlecat: https://github.com/metabrainz/critiquebrainz/pull/73#discussion_r95177825. In this function alone you have `admins`, `moderators`, and `users` which all mean exactly the same thing. Perhaps the config variable should be named `MODERATORS` instead of `ADMINS`, but that's out of the scope of this issue. You can just do `moderators = current_app.config['ADMINS']` and be done with it.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,25 @@
...
+    if config_list != admins:
+        for user in missing_users:  # Add them to the main list of admins
+            url = "https://gravatar.com/avatar/{0}{1}" \
+                .format(md5(user.encode('utf-8')).hexdigest(), "?d=identicon")
+            user = {'display_name': user,
+                    'musicbrainz_id': None,
+                    'avatar_url': url,
+                    }
+            moderators.append(user)
```
gentlecat: https://github.com/metabrainz/critiquebrainz/pull/73#discussion_r95178925

critiquebrainz/frontend/views/test/test_moderators.py
tigranl: I am not sure. Maybe I should check for an empty list before it is used in sql request?

critiquebrainz/db/users_test.py
tigranl: The string would be too long, don't you think?

critiquebrainz/frontend/views/moderators.py
tigranl: No, I can't change change it that way, because I need to use `admins` list to compare it with information on users from db, and find out which users don't have CritiqueBrainz account.

gentlecat: I'm talking about naming of the variables.

Here's how I would write that view:
```python
@moderators_bp.route('/')
def moderators():
    mb_usernames = set(current_app.config['ADMINS'])
    mods_data = []
    mb_usernames_found = set()
    for mod in db_users.get_many_by_mb_username(mb_usernames):
        mods_data.append({
            'musicbrainz_id': mod['musicbrainz_id'],
            'avatar_url': get_gravatar_url(
                use_email=mod['show_gravatar'],
                email=mod['email'],
                alt=mod['id'],
            ),
        })
        mb_usernames_found.add(mod['musicbrainz_id'])
    for username in mb_usernames - mb_usernames_found:  # Moderators without a CB account
        mods_data.append({
            'musicbrainz_id': username,
            'avatar_url': get_gravatar_url(alt=username),
        })
    return render_template('moderators/moderators.html', moderators=mods_data)
```

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,44 @@
...
+    Args:
+        usernames: A list of MusicBrainz usernames.
+    Returns:
+        All columns of 'user' table (list of dictionaries)
+        and avatar_url (Gravatar url).
+        If the 'usernames' variable is an empty list function returns it back.
+    """
+
+    with db.engine.connect() as connection:
+        if usernames:
```
gentlecat: It's better to do
```python
if not usernames:
    return []
```

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,44 @@
+from hashlib import md5
+import sqlalchemy
+from critiquebrainz import db
+
+
+def avatar(user):
```
gentlecat: What's a `user`? If you want to implement it like that, then the argument must be documented. Don't make people read the function to guess the type and structure you require.

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,44 @@
...
+        if usernames:
+            result = connection.execute(sqlalchemy.text("""
+              SELECT id,
+                      display_name,
+                      email,
+                      created,
+                      musicbrainz_id,
+                      show_gravatar
+              FROM "user"
+              WHERE display_name IN :users"""), {'users': tuple(usernames)})
```
gentlecat:
```python
result = connection.execute(sqlalchemy.text("""
    SELECT id,
           display_name,
           email,
           created,
           musicbrainz_id,
           show_gravatar
      FROM "user"
     WHERE display_name IN :users
"""), {
    'users': tuple(usernames),
})
```

gentlecat: And it would be better to make the username comparison case-insensetive, since they currently can be used like that. You can do something like `WHERE musicbrainz_id ILIKE ANY(...)`.

critiquebrainz/db/users_test.py
```diff
@@ -0,0 +1,32 @@
...
+    def setUp(self):
+        super(UserTestCase, self).setUp()
+        self.user1 = User.get_or_create("test", musicbrainz_id='0')
+        self.user2 = User.get_or_create("test1", musicbrainz_id='1')
+
+    def test_get_many_by_mb_username(self):
+        users = [(self.user1.display_name, self.user1.musicbrainz_id),
+                 (self.user2.display_name, self.user2.musicbrainz_id)]
+        dbresults = get_many_by_mb_username([self.user1.display_name,
+                                             self.user2.display_name])
```
gentlecat: `display_name` is not the same thing as `musicbrainz_id` (MusicBrainz username).

critiquebrainz/db/users.py
tigranl: I adjusted indentation of column, I don't know why it's changed in github. Maybe need to set up my Pycharm for sql.

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    url = "https://gravatar.com/avatar/{0}{1}"
+    if type(user) is dict:
+        if user['email'] and user['show_gravatar']:
+            return url.format(md5(user['email'].encode('utf-8')).hexdigest(),
+                              "?d=identicon&r=pg")
+        else:
+            return url.format(md5(str(user['id']).encode('utf-8')).hexdigest(),
+                              "?d=identicon")
+    return url.format(md5(user.encode('utf-8')).hexdigest(), "?d=identicon")
```
gentlecat: Notice how `url.format(md5(<...>.encode('utf-8')).hexdigest(), "?d=identicon&r=pg")` is defined three times and pretty much identical. What if you just select a string to be hashed and then build a URL at the end?

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,52 @@
+from hashlib import md5
+import sqlalchemy
+from critiquebrainz import db
+
+
+def avatar(user):
+    """Args:
+        user: user object from db (dict) or username string
```
gentlecat: While `dict` is technically an object, it might be confusing to people. Just say that it requires a dictionary with user data. Also specify exactly which data is required.

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,52 @@
...
+    Args:
+        usernames: A list of MusicBrainz usernames.
+    Returns:
+        All columns of 'user' table (list of dictionaries)
+        and avatar_url (Gravatar url).
+        If the 'usernames' variable is an empty list function returns it back.
+    """
+
+    with db.engine.connect() as connection:
+        if not usernames:
```
gentlecat: Does this need to be inside the `with` block?

critiquebrainz/db/users_test.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def test_avatar(self):
+        user = {'display_name': 'test2',
+                'id': '3',
+                'email': None,
+                'show_gravatar:': False,
+                }
+        gravatar_url = "https://gravatar.com/avatar/{0}{1}"
+        test_link = gravatar_url.format(md5(user['id'].encode('utf-8'))
+                                        .hexdigest(), "?d=identicon")
```
gentlecat: You are still hashing a string here. So this is pretty much the same as the function implementation, which makes the test kind of pointless. Ideally there would be just a hardcoded string with a result (URL) that you expect to get when you pass a specific set of arguments.

gentlecat: And what's the purpose of having `?d=identicon` string injected with the format function instead of having it there in the first place?

critiquebrainz/frontend/templates/moderators/moderators.html
```diff
@@ -0,0 +1,20 @@
...
+{% block content %}
+<div class="container">
+  <h1>{{ _('Moderators') }}</h1>
+  <div id="list" style="margin-top: 30px">
+    {% for moderator in moderators %}
+      {% if moderator.id %}
+        <img src="{{ moderator.avatar_url }}" width="25px" height="25px">
+        <a href="{{ url_for('user.info', user_id=moderator.id) }}" style="margin-right: 3px">{{ moderator.display_name }}</a>
+      {% else %}
+        <img src="{{ moderator.avatar_url }}" width="25px" height="25px">
```
gentlecat: This is the same in both. Maybe it can be outside of this conditional?

critiquebrainz/frontend/templates/moderators/moderators.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'base.html' %}
+
+{% block title %}{{ _('Moderators') }} - CritiqueBrainz{% endblock %}
+
+{% block content %}
+<div class="container">
+  <h1>{{ _('Moderators') }}</h1>
+  <div id="list" style="margin-top: 30px">
```
gentlecat: Styling should go into the Less files. This includes `width` and `height` that are defined in some places below.

critiquebrainz/frontend/views/moderators.py
```diff
@@ -0,0 +1,22 @@
...
+@moderators_bp.route('/')
+def moderators():
+    admins = current_app.config['ADMINS']
+    moderators = db_users.get_many_by_mb_username(admins)
+    usernames = [moderator['display_name'] for moderator in moderators]
+    # list of MB users without CritiqueBrainz account
+    missing_users = list(set(admins) - set(usernames))
+    if missing_users:
+        for user in missing_users:  # Add them to the main list of admins
+            moderators.append({'display_name': None,
```
gentlecat: If `display_name` is never used, what's the reason for inserting it here?

critiquebrainz/frontend/templates/moderators/moderators.html
tigranl: But it's just a few lines,  do you think we need to have a less file for it?

gentlecat: Few lines here, few lines there, and it becomes a mess. We already have a place where custom styling for pages must go: https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/frontend/static/styles/main.less.

critiquebrainz/frontend/templates/moderators/moderators.html
tigranl: @Gentlecat, In that case, it will duplicate some users.

critiquebrainz/frontend/templates/moderators/moderators.html
```diff
@@ -5,14 +5,14 @@
 {% block content %}
 <div class="container">
```
gentlecat: Other pages have all style items wrapped into a block with a unique ID. This is done to prevent conflicts.

You can add `id="moderators-list"`, for example, and wrap all the items into that id:
```less
#moderators-list {
  ...
}
```

critiquebrainz/frontend/static/styles/main.less
```diff
@@ -490,3 +490,22 @@ a#edit-review { margin-top: 20px; }
...
+}
+
+#list a {
+  margin-right: 3px;
+}
+
+#list p {
+  display: inline-block;
+  margin-right: 3px;
+}
```
gentlecat: This can be simplified:
```less
// Moderator page
#list {
  margin-top: 30px;

  img {
    width: 25px;
    height: 25px;
  }

  a {
    margin-right: 3px;
  }

  p {
    display: inline-block;
    margin-right: 3px;
  }
}
```

gentlecat: `list` ID is a bit too generic though, I suggested an alternative in another comment.

critiquebrainz/frontend/static/styles/main.less
```diff
@@ -490,3 +490,22 @@ a#edit-review { margin-top: 20px; }
     margin-top: 25px;
   }
 }
+
+// Moderator page
+
+#list {
+  margin-top: 30px;
+}
+#list img {
```
gentlecat: We already have a class for an avatar, which can be used here. https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/frontend/static/styles/main.less#L64-L70

critiquebrainz/db/users.py
```diff
@@ -5,19 +5,19 @@
 
 def avatar(user):
     """Args:
-        user: user object from db (dict) or username string
+        user: user data (dict or str)
     Returns:
         URL to gravatar image
     """
     url = "https://gravatar.com/avatar/{0}{1}"
+    link = (user, "?d=identicon")
```
gentlecat: I don't understand why you need to inject `?d=identicon`. It can just be `?d=identicon&r=pg` everywhere, there's no need to use `format` for it.

critiquebrainz/db/users_test.py
```diff
@@ -23,8 +22,6 @@ def test_avatar(self):
                 'email': None,
                 'show_gravatar:': False,
                 }
-        gravatar_url = "https://gravatar.com/avatar/{0}{1}"
-        test_link = gravatar_url.format(md5(user['id'].encode('utf-8'))
-                                        .hexdigest(), "?d=identicon")
+        gravatar_url = "https://gravatar.com/avatar/eccbc87e4b5ce2fe28308fd9f2a7baf3?d=identicon"
         link = avatar(user)
-        self.assertEqual(test_link, link)
+        self.assertEqual(gravatar_url, link)
```
gentlecat: Nice. Now you can get rid of two variables:
```python
self.assertEqual(
    avatar(user),
    "https://gravatar.com/avatar/eccbc87e4b5ce2fe28308fd9f2a7baf3?d=identicon"
)
```

gentlecat:
> In that case, it will duplicate some users.

Why would it do that?

critiquebrainz/db/users.py
```diff
@@ -0,0 +1,51 @@
+from hashlib import md5
+import sqlalchemy
+from critiquebrainz import db
+
+
+def avatar(user):
+    """Args:
+        user: user data (dict or str)
```
gentlecat: This is still not enough. What is user data? If you want to implement it like that, then the argument must be documented. Don't make people read the function to guess the type and structure you require.

Here's an example:
```python
"""Generates a Gravatar image URL based on user data.

Args:
    user: Dictionary with user data: {
            "id": <user ID (int)>,
            "show_gravatar": <wether this user wants their email address used to get Gravatar (bool)>
            "email": <email address (string)>,
        }

Returns:
    Generated URL of the Gravatar image for this user.
```

Said this before: https://github.com/metabrainz/critiquebrainz/pull/73#discussion_r96002699:

> Also specify exactly which data is required.

*I don't think it's a good idea to accept different types of `user` in this case. Settle on one.*

gentlecat: And at this point I don't see why there can't be just 3 arguments:
* id
* show_gravatar
* email

instead of the same data wrapped into a dictionary.

Again, suggested before in https://github.com/metabrainz/critiquebrainz/pull/73#discussion_r95694397.

tigranl: Because it was meant to make a universal function for db, so there is no need to write another one in future. I don't see anything bad in this implementation.

tigranl:
Sorry for brackets, it's just my bad habit.

critiquebrainz/db/users.py
gentlecat: Well, it's also used outside its own module. Why would you need to write another one?

gentlecat:
Sorry you didn't have enough time to finish this. I'll merge what you currently have and clean up afterwards. Thanks a lot for your work!

PR #74, opened 2017-01-06
Added import of gettext in views/release.py

anshuman73:
It seems that the gettext function wasn't even being imported in that script. Fixed


anshuman73:
Hmm, I can't even see these commits when I try to rebase.

anshuman73:
hehe, fixed !

PR #75, opened 2017-01-06
PEP8

tigranl:
I have noticed some issues with PEP8, so I would like to offer a small fix.

### DIFF
--- a/critiquebrainz/data/model/user.py
+++ b/critiquebrainz/data/model/user.py
@@ -13,21 +13,26 @@ import hashlib
 class User(db.Model, AdminMixin, DeleteMixin):
     __tablename__ = 'user'
 
-    id = db.Column(UUID, primary_key=True, server_default=db.text('uuid_generate_v4()'))
+    id = db.Column(UUID, primary_key=True, server_default=
+                            db.text('uuid_generate_v4()'))
     display_name = db.Column(db.Unicode, nullable=False)
     email = db.Column(db.Unicode)
     created = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
     musicbrainz_id = db.Column(db.Unicode, unique=True)
-    show_gravatar = db.Column(db.Boolean, nullable=False, server_default="False")
-    is_blocked = db.Column(db.Boolean, nullable=False, server_default="False")
+    show_gravatar = db.Column(db.Boolean, nullable=False,
+                              server_default="False")
+    is_blocked = db.Column(db.Boolean, nullable=False,
+                           server_default="False")
 
     spam_reports = db.relationship('SpamReport', cascade='delete', backref='user')
     clients = db.relationship('OAuthClient', cascade='delete', backref='user')
     grants = db.relationship('OAuthGrant', cascade='delete', backref='user')
     tokens = db.relationship('OAuthToken', cascade='delete', backref='user')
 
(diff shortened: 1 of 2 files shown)


critiquebrainz/data/model/user.py
```diff
@@ -13,21 +13,26 @@
 class User(db.Model, AdminMixin, DeleteMixin):
     __tablename__ = 'user'
 
-    id = db.Column(UUID, primary_key=True, server_default=db.text('uuid_generate_v4()'))
+    id = db.Column(UUID, primary_key=True, server_default=
+                            db.text('uuid_generate_v4()'))
```
Freso: Probably better to move `server_default=` to the new line as well.

Freso: Also, mind indentation. Current one seems a bit arbitrary.

critiquebrainz/data/model/user.py
```diff
@@ -111,9 +112,12 @@ def stats(self):
             votes_this_month=self.votes_since_count(date(today.year, today.month, 1)))
 
     def _reviews_since(self, date):
-        rev_q = db.session.query(Revision.review_id, db.func.min(Revision.timestamp).label('creation_time'))\
-            .group_by(Revision.review_id).subquery('time')
-        return self._reviews.outerjoin(rev_q, Review.id == rev_q.c.review_id).filter(rev_q.c.creation_time >= date)
+        rev_q = db.session.query(Revision.review_id,
+                                 db.func.min(Revision.timestamp).
+                                 label('creation_time')). \
```
Freso: Shouldn't the trailing `.` go at the beginning of the next line? That's how you've done elsewhere AFAICT.

manage.py
```diff
@@ -117,31 +122,43 @@ def init_db(skip_create_db=False):
...
     """
     hostname, port, db, username, password = data_utils.explode_db_uri(db_uri)
     if hostname not in ['localhost', '127.0.0.1']:
         raise Exception('Cannot configure a remote database')
 
     # Checking if user already exists
-    retv = subprocess.check_output('sudo -u postgres psql -t -A -c "SELECT COUNT(*) FROM pg_user WHERE usename = \'%s\';"' % username, shell=True)
+    retv = subprocess.check_output('sudo -u postgres psql -t -A -c '
+                                   '"SELECT COUNT(*) FROM pg_user WHERE '
+                                   'usename = \'%s\';"' % username, shell=True)
```
Freso: Is `usename` here a typo? (Yes, I realise that the typo also exists in the source, so I'd say this should be fixed in its own commit if it is a mistake to begin with.)

critiquebrainz/data/model/user.py
tigranl: Yes, I missed this line.

critiquebrainz/data/model/user.py
```diff
@@ -57,31 +63,23 @@ def list(cls, limit=None, offset=None):
...
-            return False
+        return self.reviews_today_count() >= self.user_type.reviews_per_day
 
     @property
     def is_vote_limit_exceeded(self):
-        if self.votes_today_count() >= self.user_type.votes_per_day:
-            return True
-        else:
-            return False
+        return self.votes_today_count() >= self.user_type.votes_per_day
```
gentlecat: I can accept changes to 3 methods above, the rest doesn't have much point since we don't exactly follow PEP 8. For example, we don't use maximum 79 character length rule.

PR #78, opened 2017-01-18
Move review count and user count to models.

ferbncode:
Move total review count and user count functions from views to models.


gentlecat:
@brainzbot, add to whitelist.

PR #79, opened 2017-01-26
CB-241: Make "Load Older Revisions" button function

ferbncode:
With reviews more than 10 revisions, the "Load older revisions" button must load older revisions. For instance, in the revisions of [this](https://critiquebrainz.org/review/81f0aa06-8b9c-4490-b090-7e54614fe59e/revisions) review, load older revisions button is not working. The console error in browser showed that the load_more() function is not defined.

### DIFF
--- a/critiquebrainz/frontend/templates/review/revisions.html
+++ b/critiquebrainz/frontend/templates/review/revisions.html
@@ -32,7 +32,7 @@
         <tbody id="results">{% include 'review/revision_results.html' %}</tbody>
       </table>
       {% if count > limit %}
-        <button id="more-button" type="button" class="btn btn-primary" onclick="load_more();" >{{ _('Load older revisions') }}</button>
+        <button id="more-button" type="button" class="btn btn-primary" onclick="load_more()">{{ _('Load older revisions') }}</button>
         <span id="loading-message" class="text-muted" style="display:none;">{{ _('Loading older revisions...') }}</span>
       {% endif %}
     </form>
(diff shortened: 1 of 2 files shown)


critiquebrainz/frontend/templates/review/revisions.html
```diff
@@ -59,7 +59,7 @@
 
       {% if count > limit %}
         var current_page = 0;
-        function load_more() {
+        $("#more-button").click(function () {
           var more_button = $("#more-button");
```
gentlecat: Perhaps this can be moved out then to avoid duplication?

critiquebrainz/frontend/templates/review/revisions.html
```diff
@@ -32,7 +32,7 @@
         <tbody id="results">{% include 'review/revision_results.html' %}</tbody>
       </table>
       {% if count > limit %}
-        <button id="more-button" type="button" class="btn btn-primary" onclick="load_more();" >{{ _('Load older revisions') }}</button>
+        <button id="more-button" type="button" class="btn btn-primary" >{{ _('Load older revisions') }}</button>
```
gentlecat: Don't leave spaces before `>`.

gentlecat:
Let's not amend commits while this is being reviewed. It's difficult to follow the changes that way.

critiquebrainz/frontend/templates/review/revisions.html
```diff
@@ -42,6 +42,28 @@
...
+            .done(function (data) {
+              loading_message.hide();
+              $("#results").append(data.results);
+              if (data.more === true) more_button.show();
+            })
+            .fail(function () {
+              alert({{ _('Failed to load older revisions!') | tojson }});
+            });
+      }
+    {% endif %}
```
gentlecat: Can you explain why this function has to be outside the `$(document).ready(...)`?

critiquebrainz/frontend/templates/review/revisions.html
```diff
@@ -32,7 +32,7 @@
         <tbody id="results">{% include 'review/revision_results.html' %}</tbody>
       </table>
       {% if count > limit %}
-        <button id="more-button" type="button" class="btn btn-primary" onclick="load_more();" >{{ _('Load older revisions') }}</button>
+        <button id="more-button" type="button" class="btn btn-primary" onclick="load_more()">{{ _('Load older revisions') }}</button>
```
gentlecat: Why did you add `onclick` back instead of using `click()` method on `$("#more-button")`?

critiquebrainz/frontend/templates/review/revisions.html
ferbncode: I tried to move the function(with the ```click()``` method on``` $("$more-button"))``` outside of the block for scripts but it stopped working(Maybe because jQuery wasn't getting loaded). It was getting loaded at the end). So, I tried to debug what was the problem with the ```onclick()```. I found [this](http://stackoverflow.com/questions/3371632/why-cant-i-use-onclick-to-execute-a-function-inside-a-jquery-document-ready) (I am new to js and learning js 🙂 ). So, the scope of the js function onclick was searching for should be global. I should have tried the first way in the link.

PR #80, opened 2017-01-28
CB-230: Migrate revisions-related code off the ORM

ferbncode:
Migrate revision code off the SQLAlchemy ORM.

### DIFF
--- a/critiquebrainz/db/revision.py
+++ b/critiquebrainz/db/revision.py
@@ -3,11 +3,13 @@ from critiquebrainz.db import exceptions as db_exceptions
 import sqlalchemy
 
 
-def get(review_id):
+def get(review_id, limit=1, offset=0):
     """Get revisions on a review ordered by the timestamp
 
     Args:
         review_id (uuid): ID of review
+        limit (int): The number of revisions to return(default=1)
+        offset (int): The number of revisions to skip from top(default=0)
 
     Returns:
         List of dictionaries of revisions of the review
@@ -18,8 +20,6 @@ def get(review_id):
             "timestamp": (datetime),
             "text": (string)
         }
-
-        and the number(count) of reviews
     """
     with db.engine.connect() as connection:
         result = connection.execute(sqlalchemy.text("""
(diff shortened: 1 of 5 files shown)


critiquebrainz/db/revision.py
```diff
@@ -0,0 +1,86 @@
...
+
+        and the number(count) of reviews
+    """
+    with db.engine.connect() as connection:
+        if order_desc == True:
+            result = connection.execute(sqlalchemy.text("""
+                SELECT id, text, timestamp
+                  FROM revision
+                 WHERE review_id = :review_id
+              ORDER BY timestamp DESC;
```
gentlecat: And `;` at the end of a query is not needed here.

critiquebrainz/db/revision.py
```diff
@@ -0,0 +1,86 @@
...
+                         ON vote.revision_id = revision.id
+                      WHERE review_id = :review_id
+                   ORDER BY timestamp DESC;
+        """),{
+            "review_id": review_id,
+        })
+
+        rows = result.fetchall()
+        if not rows:
+            raise db_exceptions.NoDataFoundException("Cannot find votes for the review")
```
gentlecat: "Cannot find votes for a specified review" is a better wording, I think. Would be even better if it had a review ID in it. Like, "Cannot find votes for review (ID: <review_id>)"

critiquebrainz/frontend/views/review.py
```diff
@@ -125,13 +119,14 @@ def revisions(id):
...
-
-    revisions = Revision.query.filter_by(review=review)
-    count = revisions.count()
-    revisions = revisions.order_by(desc(Revision.timestamp)).limit(RESULTS_LIMIT)
+    try:
+        revisions, count = db_revision.get(id, order_desc=True)
+    except db_exceptions.NoDataFoundException:
+        raise NotFound(gettext("The revision(s) you are looking for does not exist."))
+    votes = db_revision.get_votes(id)
+    revisions = revisions[: RESULTS_LIMIT]
```
gentlecat: No need to have a space after `:`.

critiquebrainz/db/revision.py
```diff
@@ -0,0 +1,79 @@
...
+          ORDER BY timestamp DESC
+        """), {
+            "review_id": review_id,
+        })
+
+        rows = result.fetchall()
+        if not rows:
+            raise db_exceptions.NoDataFoundException("Cannot find specified review.")
+        rows = [dict(row) for row in rows]
+        count = len(rows)
```
gentlecat: This doesn't look correct. This should be a total number of rows.

gentlecat: Oh, sorry. I misunderstood this. There are places where total count should be calculated separately, but it's not the case here. This wasn't about the naming, `count` is a good name for that.

critiquebrainz/frontend/views/review.py
```diff
@@ -125,13 +119,14 @@ def revisions(id):
...
-
-    revisions = Revision.query.filter_by(review=review)
-    count = revisions.count()
-    revisions = revisions.order_by(desc(Revision.timestamp)).limit(RESULTS_LIMIT)
+    try:
+        revisions, count = db_revision.get(id)
+    except db_exceptions.NoDataFoundException:
+        raise NotFound(gettext("The revision(s) you are looking for does not exist."))
+    votes = db_revision.get_votes(id)
+    revisions = revisions[:RESULTS_LIMIT]
```
gentlecat: The problem here is that you are retrieving all revisions of a review just to throw them away. One way to improve this is to allow passing `limit` and `offset` arguments into the `get` function.

gentlecat:
Deployed. Thank you!

PR #82, opened 2017-02-12
CB-230: Migrate user code off the ORM

ferbncode:
Migrated users code off the ORM and created a seperate class for `User` in `db/user.py` using the SQLAlchemy core code for `user`. Made changes accordingly in the `frontend` for using the migrated code. However, changes in `webservice` is yet to be made. I've added the following functions in `users.py` and corresponding tests:
+ `get_by_id` for fetching user by id
+ `get_by_mbid` for fetching user by mbid
+ `get or create` and `create` for creating new user
+ `get_count` for counting number of users
+ `list` for listing users with `offset` and `limit`
+ `block` and `unblock` for blocking and unblocking users
+ `karma` for user karma
+ `has_voted`, `get_votes` for user has voted and user votes since a date
+ `reviews`, `get_reviews` for user reviews and user reviews since a date
+ `update` and `delete` for updating and deleting user info
+ `clients` and `tokens` for user registered clients and tokens

I've made changes in `Review.create`, `SpamReport.create` and `Vote.create` to use `user_id` rather that `data/User` object in order to use the migrated code.

### DIFF
--- a/critiquebrainz/db/users.py
+++ b/critiquebrainz/db/users.py
@@ -1,6 +1,9 @@
 from hashlib import md5
+import critiquebrainz.db.revision as db_revision
 import sqlalchemy
 from critiquebrainz import db
+from datetime import datetime
+import uuid
 
 
 def gravatar_url(source, default="identicon", rating="pg"):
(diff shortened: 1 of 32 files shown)


critiquebrainz/db/users.py
```diff
@@ -70,3 +73,543 @@ def get_many_by_mb_username(usernames):
...
+            "id": (uuid)
+            "display_name": (str)
+            "email": (str)
+            "created": (datetime)
+            "musicbrainz_id": (str)
+            "show_gravatar": (bool)
+            "is_blocked": (bool)
+        }
+    """
+    id = str(uuid.uuid4())
```
gentlecat: You can avoid making a variable and just move the `str(uuid.uuid4())` call into a dict in `sqlalchemy.text` below.

critiquebrainz/db/users.py
```diff
@@ -70,3 +73,543 @@ def get_many_by_mb_username(usernames):
...
+                "musicbrainz_id": musicbrainz_id,
+                "show_gravatar": show_gravatar,
+                "is_blocked": is_blocked
+            })
+        new_id = result.fetchone()[0]
+        user = get_by_id(new_id)
+    return user
+
+
+def get_by_mbid(musicbrainz_id):
```
gentlecat: There's one problem with current naming in the schema... It's not exactly correct to call it `musicbrainz_id`, it's actually a username, which is not the same thing as MusicBrainz ID or ID of a user in MusicBrainz.

We can start improving this here by actually calling it `musicbrainz_username` in our database interface. There's no need to rename the row in the schema for now.

critiquebrainz/db/users.py
```diff
@@ -70,3 +73,543 @@ def get_many_by_mb_username(usernames):
...
+            WHERE musicbrainz_id = :musicbrainz_id
+        """), {
+            "musicbrainz_id": musicbrainz_id
+        })
+        row = result.fetchone()
+    return dict(row) if row else None
+
+
+def get_or_create(display_name, musicbrainz_id, **kwargs):
+    """Get user from display_name and musicbrainz_id
```
gentlecat: This is not what this function seems to be doing. It tries to retrieve a user by their MusicBrainz username, and if the user is not found it creates a new one with a bunch of data that is provided.

Here's what I'll suggest to do here: move `display_name` into `kwargs` and rename `kwargs` into `new_user_data` to make it more descriptive.

gentlecat: It would actually be better to accept raw dict with user data instead of expanding it into arguments. So that function will be used like this:
```python
get_or_create(musicbrainz_username="example", new_user_data={
    "display_name": "Example user",
    "email": "user@example.org",
    ...
})
```

Then that dictionary can be passed into `create`.

critiquebrainz/db/users.py
```diff
@@ -70,3 +73,543 @@ def get_many_by_mb_username(usernames):
...
+            "is_blocked": (bool)
+        }
+    """
+    user = get_by_mbid(musicbrainz_id)
+    if not user:
+        user = create(display_name, musicbrainz_id=musicbrainz_id, **kwargs)
+    return user
+
+
+def get_count():
```
gentlecat: `total_count` might be a better name.

critiquebrainz/db/users.py
```diff
@@ -70,3 +73,543 @@ def get_many_by_mb_username(usernames):
...
+        result = connection.execute(sqlalchemy.text("""
+            SELECT count(*)
+              FROM "user"
+        """))
+        count = result.fetchone()[0]
+    return count
+
+
+def list(limit=None, offset=0):
+    """Returns the list of users of critiquebrainz
```
gentlecat: **C**ritique**B**rainz

critiquebrainz/db/users.py
```diff
@@ -70,3 +73,543 @@ def get_many_by_mb_username(usernames):
...
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT count(*)
+              FROM "user"
+        """))
+        count = result.fetchone()[0]
+    return count
+
+
+def list(limit=None, offset=0):
```
gentlecat: `list` is a keyword, so it's better to avoid using it like that.

critiquebrainz/db/users.py
```diff
@@ -70,3 +73,543 @@ def get_many_by_mb_username(usernames):
...
+            "created": (datetime)
+            "musicbrainz_id": (str)
+            "show_gravatar": (bool)
+            "is_blocked": (bool)
+        }
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT id, display_name, email, created,
+                   musicbrainz_id, show_gravatar, is_blocked
```
gentlecat: Each row name should be on a new line. https://github.com/metabrainz/guidelines/blob/master/Python.md#sql

critiquebrainz/db/users.py
```diff
@@ -70,3 +73,543 @@ def get_many_by_mb_username(usernames):
...
+        karma(int): the karma of the user
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT review.user_id, vote
+              FROM ((vote
+         LEFT JOIN revision
+                ON revision.id = revision_id)
+         LEFT JOIN review
+                ON review.id = review_id)
```
gentlecat: You might need to add another level of indentation by parentheses to make this easier to read.

critiquebrainz/db/users.py
```diff
@@ -70,3 +73,543 @@ def get_many_by_mb_username(usernames):
...
+            "date": date
+        })
+
+        rows = result.fetchall()
+        rows = [dict(row) for row in rows]
+
+    return rows
+
+
+def get_reviews(user_id, date='1-1-1970'):
```
gentlecat: What's this default `date`?

critiquebrainz/db/users.py
```diff
@@ -70,3 +73,543 @@ def get_many_by_mb_username(usernames):
...
+            "display_name": display_name,
+            "show_gravatar": show_gravatar,
+            "email": email
+        })
+
+
+def delete(user_id):
+    """Delete a user
+    Deletes user, votes, reviews, revisions and spam reports
+    by it
```
gentlecat: `This function deletes user and all of the information associated with them.`

gentlecat: You should also end all your sentences in docstrings with `.`.

critiquebrainz/db/users.py
```diff
@@ -70,3 +73,543 @@ def get_many_by_mb_username(usernames):
...
+        """), {
+            "user_id": user_id
+        })
+        connection.execute(sqlalchemy.text("""
+            DELETE
+              FROM "vote"
+             WHERE user_id = :user_id
+        """), {
+            "user_id": user_id
+        })
```
gentlecat: Associated information shouldn't be deleted like this. See `ON DELETE` rules in https://www.postgresql.org/docs/9.5/static/ddl-constraints.html.

critiquebrainz/db/users.py
ferbncode: In `db/users.py`, `get_reviews` and  `get_votes` were used  for `reviews_since` and `votes_since` in `User` class (to fetch reviews and votes from a certain date). To fetch all reviews or votes, I set`date` to start of Unix time. I changed the function name `get_reviews_since` and `get_vote_since` to increase understanding. Should it be done in some other preferred way?

gentlecat: I'd rather have this date optional. If it's specified then additional `WHERE` clause can be added.

Also, `from_date` is a more clear name for this than just `date`.

critiquebrainz/db/users.py
```diff
@@ -65,8 +68,570 @@ def get_many_by_mb_username(usernames):
...
+        rows = [dict(row) for row in rows]
+
+    return rows
+
+
+def update(user, display_name=None, email=None, show_gravatar=None):
+    """Update info of a user.
+
+    Args:
+        user: User object whose info is to be updated.
```
gentlecat: I think we shouldn't mix this stuff and just accept user's ID. You can add rows for updating dynamically depending on which arguments are specified. All of them can be merged into a dictionary of keyword arguments too.

See https://github.com/metabrainz/acousticbrainz-server/blob/4543427f492f18d89341ba190ba07eade7890753/db/dataset.py#L60-L95 for example.

critiquebrainz/db/users.py
```diff
@@ -65,8 +68,570 @@ def get_many_by_mb_username(usernames):
...
+    """Get the oauth_tokens from a user_id.
+
+    Args:
+        user_id(uuid): ID of the user.
+
+    Returns:
+        List of dictionaries of client tokens of the user
+        {
+            "id": (int),
+            "client_id": (Unicode),
```
gentlecat: Why not just "string"? All of them are unicode in Python 3.

PR #83, opened 2017-02-17
CB-120: Show more reviews from the same author on review page

ferbncode:
Show more reviews (randomly sorted) from the same author on the review page
only if the author has written other reviews. If the author has written only one review, the review 
page stays exactly the same.

When some user is signed in:
signed in

When no one has logged in:
signed out

Single review written:
single review

### DIFF
--- a/critiquebrainz/data/model/review.py
+++ b/critiquebrainz/data/model/review.py
@@ -144,6 +144,7 @@ class Review(db.Model, DeleteMixin):
                 False if not.
             inc_hidden: True if reviews marked as hidden should be included,
                 False if not.
+            exclude: List of id of reviews to exclude.
 
         Returns:
             Pair of values: list of reviews that match applied filters and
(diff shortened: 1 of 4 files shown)


critiquebrainz/frontend/views/review.py
```diff
@@ -85,7 +85,12 @@ def entity(id, rev=None):
     else:  # otherwise set vote to None, its value will not be used
         vote = None
     review.text_html = markdown(revision['text'], safe_mode="escape")
-    return render_template('review/entity/%s.html' % review.entity_type, review=review, spotify_mappings=spotify_mappings, soundcloud_url=soundcloud_url, vote=vote)
+
+    user_all_reviews, review_count = Review.list(user_id=review.user_id, sort="random")
+    if review_count > 0:
+        user_all_reviews.remove(review)
```
gentlecat: This seems a bit inefficient having to retrieve all the reviews just to review one. How about having an `exclude` argument for `Review.list` function that accepts a list of review IDs to exclude from results?

critiquebrainz/data/model/review.py
```diff
@@ -180,6 +181,11 @@ def list(cls, **kwargs):
         if user_id is not None:
             query = query.filter(Review.user_id == user_id)
 
+        exclude = kwargs.pop('exclude', None)
+        if exclude is not None:
+            for review_id in exclude:
+                query = query.filter(Review.id != review_id)
```
gentlecat: Isn't it better to use "NOT IN" with a list instead? https://www.postgresql.org/docs/9.6/static/functions-comparisons.html

PR #85, opened 2017-03-05
Link "Read revision" buttons in reports page.

ferbncode:
Add redirect using `revision id` to `review/revisions/rev` page and add links of `revision` buttons in `report_results` page.

### DIFF
--- a/critiquebrainz/db/revision.py
+++ b/critiquebrainz/db/revision.py
@@ -101,3 +101,32 @@ def get_votes(review_id):
             elif row.vote == True:
                 votes[revision]['positive'] += 1
     return votes
+
+def get_revision_number(review_id, revision_id):
+    """Get revision number of the review from the revision_id.
+
+    Args:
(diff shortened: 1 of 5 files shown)


critiquebrainz/db/revision.py
```diff
@@ -101,3 +101,32 @@ def get_votes(review_id):
...
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT row_number
+              FROM
+                   (SELECT row_number() over(order by timestamp),
+                           id
+                      FROM revision
+                     WHERE review_id = :review_id)
+                        AS indexedrevisions
```
gentlecat: Small nitpick. I think it will look a bit better like this:
```sql
FROM (
    SELECT row_number() over(order by timestamp),
           id
      FROM revision
     WHERE review_id = :review_id
) AS indexed_revisions
WHERE id = :revision_id
```
:)

PR #88, opened 2017-03-13
CB-251: Fix ISE on some place review pages.

ferbncode:
Some `places` which may have a `life_span` attribute, may not have a `begin` (year) attribute. For example, see review: https://critiquebrainz.org/review/51fae187-b852-453c-9262-678b825ec31f.

### DIFF
--- a/critiquebrainz/frontend/templates/review/entity/place.html
+++ b/critiquebrainz/frontend/templates/review/entity/place.html
@@ -21,9 +21,10 @@
     {% endif %}
 
     {{ _('%(place)s', place=event_name) }}
-
     {% if place['life-span'] %}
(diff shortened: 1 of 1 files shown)


critiquebrainz/frontend/templates/review/entity/place.html
```diff
@@ -21,9 +21,10 @@ <h2 id="title">
     {% endif %}
 
     {{ _('%(place)s', place=event_name) }}
-
     {% if place['life-span'] %}
-      <small>{{ place['life-span']['begin'][:4] }}</small>
+      {% if place['life-span']['begin'] %}
```
gentlecat: Can this `if` statement be combined with the one above?

PR #89, opened 2017-03-14
Fix a couple language issues in the MusicBrainz release group links

mwiencek:
* Fixes a grammatical error due to a missing article.
* Adds more context to the "See all on MusicBrainz" string, which seems too ambiguous for translators on its own (see all what?). I think it reads better with "releases," anyway.


gentlecat:
@brainzbot, add to whitelist.

PR #91, opened 2017-03-19
CB-230: Migrate review related code off the ORM

ferbncode:
Functions and tests for migrating `reviews` off the ORM. Changes have been made to the `frontend` component to use the migrated code. I have added the following functions in `db/review.py` and the corresponding tests (`db/review_test.py`):

+ `get_by_id` for getting review by id.
+ `hide` and `unhide` for hiding and unhiding reviews.
+ `get_count` for review count.
+ `update` for updating a reviews info.
+ `create` for creating a review.
+ `delete` for deleting a review.
+ `list` for list reviews and filter them on various filters.
+ `get_popular` for getting popular reviews.
+ `revision.create` for creating a revision.
+ `votes` for getting votes of a revision.

### DIFF
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -1,6 +1,7 @@
 import sqlalchemy
 from critiquebrainz import db
 from critiquebrainz.db import exceptions as db_exceptions
+from flask import abort
 from brainzutils import cache
 from werkzeug.exceptions import BadRequest
 from random import shuffle
@@ -10,15 +11,24 @@ from critiquebrainz.db.user import User
 import critiquebrainz.db.users as db_users
 from critiquebrainz.db.user import User
 import uuid
+from datetime import datetime, timedelta
 
 REVIEW_CACHE_NAMESPACE = "Review"
 DEFAULT_LICENSE_ID = "CC BY-SA 3.0"
 DEFAULT_LANG = "en"
 ENTITY_TYPES = [
-    'event',
-    'place',
-    'release_group',
+    "event",
+    "place",
+    "release_group",
 ]
+
+def get_or_404(review_id):
+    """Get a review using review ID or abort with error 404."""
+    try:
+        review = get_by_id(review_id)
+    except db_exceptions.NoDataFoundException:
+        abort(404)
+    return review
 
 
 def to_dict(review, confidential=False):
@@ -27,7 +37,7 @@ def to_dict(review, confidential=False):
     review["id"] = str(review["id"])
     review["entity_id"] = str(review["entity_id"])
     review["last_updated"] = review["last_revision"]["timestamp"]
-    review['last_revision']['review_id'] = str(review['last_revision']['review_id'])
+    review["last_revision"]["review_id"] = str(review["last_revision"]["review_id"])
     return review
 
 
@@ -41,7 +51,7 @@ def get_by_id(review_id):
         {
             "id": uuid,
             "entity_id": uuid,
-            "entity_type": 'release_group', 'event', 'place',
+            "entity_type": str("release group", "event", "place"),
             "user_id": uuid,
             "user": dict,
             "edits": int,
@@ -54,48 +64,45 @@ def get_by_id(review_id):
             "last_revision: dict,
             "votes": dict,
             "rating": int,
-            "text": str, "created": datetime,
+            "text": str,
+            "created": datetime,
             "license": dict,
         }
     """
     with db.engine.connect() as connection:
         result = connection.execute(sqlalchemy.text("""
             SELECT review.id AS id,
-                   entity_id,
-                   entity_type,
-                   user_id,
-                   edits,
-                   is_draft,
-                   is_hidden,
-                   license_id,
-                   language,
-                   source,
-                   source_url,
+                   review.entity_id,
+                   review.entity_type,
+                   review.user_id,
+                   review.edits,
+                   review.is_draft,
+                   review.is_hidden,
+                   review.license_id,
+                   review.language,
+                   review.source,
+                   review.source_url,
                    revision.id AS last_revision_id,
-                   timestamp,
-                   text,
-                   email,
+                   revision.timestamp,
+                   revision.text,
+                   "user".email,
                    "user".created as user_created,
-                   display_name,
-                   show_gravatar,
-                   musicbrainz_id,
-                   is_blocked,
+                   "user".display_name,
+                   "user".show_gravatar,
+                   "user".musicbrainz_id,
+                   "user".is_blocked,
                    created_time.created,
                    license.full_name,
                    license.info_url
               FROM review
-              JOIN revision
-                ON revision.review_id = review.id
-              JOIN "user"
-                ON "user".id = review.user_id
-              JOIN license
-                ON license.id = license_id
+              JOIN revision ON revision.review_id = review.id
+              JOIN "user" ON "user".id = review.user_id
+              JOIN license ON license.id = license_id
               JOIN (
                     SELECT review.id,
                            timestamp AS created
                       FROM review
-                      JOIN revision
-                        ON review.id = revision.review_id
+                      JOIN revision ON review.id = revision.review_id
                      WHERE review.id = :review_id
                   ORDER BY revision.timestamp ASC
                      LIMIT 1
@@ -108,7 +115,7 @@ def get_by_id(review_id):
 
         review = result.fetchone()
         if not review:
-            raise db_exceptions.NoDataFoundException("No review found with review ID: {}".format(review_id))
+            raise db_exceptions.NoDataFoundException("Can't find review with ID: {id}".format(id=review_id))
 
         review = dict(review)
         review["last_revision"] = {
@@ -147,7 +154,7 @@ def hide(review_id):
     with db.engine.connect() as connection:
         connection.execute(sqlalchemy.text("""
             UPDATE review
-               SET is_hidden='true'
+               SET is_hidden='t'
              WHERE id = :review_id
         """), {
             "review_id": review_id,
@@ -163,25 +170,20 @@ def unhide(review_id):
     with db.engine.connect() as connection:
         connection.execute(sqlalchemy.text("""
             UPDATE review
-               SET is_hidden='fa'
+               SET is_hidden='f'
              WHERE id = :review_id
         """), {
             "review_id": review_id,
         })
 
 
-def get_count(**kwargs):
+def get_count(*, is_draft=False, is_hidden=False):
     """Get number of reviews in CritiqueBrainz.
 
     Args:
         is_draft (bool): True if drafted reviews are to be counted.
         is_hidden (bool): True if hidden reviews are to be counted.
     """
-    is_drafted = kwargs.pop('is_draft', False)
-    is_hidden = kwargs.pop('is_hidden', False)
-    if(kwargs):
-        raise TypeError('Unexpected **kwargs: %r' % kwargs)
-
     with db.engine.connect() as connection:
         result = connection.execute(sqlalchemy.text("""
             SELECT count(*)
@@ -189,13 +191,14 @@ def get_count(**kwargs):
              WHERE is_draft = :is_drafted
                AND is_hidden = :is_hidden
         """), {
-            "is_drafted": is_drafted,
+            "is_drafted": is_draft,
             "is_hidden": is_hidden,
         })
         return result.fetchone()[0]
 
 
-def update(review_id, drafted, **kwargs):
+def update(*, review_id, drafted, text, license_id=None,
+           language=None, is_draft=None):
     """Update contents of this review.
 
     Args:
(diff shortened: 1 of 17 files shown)


critiquebrainz/db/review.py
```diff
@@ -0,0 +1,608 @@
...
+        connection.execute(sqlalchemy.text("""
+            COMMIT;
+        """))
+        review_id = result.fetchone()[0]
+        review_revision = db_revision.create(review_id, text)
+        cache.invalidate_namespace(REVIEW_CACHE_NAMESPACE)
+    return get_by_id(review_id)
+
+
+def list_reviews(**kwargs):
```
gentlecat: So the reason that all arguments were put into `**kwargs` is to force users of this function to explicitly specify them by keywords only. This was implemented before CB switched to Python 3. Python 3 has a better way of handling this: https://www.python.org/dev/peps/pep-3102/.

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,608 @@
...
+               email,
+               "user".created as user_created,
+               musicbrainz_id,
+               MIN(revision.timestamp) as created,
+               SUM(
+                CASE
+                 WHEN vote='t' THEN 1
+                 ELSE 0
+                END
+               ) AS votes_positive_count,
```
gentlecat: A couple of problems here. The goal of splitting parts of a query into multiple rows is to make it more readable. It doesn't mean that everything needs to be like that. This query doesn't even fit on my screen.

Statement like this is still readable:
```sql
SUM(
    CASE WHEN vote='t' THEN 1 ELSE 0 END
) AS votes_positive_count,
```

gentlecat: And make sure that you use 4 spaces for indentation, not 1.

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,608 @@
...
+               license.info_url
+          FROM review
+          JOIN revision
+            ON review.id = revision.review_id
+     LEFT JOIN vote
+            ON vote.revision_id = revision.id
+          JOIN "user"
+            ON review.user_id = "user".id
+          JOIN license
+            ON license.id = license_id
```
gentlecat: Same with these JOINs. This is fine:
```sql
JOIN license ON license.id = license_id
```

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,608 @@
...
+                    FROM review
+                    JOIN revision
+                      ON review.id = review_id
+                GROUP BY review.id
+                     ) AS latest
+                  ON latest.review_uuid = revision.review_id
+                 AND latest.latest_timestamp = revision.timestamp
+               ) AS latest_revision
+            ON review.id = latest_revision.review_id
+            {}
```
gentlecat: Placeholder names should be explicitly specified: `{where_clause}` instead of `{}`, etc.

critiquebrainz/db/vote.py
```diff
@@ -56,6 +56,7 @@ def submit(user_id, revision_id, vote):
             "vote": vote,
             "rated_at": datetime.utcnow(),
         })
+        return get(user_id, revision_id)
```
gentlecat: Is data that returned from `submit` used somewhere?

gentlecat: Also don't forget to document the return value.

critiquebrainz/frontend/views/index.py
```diff
@@ -13,17 +13,17 @@
...
         preview = markdown(review['text'], safe_mode="escape")
         review['preview'] = ''.join(BeautifulSoup(preview, "html.parser").findAll(text=True))
 
     # Recent reviews
-    recent_reviews, _ = Review.list(sort='created', limit=9)
+    recent_reviews, _ = db_review.list_reviews(sort='created', limit=9)
 
     # Statistics
-    review_count = format_number(Review.get_count(is_draft = False))
+    review_count = format_number(db_review.get_count(is_draft = False))
```
gentlecat: There should've been no spaces around `=`.

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,625 @@
...
+            connection.execute(query, updated_info)
+    # Create new revision
+    new_revision = db_revision.create(review_id, text)
+    review = get_by_id(review_id)
+    cache.invalidate_namespace(REVIEW_CACHE_NAMESPACE)
+    return review
+
+
+def create(*, release_group=None, entity_id=None,
+           entity_type=None, user_id, is_draft, text,
```
gentlecat: I think this is the perfect time to refactor uses of this function a bit.

1. `entity_type` and `entity_id` arguments shouldn't be optional.
2. `release_group` argument should be removed and its uses replaced by `entity_id` with `entity_type` set to `release_group`.

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,625 @@
...
+           language=DEFAULT_LANG, license_id=DEFAULT_LICENSE_ID,
+           source=None, source_url=None):
+    """Create a new review on an entity.
+    Args:
+        release_group, entity_id (uuid): ID of the release group to be reviewed.
```
gentlecat: Not only release group anymore, since we can review other entity types too.

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,625 @@
...
+        entity_id (str): Type of the entity reviewed.
```
gentlecat: Should this be `entity_type` instead?

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,625 @@
...
+         RETURNING id;
+       """), {
+           "id": str(uuid.uuid4()),
+           "entity_id": entity_id,
+           "entity_type": entity_type,
+           "user_id": user_id,
+           "edits": 0,
+           "is_draft": is_draft,
+           "is_hidden": False,
+           "language": language,
```
gentlecat: It might be a good idea to check that provided language is one of the supported ones.

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,625 @@
...
+
+        review_id = result.fetchone()[0]
+        review_revision = db_revision.create(review_id, text)
+        cache.invalidate_namespace(REVIEW_CACHE_NAMESPACE)
+    return get_by_id(review_id)
+
+
+def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None,
+                 entity_type=None, license_id=None, user_id=None,
+                 language=None, exclude=None, sort=None, limit=None,
```
gentlecat: It would be a good idea to provide a reasonable default `limit` to avoid issues in case it's not provided. 20 sounds like a good number.

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,625 @@
...
+        user_id (uuid): UUID of the author.
+        sort (str): Order of the returned reviews. Can be either "rating" (order by rating), or "created"
+                    (order by creation time), or "random" (order randomly).
+        limit (int): Maximum number of reviews returned by this method.
+        offset (int): Offset that can be used in conjuction with the limit.
+        language (str): Language (code) of the returned reviews.
+        license_id (str): License of the returned reviews.
+        inc_drafts (bool): True if reviews marked as drafts should be included. False if not.
+        inc_hidden (bool): True if reviews marked as hidden should be included. False if not.
+        exclude (list): List of id of reviews to exclude.
```
gentlecat: `List of reviews (their IDs) to exclude from results.`

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,625 @@
...
+
+    Returns:
+        Dictionary of reviews.
+        Total number of reviews.
```
gentlecat: `Total number of reviews that match specified filters.`

critiquebrainz/frontend/views/review.py
```diff
@@ -226,45 +225,50 @@ def preview():
...
         else:
             license_choice = None
-        review.update(text=form.text.data, is_draft=(form.state.data == 'draft'),
-                      license_id=license_choice, language=form.language.data)
+        review = db_review.update(
+            review_id=review["id"],
+            drafted=review["is_draft"],
+            text=form.text.data,
+            is_draft=(form.state.data == 'draft'),
+            license_id=license_choice, language=form.language.data,
```
gentlecat: Can you put `language` argument on a new line?

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,625 @@
...
+    """Get a review using review ID or abort with error 404."""
+    try:
+        review = get_by_id(review_id)
+    except db_exceptions.NoDataFoundException:
+        abort(404)
+    return review
+
+
+def to_dict(review, confidential=False):
+    review["user"] = User(db_users.get_by_id(review.pop("user_id")))
```
ferbncode: In `review_credit` macro present in `macros.html`, `review.user.avatar` is used to show user avatar which is a property in the `User` class. The macro is used in `popular reviews` section and  `browse reviews` pages.

gentlecat: Got it. Can improve that part later, if necessary.

critiquebrainz/db/review.py
```diff
@@ -306,12 +272,8 @@ def create(*, release_group=None, entity_id=None,
...
         }
     """
-    if release_group:
-       entity_id = release_group
-       entity_type = "release_group"
-    else:
-        if not entity_type or not entity_id:
-          raise TypeError("Entity Type or ID not defined")
+    if language not in supported_languages:
+        raise TypeError("Language: {} is not supported".format(language))
```
gentlecat: This doesn't look like a `TypeError`. Take a look its description: https://docs.python.org/3.6/library/exceptions.html#TypeError. Someone might pass an invalid language as a string.

[`ValueError`](https://docs.python.org/3.6/library/exceptions.html#ValueError) would be the right one to use here.

critiquebrainz/frontend/views/review.py
```diff
@@ -365,7 +366,7 @@ def hide(id):
 
     form = AdminActionForm()
     if form.validate_on_submit():
-        db_review.hide(review["id"])
+        db_review.set_hidden_state(review["id"], is_hidden="True")
```
gentlecat: The `is_hidden` argument should be a `bool`. Pretty sure you can pass it straight to psycopg too.

critiquebrainz/frontend/views/review.py
```diff
@@ -382,6 +383,6 @@ def hide(id):
 @admin_view
 def unhide(id):
     review = db_review.get_or_404(id)
-    db_review.unhide(review["id"])
+    db_review.set_hidden_state(review["id"], is_hidden="False")
```
gentlecat: Same issue with `is_hidden` here.

gentlecat: If you have to accept boolean arguments as strings in an interface, it's a good sign that something is not right. :)

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,587 @@
...
+    "event",
+    "place",
+    "release_group",
+]
+
+supported_languages = []
+for lang in list(pycountry.languages):
+    if 'iso639_1_code' in dir(lang):
+        supported_languages.append(lang.iso639_1_code)
+
```
gentlecat: Minor thing: there should be two spaces between functions and other code.

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,587 @@
...
+        license_id(optional, str): Updated license id of a review.
+        is_draft(optional, bool): Whether to publish review or keep it drafted.
+        text(optional, str): Updated text of a review.
+    """
+    updates = []
+    updated_info = {}
+
+    if license_id is not None:
+        if not drafted: # If trying to convert published review into draft.
+            raise BadRequest(lazy_gettext("Changing license of a published review is not allowed."))
```
gentlecat: One other important thing that I missed. We should try to avoid making Flask views and database access code closely tied. That means not using exceptions and functions from Flask.

Modules in the `critiquebrainz.db` package should raise exceptions from `critiquebrianz.db.exceptions`. I understand that this is not the old implementation, but that's the perfect time to make this improvement. This would mean we could easily use DB access code without Flask.

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,587 @@
...
+    "place",
+    "release_group",
+]
+
+supported_languages = []
+for lang in list(pycountry.languages):
+    if 'iso639_1_code' in dir(lang):
+        supported_languages.append(lang.iso639_1_code)
+
+def get_or_404(review_id):
```
gentlecat: We probably shouldn't have this function here. As I described in another comment, this makes `db` package closely tied to Flask itself. This function can be moved into packages with views if it's used more than once.

gentlecat:
Looks like there's a small conflict with another change. Would you mind making an update please?

gentlecat:
Great. 👍

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,577 @@
...
+        }
+    """
+    if language not in supported_languages:
+        raise ValueError("Language: {} is not supported".format(language))
+
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            INSERT INTO review
+            VALUES (:id, :entity_id, :entity_type, :user_id, :edits, :is_draft,
+            :is_hidden, :license_id, :language, :source, :source_url)
```
gentlecat: This way of insertion is unsafe and caused CB-258. Order of columns needs to be explicitly specified. Here you are making an assumption that `entity_type` is third, which it might be when a new instance of CB database is created, but it isn't in the production database.

critiquebrainz/db/review.py
```diff
@@ -0,0 +1,577 @@
...
+
+    Args:
+        limit: Maximum number of reviews to return.
+
+    Returns:
+        Randomized list of popular reviews which are converted into dictionaries using to_dict method.
+    """
+    cache_key = cache.gen_key("popular_reviews", limit)
+    reviews = cache.get(cache_key, REVIEW_CACHE_NAMESPACE)
+    reviews = None
```
gentlecat: What's the point of retrieving reviews from cache and then getting rid of them in the next statement?

gentlecat:
Next time make sure that your PR is small and easy to review. Otherwise, as evidenced here, this causes many issues.

PR #93, opened 2017-03-29
CB-230: Migrate Moderation Log code off the ORM.

ferbncode:
Added functions `moderation_log.create` and `moderation_log.list_logs` and corresponding tests for managing logs.

### DIFF
--- a/critiquebrainz/db/moderation_log.py
+++ b/critiquebrainz/db/moderation_log.py
@@ -12,7 +12,7 @@ ACTION_BLOCK_USER = "block_user"
 
 def create(*, admin_id, review_id=None, user_id=None,
            action, reason):
-    """Log blocking a user or hiding a review with IDs.
+    """Make a record in the moderation log.
 
     Args:
         admin_id (uuid): ID of the admin.
@@ -22,7 +22,7 @@ def create(*, admin_id, review_id=None, user_id=None,
         reason (str): Reason for blocking a user or hiding a review.
     """
     if not review_id and not user_id:
-        raise TypeError("No review ID or user ID specified.")
+        raise ValueError("No review ID or user ID specified.")
     if action != ACTION_BLOCK_USER and action != ACTION_HIDE_REVIEW:
         raise TypeError("Please specify a valid action.")
     with db.engine.connect() as connection:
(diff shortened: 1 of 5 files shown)


critiquebrainz/db/moderation_log.py
```diff
@@ -0,0 +1,139 @@
...
+
+    Args:
+        admin_id (uuid): ID of the admin.
+        user_id (uuid): ID of the user blocked.
+        review_id (uuid): ID of the review hidden.
+        action (str): str("hide_review", "block_user").
+        reason (str): Reason for blocking a user or hiding a review.
+    """
+    if not review_id and not user_id:
+        raise TypeError("No review ID or user ID specified.")
```
gentlecat: [`ValueError`](https://docs.python.org/3.6/library/exceptions.html#ValueError) would be more appropriate here and below.

critiquebrainz/db/moderation_log.py
```diff
@@ -0,0 +1,139 @@
...
+import sqlalchemy
+from datetime import datetime
+
+
+ACTION_HIDE_REVIEW = "hide_review"
+ACTION_BLOCK_USER = "block_user"
+
+def create(*, admin_id, review_id=None, user_id=None,
+           action, reason):
+    """Log blocking a user or hiding a review with IDs.
```
gentlecat: > blocking a user or hiding a review

This doesn't have to be so specific. You could just say something like "Make a record in the moderation log."

critiquebrainz/db/moderation_log.py
```diff
@@ -0,0 +1,139 @@
...
+         LIMIT :limit
+        OFFSET :offset
+    """.format(where_clause=where_clause))
+
+    with db.engine.connect() as connection:
+        result = connection.execute(query, filter_data)
+        logs = result.fetchall()
+        logs = [dict(log) for log in logs]
+        if logs:
+            for log in logs:
```
gentlecat: Instead of having two loops you could just do something like
```python
logs = []
for log in result.fetchall():
    ...
```

critiquebrainz/db/moderation_log.py
```diff
@@ -0,0 +1,139 @@
...
+        admin_id (uuid): ID of the admin.
+        user_id (uuid): ID of the user blocked.
+        review_id (uuid): ID of the review hidden.
+        action (str): str("hide_review", "block_user").
+        reason (str): Reason for blocking a user or hiding a review.
+    """
+    if not review_id and not user_id:
+        raise ValueError("No review ID or user ID specified.")
+    if action != ACTION_BLOCK_USER and action != ACTION_HIDE_REVIEW:
+        raise TypeError("Please specify a valid action.")
```
gentlecat: Same problem with exception. Make sure to use the right class everywhere.

PR #95, opened 2017-03-31
CB-230: Migrate OAuth Client code off the ORM.

ferbncode:
Added functions and tests to migrate code for `OAuthClient` model. Also, made changes to `frontend` for using migrated code.

### DIFF
--- a/critiquebrainz/db/oauth_client.py
+++ b/critiquebrainz/db/oauth_client.py
@@ -4,8 +4,12 @@ from critiquebrainz.utils import generate_string
 import sqlalchemy
 
 
+CLIENT_ID_LENGTH = 20
+CLIENT_SECRET_LENGTH = 40
+
+
 def create(*, user_id, name, desc, website, redirect_uri):
-    """Create new OAuth client and generates secret key for it.
+    """Creates new OAuth client and generates secret key for it.
 
     Args:
         user_id (uuid): ID of the user who manages the client.
@@ -25,24 +29,19 @@ def create(*, user_id, name, desc, website, redirect_uri):
             "redirect_uri": redirect_uri,
         }
     """
-    client_id = generate_string(20)
-    client_secret = generate_string(40)
-    client = {
-        "client_id": client_id,
-        "client_secret": client_secret,
+    with db.engine.connect() as connection:
+        connection.execute(sqlalchemy.text("""
+            INSERT INTO oauth_client VALUES(:client_id, :client_secret, :redirect_uri,
+                :user_id, :name, :desc, :website)
+        """), {
+        "client_id": generate_string(CLIENT_ID_LENGTH),
+        "client_secret": generate_string(CLIENT_SECRET_LENGTH),
         "redirect_uri": redirect_uri,
         "user_id": user_id,
         "name": name,
         "website": website,
         "desc": desc,
-    }
-    with db.engine.connect() as connection:
-        connection.execute(sqlalchemy.text("""
-            INSERT INTO oauth_client VALUES(:client_id, :client_secret, :redirect_uri,
-                :user_id, :name, :desc, :website)
-        """), client)
-
-    return client
+    })
 
 
 def update(*, client_id, name=None, desc=None, website=None, redirect_uri=None):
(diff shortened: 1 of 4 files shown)


critiquebrainz/db/oauth_client.py
```diff
@@ -0,0 +1,137 @@
+from critiquebrainz import db
+from critiquebrainz.db import exceptions as db_exceptions
+from critiquebrainz.utils import generate_string
+import sqlalchemy
+
+
+def create(*, user_id, name, desc, website, redirect_uri):
+    """Create new OAuth client and generates secret key for it.
```
gentlecat: Should be either "Create**s** new OAuth client and generates a secret key for it." or "Create new OAuth client and generate a secret key for it."

critiquebrainz/db/oauth_client.py
```diff
@@ -0,0 +1,137 @@
...
+            "client_secret": str,
+            "user_id": uuid,
+            "name": str,
+            "desc": desc,
+            "website": website,
+            "redirect_uri": redirect_uri,
+        }
+    """
+    client_id = generate_string(20)
+    client_secret = generate_string(40)
```
gentlecat: These variables seem unnecessary. Calls to `generate_string` can be moved into the dictionary below.

I'd also move lengths into "constants" in this module: `CLIENT_ID_LENGTH` and `CLIENT_SECRET_LENGTH`.

critiquebrainz/db/oauth_client.py
```diff
@@ -0,0 +1,137 @@
...
+        "website": website,
+        "desc": desc,
+    }
+    with db.engine.connect() as connection:
+        connection.execute(sqlalchemy.text("""
+            INSERT INTO oauth_client VALUES(:client_id, :client_secret, :redirect_uri,
+                :user_id, :name, :desc, :website)
+        """), client)
+
+    return client
```
gentlecat: Is the return value being used anywhere outside of tests? If not, you can remove it and move dictionary directly into the `connection.execute` call.

critiquebrainz/db/oauth_client.py
```diff
@@ -0,0 +1,137 @@
...
+        website(str, optional): Client web site.
+        redirect_uri(str, optional): URI where responses will be sent.
+    """
+    updates = []
+    update_data = {}
+    if name is not None:
+        updates.append("name = :name")
+        update_data["name"] = name
+    if desc is not None:
+        updates.append(""" "desc" = :desc""")
```
gentlecat: `"""` should be used for multi-line strings and docstrings. Here you can use `'` instead.

critiquebrainz/db/oauth_client.py
```diff
@@ -0,0 +1,137 @@
...
+                   "desc",
+                   website
+              FROM oauth_client
+             WHERE client_id = :client_id
+        """), {
+            "client_id": client_id,
+        })
+        row = result.fetchone()
+        if not row:
+            raise db_exceptions.NoDataFoundException("No client with ID: {} found".format(client_id))
```
gentlecat: Minor thing, but I'd expect text after `:` to be just the ID. `Can't find OAuth client with ID: %(id)s` would be a better message, I think.

critiquebrainz/frontend/views/profile_apps.py
```diff
@@ -37,13 +38,20 @@ def create():
...
+        abort(404)
+    if application["user_id"] != current_user.id:
         raise NotFound()
     form = ApplicationForm()
     if form.validate_on_submit():
-        application.update(name=form.name.data, desc=form.desc.data,
-                           website=form.website.data, redirect_uri=form.redirect_uri.data)
+        db_oauth_client.update(
+            client_id=application["client_id"],
+            name=form.name.data, desc=form.desc.data,
```
gentlecat: Don't stack arguments like that. It's easy to miss it if someone is expecting each to be on a new line.

critiquebrainz/frontend/views/profile_apps.py
```diff
@@ -57,10 +65,13 @@ def edit(client_id):
 @profile_apps_bp.route('/<client_id>/delete')
 @login_required
 def delete(client_id):
-    client = OAuthClient.query.get_or_404(client_id)
-    if client.user != current_user:
+    try:
+        application = db_oauth_client.get(client_id)
+    except db_exceptions.NoDataFoundException:
+        abort(404)
```
gentlecat: `raise NotFound` and `abort(404)` give pretty much the same result. Let's just stick with `NotFound`. Can also pass a message from the "parent" exception.

PR #97, opened 2017-04-11
CB-198: Filter "browse/view all reviews" by entity type

ps2611:
It needs work on the UI but are the changes according to what the ticket was for?

### DIFF
--- a/critiquebrainz/frontend/templates/review/browse.html
+++ b/critiquebrainz/frontend/templates/review/browse.html
@@ -5,19 +5,22 @@
 
 {% block content %}
 <h2>{{ _('Reviews') }}</h2>
-<form method="GET" action="{{ url_for('review.browse') }}">
-  <div class="form-group" style="display: inline-block;">
-    <select class="form-control input-sm" name="entity_type" onchange="this.form.submit()"> 
-      <option value="None">{{ _('All') }}</option>
-      <option value="release_group" {% if entity_type == 'release_group' %} selected {% endif %}>{{ _('Release group') }}</option>
-      <option value="event" {% if entity_type == 'event' %} selected {% endif %}>{{ _('Event') }}</option>
-      <option value="place" {% if entity_type == 'place' %} selected {% endif %}>{{ _('Place') }}</option>
-    </select>
-  </div>
-</form>
 <p><em class="text-muted">{{ _('Ordered by creation date') }}</em></p>
+<ul class="nav nav-pills"> 
+  <li role="presentation" {{ "class=active" if entity_type == None }}>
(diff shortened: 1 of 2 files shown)


gentlecat:
@brainzbot, ok to test.

gentlecat:
Please don't forget to assign that ticket to yourself and mark it as "In review". That way it's easier for us all to keep track of what other people are working on.

critiquebrainz/frontend/templates/review/browse.html
```diff
@@ -5,6 +5,16 @@
 
 {% block content %}
 <h2>{{ _('Reviews') }}</h2>
+<form method="GET" action="{{ url_for('review.browse') }}">
+  <div class="form-group" style="display: inline-block;">
+    <select class="form-control input-sm" name="entity_type" onchange="this.form.submit()">
```
gentlecat: I think it would look better if we used "[nav-pills](https://getbootstrap.com/components/#nav-pills)" here (similar to what we have at https://metabrainz.org/reports). A row that looks like `Entity type: <All> <Release group> <Event> <Place>`.

critiquebrainz/frontend/templates/review/browse.html
```diff
@@ -5,6 +5,16 @@
 
 {% block content %}
 <h2>{{ _('Reviews') }}</h2>
+<form method="GET" action="{{ url_for('review.browse') }}">
+  <div class="form-group" style="display: inline-block;">
+    <select class="form-control input-sm" name="entity_type" onchange="this.form.submit()"> 
+      <option value="None">{{ _('All') }}</option>
```
gentlecat: Can just have an `all` value here instead.

critiquebrainz/frontend/templates/review/browse.html
ps2611: I did not include 'Entity type:'  because it broke the line and looked congested on smaller screens.
Do we need to do that? I think, with all the entity types listed it's quite self-explanatory.

critiquebrainz/frontend/templates/review/browse.html
ps2611: Did this in the latest commit

critiquebrainz/frontend/templates/review/browse.html
ps2611: potrait_mode

gentlecat: I've been considering that it would be useful when we have more types of filters in the future. But if that's too much of a problem now, we can improve this later.

PR #98, opened 2017-04-12
CB-230: Migrate license related code off the ORM

ferbncode:
Added function `license.create` and `license.delete` for creating and deleting licenses. Made changes in their usages accordingly.

### DIFF
--- a/critiquebrainz/db/license.py
+++ b/critiquebrainz/db/license.py
@@ -13,7 +13,7 @@ def create(*, id, full_name, info_url=None):
     Returns:
         Dict containing the following attributes
         {
-            "id": (str),,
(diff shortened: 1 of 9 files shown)


critiquebrainz/db/license.py
```diff
@@ -0,0 +1,47 @@
...
+    """Create a new license.
+
+    Args:
+        id (str): ID of the license.
+        full_name (str): Full name of the license.
+        info_url (str): Info URL for the license.
+    Returns:
+        Dict containing the following attributes
+        {
+            "id": (str),,
```
gentlecat: Looks like a typo there.

PR #99, opened 2017-04-21
CB-230: Migrate OAuth Token related code off the ORM

ferbncode:
Added functions `create`, `list_tokens`, `delete` and `get_scopes` for managing `OAuth Tokens`. Also, added corresponding tests.

### DIFF
--- a/critiquebrainz/db/oauth_token.py
+++ b/critiquebrainz/db/oauth_token.py
@@ -130,7 +130,7 @@ def delete(*, client_id=None, refresh_token=None, user_id=None):
 
 
 def get_scopes(token_id):
-    """Returns the scopes of an application. CritiqueBrainz provides the scopes: review, vote and user.
+    """Returns the scopes of an application.
 
     Args:
         token_id (int): ID of the OAuth Token.
@@ -145,7 +145,7 @@ def get_scopes(token_id):
         """), {
             "token_id": token_id,
         })
-        scopes = result.fetchone()[0]
+        scopes = result.fetchone()
         if not scopes:
             return list()
-    return scopes.split()
+    return scopes[0].split()
--- a/critiquebrainz/db/oauth_token.py
+++ b/critiquebrainz/db/oauth_token.py
@@ -1,0 +1,151 @@
+from critiquebrainz import db
+from critiquebrainz.db import exceptions as db_exceptions
+import sqlalchemy
+
+
+def create(*, client_id, scopes, access_token, refresh_token, expires, user_id):
+    """Add OAuth Token to database.
+
+    Args:
+        client_id (str): ID of the OAuth Client.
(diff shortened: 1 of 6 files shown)


critiquebrainz/db/oauth_token.py
```diff
@@ -0,0 +1,151 @@
...
+    with db.engine.connect() as connection:
+        connection.execute(sqlalchemy.text("""
+            DELETE
+              FROM oauth_token
+             WHERE {filterstr}
+        """.format(filterstr=filterstr)), filter_data)
+
+
+def get_scopes(token_id):
+    """Returns the scopes of an application. CritiqueBrainz provides the scopes: review, vote and user.
```
gentlecat: There's probably no need to specify here which scopes CB provides.

critiquebrainz/db/oauth_token_test.py
```diff
@@ -0,0 +1,71 @@
...
+    def test_get_scopes(self):
+        oauth_token = db_oauth_token.create(
+            client_id=self.oauth_client["client_id"],
+            access_token="Test Access Token",
+            refresh_token="Test Refresh Token",
+            expires=datetime.now() + timedelta(seconds=200),
+            user_id=self.user.id,
+            scopes="Test Scopes",
+        )
+        self.assertIn("Test", db_oauth_token.get_scopes(oauth_token["id"]))
```
gentlecat: Here you should also test what happens when you try to retrieve scopes from a token that doesn't exist.

critiquebrainz/ws/oauth/provider.py
```diff
@@ -106,15 +112,14 @@ def persist_grant(client_id, code, scopes, expires, redirect_uri, user_id):
...
-        db.session.commit()
+        token = db_oauth_token.create(
+            client_id=client_id,
+            scopes=scope,
+            access_token=access_token,
+            refresh_token=refresh_token,
+            expires=expires,
+            user_id=user_id,
+        )
         return token
```
gentlecat: Here you can just do `return db_oauth_token.create(...)` without creating a variable.

critiquebrainz/db/oauth_token_test.py
```diff
@@ -69,3 +69,5 @@ def test_get_scopes(self):
             scopes="Test Scopes",
         )
         self.assertIn("Test", db_oauth_token.get_scopes(oauth_token["id"]))
+        db_oauth_token.delete(client_id=self.oauth_client["client_id"], refresh_token="Test Refresh Token")
+        self.assertEqual([], db_oauth_token.get_scopes(oauth_token["id"]))
```
gentlecat: I think it would be better if it thrown exception of some kind instead. Because otherwise there's no way to know if the token exists and it has no scope of if there's no token at all.

PR #102, opened 2017-05-16
CB-230: Migrate ws, frontend and db completely off the ORM

ferbncode:
Made changes to `frontend`, `db` and `ws` code to use migrated code. Also added `drop_tables.sql` and `drop_types.sql` to help `setUp()` and `tearDown()` in tests. Only functions remaining in `dump_manager.py` use ORM related code.


critiquebrainz/data/testing.py
```diff
@@ -14,9 +14,13 @@ def create_app(self):
...
 
     def setUp(self):
-        db.create_all()
+        self.reset_db()
         # TODO(roman): Add stuff form fixtures.
 
     def tearDown(self):
-        db.session.remove()
-        db.drop_all()
+        pass
```
gentlecat: Seems like `tearDown` can be just removed there if it's not needed.

critiquebrainz/frontend/testing.py
```diff
@@ -14,12 +14,16 @@ def create_app(self):
...
 
     def setUp(self):
-        db.create_all()
+        self.reset_db()
         # TODO(roman): Add stuff form fixtures.
 
     def tearDown(self):
-        db.session.remove()
-        db.drop_all()
+        pass
```
gentlecat: Same here. If `tearDown` is not needed it can just be removed.

critiquebrainz/ws/testing.py
```diff
@@ -18,12 +18,16 @@ def create_app(self):
...
 
     def setUp(self):
-        db.create_all()
+        self.reset_db()
         # TODO(roman): Add stuff form fixtures.
 
     def tearDown(self):
-        db.session.remove()
-        db.drop_all()
+        pass
```
gentlecat: And here.

manage.py
```diff
@@ -102,7 +102,7 @@ def init_db(skip_create_db=False):
         create_extension(db_uri)
 
     click.echo("Creating tables... ", nl=False)
-    data_utils.create_tables()
+    data_utils.create_all()
```
gentlecat: Have you tested this on a new instance? Does it work fine?

ferbncode: Yes, I have tested this on a new instance and it works fine 🙂

PR #103, opened 2017-05-21
CB-230: Migrate dump_manager.py off the ORM

ferbncode:
Migrated all functions i.e `public`, `import` and `json` of the `dump_manager.py` off the ORM.

### DIFF
--- a/critiquebrainz/data/dump_manager.py
+++ b/critiquebrainz/data/dump_manager.py
@@ -1,12 +1,14 @@
 from flask import current_app, jsonify
 from flask.json import JSONEncoder
 from critiquebrainz.data.utils import create_path, remove_old_archives, get_columns, slugify, explode_db_uri
+from critiquebrainz.db import license as db_license, review as db_review
 from critiquebrainz import frontend
 from critiquebrainz import db
 from time import gmtime, strftime
 from datetime import datetime
 from functools import wraps
 import subprocess
+import sqlalchemy
 import tempfile
 import tarfile
 import shutil
@@ -18,12 +20,63 @@ import os
 
 cli = click.Group()
 
+
+_TABLES = {
+    "review": (
+        "id",
+        "entity_id",
+        "entity_type",
+        "user_id",
+        "edits",
+        "is_draft",
+        "is_hidden",
+        "license_id",
+        "language",
+        "source",
+        "source_url",
+    ),
+    "revision": (
+        "id",
+        "review_id",
+        "timestamp",
+        "text",
+    ),
+    "license": (
+        "id",
+        "full_name",
+        "info_url",
+    ),
+    "user": (
+        "id",
+        "display_name",
+        "email",
+        "created",
+        "musicbrainz_id",
+        "show_gravatar",
+        "is_blocked",
+        "spam_reports",
+        "clients",
+        "grants",
+        "tokens"
+    ),
+}
+
+
 def with_request_context(f):
     @wraps(f)
     def decorated(*args, **kwargs):
         with frontend.create_app().test_request_context():
             return f(*args, **kwargs)
     return decorated
+
+
+def has_data(table_name):
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT COUNT(*)
+              FROM "{table_name}"
+        """.format(table_name=table_name)))
+        return result.fetchone()[0] > 0
 
 
 @cli.command()
@@ -92,7 +145,7 @@ def json(location, rotate=False):
     current_app.json_encoder = DumpJSONEncoder
 
     print("Creating new archives...")
-    for license in db.license.list_licenses():
+    for license in db_license.list_licenses():
         safe_name = slugify(license["id"])
         with tarfile.open(os.path.join(location, "critiquebrainz-%s-%s-json.tar.bz2" %
                 (datetime.today().strftime('%Y%m%d'), safe_name)), "w:bz2") as tar:
@@ -100,18 +153,18 @@ def json(location, rotate=False):
             license_dir = os.path.join(temp_dir, safe_name)
             create_path(license_dir)
 
-            # Finding release groups that have reviews with current license
-            entities = db.review.distinct_entities()
+            # Finding entities that have reviews with current license
+            entities = db_review.distinct_entities()
             for entity in entities:
                 entity = str(entity)
                 # Creating directory structure and dumping reviews
                 dir_part = os.path.join(entity[0:1], entity[0:2])
-                reviews = db.review.list_reviews(entity_id=entity, license_id=license["id"], limit=None)[0]
+                reviews = db_review.list_reviews(entity_id=entity, license_id=license["id"], limit=None)[0]
                 if len(reviews) > 0:
                     rg_dir = '%s/%s' % (license_dir, dir_part)
                     create_path(rg_dir)
                     f = open('%s/%s.json' % (rg_dir, entity), 'w+')
-                    f.write(jsonify(reviews=[db.review.to_dict(r) for r in reviews]).data.decode("utf-8"))
+                    f.write(jsonify(reviews=[db_review.to_dict(r) for r in reviews]).data.decode("utf-8"))
                     f.close()
 
             tar.add(license_dir, arcname='reviews')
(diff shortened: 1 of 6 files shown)


ferbncode:
The dumps created from the `public` function are a bit different from the present [dumps](ftp://ftp.musicbrainz.org/pub/musicbrainz/critiquebrainz/dump) (Only the sequence of the columns in the `review` and `license`  have changed). Thus, the dumps [here](ftp://ftp.musicbrainz.org/pub/musicbrainz/critiquebrainz/dump) need to be updated as the `import` function is dependent on how the dumps are created.
Also I have updated the `model.__version__` to `db.SCHEMA_VERSION` and changed `db.SCHEMA_VERSION` to `6` (the present value of `model.__version__`). Is it preferred this way?

critiquebrainz/db/review.py
```diff
@@ -575,3 +575,22 @@ def delete(review_id):
...
+    Returns:
+        List of dictionaries of ID(s) of entities reviewed.
+        {
+            "entity_id": (uuid),
+        }
+    """
+    with db.engine.connect() as connection:
+        results = connection.execute(sqlalchemy.text("""
+            SELECT
+          DISTINCT entity_id
```
gentlecat: I might be a better to write it as `SELECT DISTINCT entity_id` on one line so that it's obvious what's being selected. It got me confused for a moment.

The goal is not to just split up by keywords, but to split up logical parts of a statement so that it's easier to read.

critiquebrainz/data/dump_manager.py
```diff
@@ -269,47 +321,68 @@ def importer(archive):
...
-            db_connection.commit()
-    except IOError as exception:
-        if exception.errno == errno.ENOENT:
-            print("Can't find data file for %s table. Skipping." % model.__tablename__)
-        else:
-            sys.exit("Failed to open data file. Error: %s" % exception)
+def import_data(file_name, tablename, columns=None):
+
+    if tablename == "user":
+        tablename = '"user"'
```
gentlecat: Would it make sense to do that for all table names? That seems a bit too specific.

critiquebrainz/data/dump_manager.py
```diff
@@ -269,47 +321,68 @@ def importer(archive):
...
+            if columns:
+                result = subprocess.call("""
+                    psql -h {} -p {} -U {} -d {} -a -c '\COPY {tablename}({columns}) FROM {file_name}'
+                """.format(db_hostname, db_port, db_username, db_password, tablename=tablename,
+                        columns=", ".join(list(columns)), file_name=file_name), shell=True)
+                if result != 0:
+                    raise Exception("Failed to copy data for table %s" % tablename)
+            else:
+                result = subprocess.call("""
+                    psql -h {} -p {} -U {} -d {} -a -c '\COPY {tablename} FROM {file_name}'
```
gentlecat: Does this work without specification of columns?

critiquebrainz/db/__init__.py
```diff
@@ -2,7 +2,7 @@
 from sqlalchemy.pool import NullPool
 
 # This value must be incremented after schema changes on exported tables!
-SCHEMA_VERSION = 1
+SCHEMA_VERSION = 6
```
ferbncode: I thought of updating the `db.SCHEMA_VERSION` to `6` as the previous value of `model.__version__` was `6`. I have changed it back to `1`.

ferbncode:
I have modified the PR and updated it to use raw database connections and cursors. It surely makes sense to use them instead of importing via `psql`😅.

critiquebrainz/data/dump_manager.py
ferbncode: Yes, it does, if data for all the columns of a table is present in the file and columns are in the same order as the table structure. The `public` function for making db dumps creates files ensuring the same order of columns.

critiquebrainz/data/dump_manager.py
```diff
@@ -93,26 +92,26 @@ def json(location, rotate=False):
...
-        safe_name = slugify(license.id)
+    for license in db.license.list_licenses():
+        safe_name = slugify(license["id"])
         with tarfile.open(os.path.join(location, "critiquebrainz-%s-%s-json.tar.bz2" %
                 (datetime.today().strftime('%Y%m%d'), safe_name)), "w:bz2") as tar:
             temp_dir = tempfile.mkdtemp()
             license_dir = os.path.join(temp_dir, safe_name)
             create_path(license_dir)
 
             # Finding release groups that have reviews with current license
```
gentlecat: Might be a good idea to update that comment since you are here. It's not just "release groups" anymore.

critiquebrainz/data/dump_manager.py
```diff
@@ -93,26 +92,26 @@ def json(location, rotate=False):
     current_app.json_encoder = DumpJSONEncoder
 
     print("Creating new archives...")
-    for license in model.License.query.all():
-        safe_name = slugify(license.id)
+    for license in db.license.list_licenses():
```
gentlecat: `db.license`, `db.review`, and others need to be explicitly imported.
```python
from critiquebrainz.db import license...
```

critiquebrainz/data/dump_manager.py
```diff
@@ -189,19 +189,25 @@ def public(location, rotate=False):
 
     # 1. COMBINED
     # Archiving all reviews (any license)
-    REVISION_COMBINED_SQL = "SELECT %s FROM revision JOIN review " \
+    REVISION_COMBINED_SQL = "SELECT revision.id, revision.review_id, timestamp, text " \
```
gentlecat: As far as I know consistent order of columns during export and import is important. That's what `get_columns` function handles. If the order is indeed important then that function should be replaced with something that would provide a similar functionality.

critiquebrainz/data/dump_manager.py
```diff
@@ -269,47 +282,59 @@ def importer(archive):
...
+        import_data(os.path.join(temp_dir, 'cbdump', 'revision'), 'revision')
 
         shutil.rmtree(temp_dir)  # Cleanup
         print("Done!")
 
 
-def import_data(file_name, model, columns=None):
-    db_connection = db.session.connection().connection
-    cursor = db_connection.cursor()
+def import_data(file_name, tablename, columns=None):
```
gentlecat: `table_name`

critiquebrainz/data/dump_manager.py
```diff
@@ -269,47 +282,59 @@ def importer(archive):
...
             else:
                 sys.exit("Failed to open SCHEMA_SEQUENCE file. Error: %s" % exception)
 
         # Importing data
-        import_data(os.path.join(temp_dir, 'cbdump', 'user_sanitised'), model.User,
-                    ('id', 'created', 'display_name', 'musicbrainz_id'))
-        import_data(os.path.join(temp_dir, 'cbdump', 'license'), model.License)
-        import_data(os.path.join(temp_dir, 'cbdump', 'review'), model.Review)
-        import_data(os.path.join(temp_dir, 'cbdump', 'revision'), model.Revision)
+        import_data(os.path.join(temp_dir, 'cbdump', 'user_sanitised'), '"user"',
```
gentlecat: I don't think `'"user"'` is the best way to make sure that keywords are properly quoted. When I use `import_data` function I probably shouldn't worry about that kind of thing. What if all column names are actually quoted like it was done before?
```python
cursor.copy_from(f, '"%s"' % model.__tablename__, columns=columns)
```

critiquebrainz/data/dump_manager.py
```diff
@@ -269,47 +282,59 @@ def importer(archive):
...
+
+        # Checking if table already contains any data
+        cursor.execute("""
+            SELECT COUNT(*)
+              FROM {tablename}
+        """.format(tablename=tablename))
+        count = cursor.fetchone()[0]
+        if count > 0:
+            print("Table %s already contains data. Skipping." % tablename)
+            return
```
gentlecat: This part can be moved into a separate function so that the checking is done like this:
```python
if has_data(table):
    print("Table %s already contains data. Skipping." % table)
    return
```

critiquebrainz/db/__init__.py
gentlecat: I see. I didn't quite understand where that number came from at first, which is why I asked. It should be the same as it was before actually (`6`). It needs to be like that so that there are no compatibility issues when someone tries to import data dump generated from an old schema.

gentlecat:
OK, looks good now. Let me just try that with an updated base image.

PR #105, opened 2017-05-29
CB-237: Make data import work in local setup

ferbncode:
Make data import using dumps work in development setup by installing `libbz2-dev` and rebuilding python from source.

### DIFF
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -1,5 +1,20 @@
 FROM metabrainz/python:3.6
 
+ RUN set -x \
+	&& buildDeps='curl gcc libbz2-dev libc6-dev libsqlite3-dev libssl-dev make xz-utils zlib1g-dev' \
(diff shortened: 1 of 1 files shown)


ferbncode:
Importing dumps was not working due to missing module `bz2`. The following error is raised: https://gist.github.com/ferbncode/9eb0dd777c9f8a2d423f47cedce15bfa. 

A workaround to this problems based on https://stackoverflow.com/questions/12806122/missing-python-bz2-module and https://github.com/docker-library/python/issues/39 was to install libbz2-dev and other python dependencies, and rebuild python from source. This solves the issue and the data import from dumps work.

alastair:
This seems a bit heavy-handed to get bzip compression available in our setup.

Can you first check if this works using the standard python image available on docker hub? If yes, I propose that we merge these changes upstream into https://github.com/metabrainz/docker-python instead

ferbncode:
The import works using the standard python image available on docker hub. I opened a pull request on https://github.com/metabrainz/docker-python for fixing the `ImportError` for module `bz2`.

PR #107, opened 2017-06-03
CB-230: Migrate completely off the ORM

ferbncode:
Removed all ORM related code and made changes in `frontend.create_app` for not using ORM.


gentlecat:
Awesome. Thank you for helping with that!

PR #108, opened 2017-06-16
CB-231: Update documentation for setting up MB database

ferbncode:
Updated the documentation for setting up the MusicBrainz database for development and thus, added instructions for setting up by mounting a host directory containing dumps or by downloading the dumps.


ferbncode:
On second thought, this should not be in a seperate PR and should be with the MusicBrainz setup changes. So, closing this.

PR #109, opened 2017-06-16
CB-244: Schema changes for the rating system

ps2611:
- Add rating column to revision table to store rating
- Create new table avg_rating to store average rating of an entity


gentlecat:
@brainzbot, add to whitelist.

admin/sql/create_tables.sql
```diff
@@ -67,7 +67,17 @@ CREATE TABLE revision (
     id          SERIAL      NOT NULL,
     review_id   UUID,
     "timestamp" TIMESTAMP   NOT NULL,
-    text        VARCHAR     NOT NULL
+    text        VARCHAR,     
+    rating      SMALLINT CHECK (rating >= 0 AND rating <= 100)
```
gentlecat: I don't think it's a good idea to make the scale "incorrect" from the start. I think it's better to match it with the way it's going to be used.

Migration of ratings from MusicBrainz is not going to happen any time soon. When it does, we'll be able to easily convert ratings to another scale for storage, if that's even needed. But I'd rather keep it the way it's supposed to be until we need to make these changes.

gentlecat: Also, is it `rating >= 0` or `rating >= 1`?

admin/sql/create_tables.sql
```diff
@@ -67,7 +67,17 @@ CREATE TABLE revision (
...
+);
+ALTER TABLE revision ADD CONSTRAINT revision_text_rating_both_not_null_together
+    CHECK (rating is NOT NULL OR text is NOT NULL);
+
+CREATE TABLE avg_rating (
+    entity_id   UUID         NOT NULL,
+    entity_type entity_types NOT NULL,
+    rating      SMALLINT     NOT NULL CHECK (rating >= 0 AND rating <= 100),
```
gentlecat: How it this value going to be updated?

gentlecat:
You might also want to reflect the title of the pull request to indicate which part of rating system is being added since it's not a complete implementation.

admin/sql/create_tables.sql
ps2611: So, it will be rating >= 1 and rating <=5 ( on a scale of 5 as decided in the IRC discussions)

admin/sql/create_tables.sql
ps2611: It will be done as a part of db script. Every time, there's a revision with new rating, it will be updated.

gentlecat: It might be easier to use that with a database trigger instead. I believe @mwiencek suggested that before. Let us know if you need help with that.

admin/sql/create_tables.sql
mwiencek: > Migration of ratings from MusicBrainz is not going to happen any time soon. When it does, we'll be able to easily convert ratings to another scale for storage, if that's even needed.

I don't think that's a good reason to actively make it harder to. Converting to MB's scale in the future will require additional engineering time (code changes, a schema change and upgrade script, downtime), when you can just support it now for free. And it will be needed, because you can't express those ratings as an integer from 1-5 (if this were `NUMERIC(3, 2)` it'd be fine, but you'd be storing a bunch of extra bytes per row for no reason).

gentlecat: Do we even have ratings in MB that are not either 20, 40, 60, 80, or 100? My main concern is that scales are different and I don't know how we'll support other values set through the WS after they are imported into CB. Having https://tickets.metabrainz.org/browse/CB-248 implemented might help with this.

I'm fine with having the same storage as MBS does, but it doesn't seem right to me.

mwiencek: They exist, but not in huge numbers (436 release group ratings, 708 recording ratings not on a scale of 1-5).

MB's CSS will display fractional stars just fine (it's needed for the average ratings), though if you click on the stars it can only do 1-5. That said, I don't think that makes 0-100 "incorrect"; we could just as easily decide the number of stars we display is incorrect, or trivially modify the UI to support setting half-stars, etc.

Note that it's possible in MB to give something 0 stars, so there are 101 possible ratings.

gentlecat: OK, let's use the same way of storing them then.

critiquebrainz/db/avg_rating.py
```diff
@@ -0,0 +1,152 @@
...
+	with db.engine.connect() as connection:
+		connection.execute(sqlalchemy.text("""
+			DELETE 
+			  FROM avg_rating
+			 WHERE entity_id = :entity_id
+		"""), {
+			"entity_id": entity_id,
+		})
+
+def get_avg_rating(entity_id):
```
gentlecat: It would be better to name this function to simply `get` to be consistent with the rest in this module. It shouldn't be confusing what's being retrieved since it's in the name of the module.

gentlecat: Also, you probably need an `entity_type` argument since it's another PK.

gentlecat:
Can you put changes to the database access code (apart from `critiquebrainz/db/avg_rating.py`) into a separate pull request based on this one? Let's take this one step at a time and try not to add too many things too quickly.

critiquebrainz/db/avg_rating.py
ps2611: Changes noted.

ps2611:
I will make a fresh PR with upsert statement and the test script for avg_rating.

PR #110, opened 2017-06-19
CB-231: Use direct database access for entity: place

ferbncode:
Use direct database access for entity `place`. I have added functions `entity_relation_helper` and `get_place_by_id` for fetching information regarding `place`.

### DIFF
--- a/critiquebrainz/frontend/external/musicbrainz_db/includes.py
+++ b/critiquebrainz/frontend/external/musicbrainz_db/includes.py
@@ -1,0 +1,16 @@
+import critiquebrainz.frontend.external.musicbrainz_db.exceptions as mb_exceptions
+
+RELATABLE_TYPES = ['area', 'artist', 'label', 'place', 'event', 'recording', 'release', 'release-group', 'series', 'url', 'work', 'instrument']
+RELATION_INCLUDES = [entity + '-rels' for entity in RELATABLE_TYPES]
+TAG_INCLUDES = ["tags", "user-tags"]
+RATING_INCLUDES = ["ratings", "user-ratings"]
+VALID_INCLUDES = {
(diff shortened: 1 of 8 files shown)


gentlecat:
Have you considered using mbdata package for this purpose?

critiquebrainz/frontend/external/musicbrainz_db/includes.py
```diff
@@ -0,0 +1,16 @@
+import critiquebrainz.frontend.external.musicbrainz_db.exceptions as mb_exceptions
+
+RELATABLE_TYPES = ['area', 'artist', 'label', 'place', 'event', 'recording', 'release', 'release-group', 'series', 'url', 'work', 'instrument']
```
gentlecat: This row is a bit too long. Can you split it like this?
```python
RELATABLE_TYPES = [
    'area',
    ...
    'instrument',
]
```

critiquebrainz/frontend/external/musicbrainz_db/place.py
```diff
@@ -0,0 +1,50 @@
...
+from critiquebrainz.frontend.external.musicbrainz_db.serialize import to_dict_places
+from critiquebrainz.frontend.external.musicbrainz_db.helpers import entity_relation_helper
+from brainzutils import cache
+
+
+DEFAULT_CACHE_EXPIRATION = 12 * 60 * 60 # seconds (12 hours)
+THREAD_POOL_PROCESSES = 10
+
+
+def get_place_by_id(id):
```
gentlecat: It's better to be more specific here and use `mbid` instead of `id`. I usually try to avoid using `id` since it's a built in function.

gentlecat:
Haven't been able to test this due to problems with the MusicBrainz database image setup. https://chatlogs.metabrainz.org/brainzbot/metabrainz/2017-06-26/?msg=3940782&page=2

PR #113, opened 2017-06-26
Fix missing table name in function users.tokens

ferbncode:
Sorry for missing this in the initial PR.


ferbncode:
Got it. Will do it. Thanks :)

PR #116, opened 2017-06-30
CB-258: Define column sequence in INSERT sql statements.

ferbncode:
Added column sequences in `INSERT` sql statements to prevent issues due to different column sequences in database.


ferbncode:
IIRC, I may have seen a similar issue on production site when creating a user application on the `/profile/applications` page. Sorry for all the mess but we may want to fix the column sequence in all the `INSERT` queries.

gentlecat:
Don't worry, keeping track of what's happening in production is more difficult for you. Just keep that in mind when you are writing SQL statements in the future. It's better not to make assumptions and specify things explicitly in general.

I already merged your change in 710e3f5f78f16ac0063ac4557b09e352a63c7069.

PR #121, opened 2017-07-05
CB-231: Update db access function and helpers for multiple entities.

ferbncode:
Updated `entity_relation_helper` and `get_place_by_id` to fetch info for multiple entities. Function `fetch multiple places` can be used to fetch info of multple places.

### DIFF
--- a/critiquebrainz/frontend/external/musicbrainz_db/helpers.py
+++ b/critiquebrainz/frontend/external/musicbrainz_db/helpers.py
@@ -11,10 +11,7 @@ def entity_relation_helper(db, target_type, source_type, source_entity_ids, incl
         source_type (str): Type of source entity.
         source_entity_ids (list): IDs of the source entity.
         includes_data (dict): Dictionary containing includes data of entities.
-
-    Returns:
-        List of objects of relationships between entities.
-    """
+   """
     source_model = get_entity_type_model(source_type)
     target_model = get_entity_type_model(target_type)
     relation = get_link_model(source_model, target_model)
@@ -23,16 +20,16 @@ def entity_relation_helper(db, target_type, source_type, source_entity_ids, incl
         options(joinedload("link", innerjoin=True)).\
         options(joinedload("link.link_type", innerjoin=True))
     if relation.entity0.property.mapper.class_ == relation.entity1.property.mapper.class_:
-        get_relationship_links(relation, query, "entity0", "entity1", target_type, source_entity_ids, includes_data)
-        get_relationship_links(relation, query, "entity1", "entity0", target_type, source_entity_ids, includes_data)
+        _relationship_link_helper(relation, query, "entity0", "entity1", target_type, source_entity_ids, includes_data)
+        _relationship_link_helper(relation, query, "entity1", "entity0", target_type, source_entity_ids, includes_data)
     else:
         if source_model == relation.entity0.property.mapper.class_:
-            get_relationship_links(relation, query, "entity0", "entity1", target_type, source_entity_ids, includes_data)
+            _relationship_link_helper(relation, query, "entity0", "entity1", target_type, source_entity_ids, includes_data)
         else:
-            get_relationship_links(relation, query, "entity1", "entity0", target_type, source_entity_ids, includes_data)
+            _relationship_link_helper(relation, query, "entity1", "entity0", target_type, source_entity_ids, includes_data)
 
 
-def get_relationship_links(relation, query, source_attr, target_attr, target_type, source_entity_ids, includes_data):
+def _relationship_link_helper(relation, query, source_attr, target_attr, target_type, source_entity_ids, includes_data):
     """Get relationship links between two entities.
 
     Args:
@@ -43,11 +40,15 @@ def get_relationship_links(relation, query, source_attr, target_attr, target_typ
         target_type (str): Type of the target entity.
         source_entity_ids (list): IDs of the source entity.
         includes_data (dict): Dictionary containing the includes data of entities.
+
+    Note:
+        includes_data (dict) is altered to contain the relationship objects
+        keyed by the source entity mbids
     """
     source_id_attr = source_attr + "_id"
     query = query.filter(getattr(relation, source_id_attr).in_(source_entity_ids))
     query = query.options(joinedload(target_attr, innerjoin=True))
     relation_type = target_type + "-rels"
     for link in query:
-        includes_data.setdefault(getattr(link, source_id_attr), {}).setdefault('relationship_objs', {}).\
+        includes_data[getattr(link, source_id_attr)].setdefault('relationship_objs', {}).\
             setdefault(relation_type, []).append(link)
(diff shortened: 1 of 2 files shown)


critiquebrainz/frontend/external/musicbrainz_db/place.py
```diff
@@ -23,28 +23,42 @@ def get_place_by_id(mbid):
...
+        place = fetch_multiple_places(
+            mbids=[mbid],
+            includes=['artist-rels', 'place-rels', 'release-group-rels', 'url-rels'],
+        ).get(mbid)
     cache.set(key=key, val=place, time=DEFAULT_CACHE_EXPIRATION)
     return place
 
 
-def _get_place_by_id(place_id, includes=[]):
+def fetch_multiple_places(*, mbids, includes=[]):
```
paramsingh: I am not very familiar with the code, but default mutable values for function arguments (like `includes=[]` here) can cause problems. https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

critiquebrainz/frontend/external/musicbrainz_db/place.py
```diff
@@ -23,28 +23,43 @@ def get_place_by_id(mbid):
...
+        for mbid in mbids:
+            place = get_something_by_gid(query, models.PlaceGIDRedirect, mbid)
+            if not place:
+                raise mb_exceptions.NoDataFoundException("Couldn't find a place with id: {mbid}".format(mbid=mbid))
+            places.append(place)
+        place_ids = [place.id for place in places]
 
         if 'artist-rels' in includes:
-            includes_data.setdefault('relationship_objs', {})['artist-rels'] = entity_relation_helper(db, 'artist', 'place', place.id)
+            entity_relation_helper(db, 'artist', 'place', place_ids, includes_data)
```
alastair: This is a little complex to understand - I think that you're passing in `includes_data` as an argument and it's getting populated in `get_relationship_links`. Is that correct?
If that is the case, perhaps we shouldn't call the function `get_.....`. The documentation should be updated to say that this value is altered, instead of being passed in as an argument.

critiquebrainz/frontend/external/musicbrainz_db/helpers.py
```diff
@@ -23,31 +24,31 @@ def entity_relation_helper(db, target_type, source_type, source_entity_id):
...
+        includes_data (dict): Dictionary containing the includes data of entities.
     """
     source_id_attr = source_attr + "_id"
-    query = query.filter(getattr(relation, source_id_attr) == source_entity_id)
+    query = query.filter(getattr(relation, source_id_attr).in_(source_entity_ids))
     query = query.options(joinedload(target_attr, innerjoin=True))
-    return query.all()
+    relation_type = target_type + "-rels"
+    for link in query:
+        includes_data.setdefault(getattr(link, source_id_attr), {}).setdefault('relationship_objs', {}).\
```
alastair: Sometimes it's useful to use `collections.defaultdict` for things like this, where you want to be able to add arbitrary nested objects. Note that this might not serialise nicely in your redis cache and so you might need to convert it before you cache it.

critiquebrainz/frontend/external/musicbrainz_db/place.py
ferbncode: Yes, `includes_data` is getting populated in `get_relationship_links`. I have changed the name of the function to `_relationship_links_helper` (the function would not be used outside this module). Also, as suggested, I added a `Note` in the docstring stating that `includes_data` is altered to contain the relationship objects keyed by the mbid.

critiquebrainz/frontend/external/musicbrainz_db/helpers.py
ferbncode: Thanks. Updated as suggested.

critiquebrainz/frontend/external/musicbrainz_db/place.py
```diff
@@ -23,28 +23,43 @@ def get_place_by_id(mbid):
...
+    """Get info related to multiple places using their MusicBrainz IDs.
+
+    Args:
+        mbids (list): List of MBIDs of places.
+        includes (list): List of information to be included.
+
+    Returns:
+        Dictionary containing info of multiple places keyed by their mbid.
+    """
+    includes = dict() if includes is None else includes
```
ferbncode: Sorry, my bad. Fixed.

gentlecat:
Apologies go breaking the tests and causing conflicts, but there were some important issues I had to fix. Please rebase this on top of latest *master* when you can.

Thanks!

critiquebrainz/frontend/external/musicbrainz_db/helpers.py
```diff
@@ -2,18 +2,16 @@
 from sqlalchemy.orm import joinedload
 
 
-def entity_relation_helper(db, target_type, source_type, source_entity_id):
+def entity_relation_helper(db, target_type, source_type, source_entity_ids, includes_data):
```
gentlecat: Would a name like `get_relationship_info` be more informative for this function?

Also if this function doesn't return anything, what does it do? Update one of the arguments passed to it? Can you describe this (in the docstring please)?

critiquebrainz/frontend/external/musicbrainz_db/helpers.py
```diff
@@ -22,30 +20,35 @@ def entity_relation_helper(db, target_type, source_type, source_entity_id):
...
         source_attr (str): 'entity0' or 'entity1' based on which represents source model in relation table.
         target_attr (str): 'entity0' or 'entity1' based on which represents target model in relation table.
-        source_entity_id (int): ID of the source entity.
+        target_type (str): Type of the target entity.
+        source_entity_ids (list): IDs of the source entity.
+        includes_data (dict): Dictionary containing the includes data of entities.
 
-    Returns:
-        List of objects of relationships between entities.
+    Note:
```
gentlecat: This is not really a standard way of writing an extended description. Check https://github.com/metabrainz/guidelines/blob/master/Python.md#docstrings. Here's how you could write this:
```python
    """Get relationship links between two entities.

    Keep in mind that includes_data (dict) is altered to contain the relationship objects
    keyed by the source entity MBIDs.

    Args:
        relation (mbdata.model): Model relating the two entities.
        query (Session.query): Query object.
        source_attr (str): 'entity0' or 'entity1' based on which represents source model in relation table.
        target_attr (str): 'entity0' or 'entity1' based on which represents target model in relation table.
        target_type (str): Type of the target entity.
        source_entity_ids (list): IDs of the source entity.
        includes_data (dict): Dictionary containing the includes data of entities.
    """
```

critiquebrainz/frontend/external/musicbrainz_db/place.py
```diff
@@ -24,30 +25,44 @@ def get_place_by_id(mbid):
...
+        place = fetch_multiple_places(
+            mbids=[mbid],
+            includes=['artist-rels', 'place-rels', 'release-group-rels', 'url-rels'],
+        ).get(mbid)
     cache.set(key=key, val=place, time=DEFAULT_CACHE_EXPIRATION)
     return place_rel.process(place)
 
 
-def _get_place_by_id(place_id, includes=None):
+def fetch_multiple_places(*, mbids, includes=None):
```
gentlecat: In this case `mbids` can actually be made not keyword-only. It's clear how `fetch_multiple_places([<mbid_1>, <mbid_2>])` would work.

gentlecat:
@alastair, can you please check the changes?

PR #122, opened 2017-07-05
CB-231: Use direct database access for the Release Group entity

ferbncode:
Added `fetch_multiple_release_group` and `get_release_group_by_id` for fetching release group info.


critiquebrainz/frontend/external/musicbrainz_db/serialize.py
```diff
@@ -83,19 +83,56 @@ def to_dict_places(place, includes={}):
...
+    data = {
+        'name': artist_credit_name.name,
+        'artist': to_dict_artists(artist_credit_name.artist),
+    }
+    if artist_credit_name.join_phrase:
+        data['join_phrase'] = artist_credit_name.join_phrase
+    return data
+
+
 def to_dict_release_groups(release_group, includes={}):
```
alastair: I think that we should have tests for these conversion methods to make sure that we correctly create a dict from a particular releasegroup id.
I think that this will be a bit difficult - we can't really do the query against a real musicbrainz database because the data might change in the database. I'm not sure that it's worth testing only this function, I'd like to see it tested in conjunction with `fetch_multiple_release_groups`
Maybe @gentlecat has a suggestion here?

gentlecat: I'd definitely like to see some tests for all of this.

ferbncode: Thanks. Regarding the test database, we can make use ot `mbdata.create_sample_data()`. I've started some work here > https://github.com/ferbncode/critiquebrainz/commit/a27236eebacf8fd83efbcab9c993e3ee47837582 (works, but still WIP). I can make a list of testable entities (entities in [sample_data.py](https://bitbucket.org/lalinsky/mbdata/src/fd6a391d3d1c0a3085904ae65a50129a34cd1a51/mbdata/sample_data.py?at=master))

Also, these test (may) fail in case of schema_changes (in musicbrainz database). But I think that would be good for us?

ferbncode: Sorry, read the backlogs later (on irc) about the script in `musicbrainz-server` (that way might be better for testing different type of entities)

gentlecat:
@ferbncode, I assume you are working on tests with MB database now. When you finish that part, put it into a separate PR please so that it's more manageable. ;)

critiquebrainz/frontend/external/musicbrainz_db/release_group.py
```diff
@@ -0,0 +1,69 @@
...
+        for mbid in mbids:
+            release_group = get_something_by_gid(query, models.ReleaseGroupGIDRedirect, mbid)
+            if not release_group :
+                raise mb_exceptions.NoDataFoundException("Couldn't find a release group with id: {mbid}".format(mbid=mbid))
+            release_groups.append(release_group)
+        release_group_ids = [release_group.id for release_group in release_groups]
+
+        if 'artists' in includes:
+            for release_group in release_groups:
+                artist_credit_names = release_group.artist_credit.artists
```
ferbncode: As suggested, I've added a `get_entities_by_gid` function for fetching multiple entities. Also, added `joinedload` fqor `artist_credit_names`, `artists` and `release_group_meta` for saving db queries. A single `fetch_` method now would take 7 queries to the db (for any number of entities). (Previously, it was making 5 queries per mbid (seperate ones for `artist_credit_name`, `artist_credit`, `meta`, `artist_credit_name` and `release_group`)  + 6 more for relationships, releases and tags). Thanks for suggesting that :)

critiquebrainz/frontend/external/musicbrainz_db/release_group.py
```diff
@@ -0,0 +1,69 @@
...
+from critiquebrainz.frontend.external.musicbrainz_db import mb_session
+from sqlalchemy.orm import joinedload
+from critiquebrainz.frontend.external.musicbrainz_db.helpers import entity_relation_helper
+from critiquebrainz.frontend.external.musicbrainz_db.serialize import to_dict_release_groups
+import critiquebrainz.frontend.external.musicbrainz_db.exceptions as mb_exceptions
+import critiquebrainz.frontend.external.relationships.release_group as release_group_rel
+from brainzutils import cache
+
+
+DEFAULT_CACHE_EXPIRATION = 12 * 60 * 60 # seconds (12 hours)
```
ferbncode: All entities are updated after the same time. Updated as suggested.

critiquebrainz/frontend/external/musicbrainz_db/release_group.py
```diff
@@ -0,0 +1,69 @@
...
+        if 'release-group-rels' in includes:
+            entity_relation_helper(db, 'release_group', 'release_group', release_group_ids, includes_data)
+
+        if 'url-rels' in includes:
+            entity_relation_helper(db, 'url', 'release_group', release_group_ids, includes_data)
+
+        if 'work-rels' in includes:
+            entity_relation_helper(db, 'work', 'release_group', release_group_ids, includes_data)
+
+        if 'tags' in includes:
```
ferbncode: As suggeted, added a `get_tags` function for fetching entity tags (as it may be used for other entities like `artists`)

critiquebrainz/frontend/external/musicbrainz_db/release_group.py
```diff
@@ -0,0 +1,94 @@
...
+def get_release_group_by_id(mbid):
+    """Get release group using the MusicBrainz ID."""
+    key = cache.gen_key(mbid)
+    release_group = cache.get(key)
+    if not release_group:
+        release_group = fetch_multiple_release_groups(
+            [mbid],
+            includes=['artists', 'releases', 'release-group-rels', 'url-rels', 'work-rels', 'tags']
+        )[mbid]
+    cache.set(key=key, val=release_group, time=DEFAULT_CACHE_EXPIRATION)
```
gentlecat: This should be inside the `if not release_group:` block, otherwise it's being unnecessarily reset each time.

critiquebrainz/frontend/external/musicbrainz_db/release_group.py
```diff
@@ -0,0 +1,94 @@
...
+
+def get_release_group_by_id(mbid):
+    """Get release group using the MusicBrainz ID."""
+    key = cache.gen_key(mbid)
+    release_group = cache.get(key)
+    if not release_group:
+        release_group = fetch_multiple_release_groups(
+            [mbid],
+            includes=['artists', 'releases', 'release-group-rels', 'url-rels', 'work-rels', 'tags']
+        )[mbid]
```
gentlecat: It might be better to create a wrapper function for `fetch_multiple_release_groups` that retrieves only one release group at a time.

critiquebrainz/frontend/external/musicbrainz_db/release_group.py
```diff
@@ -0,0 +1,94 @@
...
+    cache.set(key=key, val=release_group, time=DEFAULT_CACHE_EXPIRATION)
+    return release_group_rel.process(release_group)
+
+
+def fetch_multiple_release_groups(mbids, *, includes=None):
+    includes = [] if includes is None else includes
+    includes_data = defaultdict(dict)
+    check_includes('release_group', includes)
+    with mb_session() as db:
+        query = db.query(models.ReleaseGroup).options(joinedload("meta"))
```
gentlecat: What does `.options(joinedload("meta"))` part do? It might be worth making a comment clarifying parts like these that might be more difficult to understand for someone unfamiliar with how mbdata works.

critiquebrainz/frontend/external/musicbrainz_db/release_group.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if 'artists' in includes:
+            query = query.options(joinedload("artist_credit")).\
+                    options(joinedload("artist_credit.artists")).\
+                    options(joinedload("artist_credit.artists.artist"))
+
+        release_groups = get_entities_by_gids(query, models.ReleaseGroup, models.ReleaseGroupGIDRedirect, mbids)
```
gentlecat: In a case like this it would be better to specify names of the arguments explicitly. When I look at `get_entities_by_gids(query, models.ReleaseGroup, models.ReleaseGroupGIDRedirect, mbids)` I have no idea what these models are unless I look up the function definition and its docstring/implementation.

critiquebrainz/frontend/templates/macros.html
```diff
@@ -96,7 +96,7 @@
 {% macro show_tags(tags) %}
   {% if tags %}
     {% for tag in tags %}
-      <a href="{{ 'https://musicbrainz.org/tag/%s'|format(tag['name']) }}">{{ tag['name'] }}</a>
+      <a href="{{ 'https://musicbrainz.org/tag/%s'|format(tag) }}">{{ tag }}</a>
```
gentlecat: Is there any reason to use `format` instead of just outputting the value directly into the template?

critiquebrainz/frontend/external/musicbrainz_db/includes.py
```diff
@@ -18,7 +18,8 @@
 TAG_INCLUDES = ["tags", "user-tags"]
 RATING_INCLUDES = ["ratings", "user-ratings"]
 VALID_INCLUDES = {
-    'place': ["aliases", "annotation"] + RELATION_INCLUDES + TAG_INCLUDES,
+    'place' : ["aliases", "annotation"] + RELATION_INCLUDES + TAG_INCLUDES,
```
gentlecat: This change looks unnecessary.

critiquebrainz/frontend/external/musicbrainz_db/place.py
```diff
@@ -25,14 +21,19 @@ def get_place_by_id(mbid):
...
+        cache.set(key=key, val=place, time=DEFAULT_CACHE_EXPIRATION)
     return place_rel.process(place)
 
 
+def _get_place_by_id(mbid):
+    place = fetch_multiple_places(
+        [mbid],
+        includes=['artist-rels', 'place-rels', 'release-group-rels', 'url-rels'],
+    )[mbid]
+    return place
```
gentlecat: Can simply `return fetch_multiple_places(...` here.

critiquebrainz/frontend/views/test/test_release_group.py
```diff
@@ -1,12 +1,20 @@
...
+        }
+
     def test_release_group_page(self):
-        # Basic release group page should be available.
-        response = self.client.get("/release-group/c2e0ff67-fb31-4443-ae0e-22ecf010463b")
+        response = self.client.get('/release-group/8ef859e3-feb2-4dd1-93da-22b91280d768')
         self.assert200(response)
-        self.assertIn("Days Are Gone", str(response.data))
-        self.assertIn("No reviews found", str(response.data))
-        # TODO(roman): Try to add review and check it's displayed there!
```
gentlecat: Is this fixed now?

critiquebrainz/frontend/templates/macros.html
```diff
@@ -96,7 +96,7 @@
 {% macro show_tags(tags) %}
   {% if tags %}
     {% for tag in tags %}
-      <a href="{{ 'https://musicbrainz.org/tag/%s'|format(tag['name']) }}">{{ tag['name'] }}</a>
+    <a href="{{ 'https://musicbrainz.org/tag/%s' % tag }}">{{ tag }}</a>
```
gentlecat: Can you fix indentation here please?

gentlecat: Also, you are still using formatting. I meant this:
```jinja2
<a href="https://musicbrainz.org/tag/{{ tag }}">{{ tag }}</a>
```

critiquebrainz/frontend/views/mapping.py
```diff
@@ -147,12 +147,11 @@ def spotify_report():
...
-        try:
-            album = spotify_api.get_album(spotify_id)
-        except ExternalServiceException:
-            flash.error(gettext("You need to specify existing album from Spotify!"))
-            return redirect(url_for('.spotify_list', release_group_id=release_group_id))
+    try:
+        album = spotify_api.get_album(spotify_id)
+    except ExternalServiceException:
+        flash.error(gettext("You need to specify existing album from Spotify!"))
+        return redirect(url_for('.spotify_list', release_group_id=release_group_id))
```
gentlecat: I don't understand this change. Can you explain?

ferbncode: I was getting https://gist.github.com/ferbncode/307bc3f3e13f5d65428aad6180ace1ed error when running pylint. I have undone this change now as it may be done separately.

ferbncode:
No, changes in `place.py` adding the wrapper `_get_place_by_id` for `fetch_multiple_places` is not a part for release group stuff. I removed that part and will make a separate PR for the same.

critiquebrainz/frontend/templates/macros.html
ferbncode: Fixed :sweat_smile:

gentlecat:
> Should they be done in a separate PR too?

That's fine.

PR #123, opened 2017-07-05
CB-261: Autheticate requests to Spotify Web API

ferbncode:
Added function `fetch_access_token` for authenticating requests to Spotify Web API. Since, CB doesn't need access to spotify users' private data, I've used `Client Credential Flow` as the [authorization flow](https://developer.spotify.com/web-api/authorization-guide/#client-credentials-flow) requiring only the client credentials for fetching access token.


default_config.py
```diff
@@ -59,6 +59,10 @@
 MUSICBRAINZ_CLIENT_ID = ""
 MUSICBRAINZ_CLIENT_SECRET = ""
 
+# Spotify
+SPOTIFY_CLIENT_ID = "EXAMPLE_CLIENT_ID"
+SPOTIFY_CLIENT_SECRET = "EXAMPLE_CLIENT_SECRET"
```
gentlecat: Just set those to be empty strings.

critiquebrainz/frontend/external/spotify.py
```diff
@@ -6,12 +6,55 @@
...
+
+    if not access_token:
+        access_token = requests.post(
+            "https://accounts.spotify.com/api/token",
+            data={"grant_type": "client_credentials"},
+            headers={"Authorization": f"Basic {auth_value}"},
+        ).json()
+        access_token = access_token.get('access_token')
+        if not access_token:
+            raise Exception("Could not fetch access token for Spotify")
```
gentlecat: Try not to use `Exception` class. This makes it very inconvenient to capture since having `except Exception` is typically a bad idea. It's better to subclass it with something specific to this module/package.

critiquebrainz/frontend/external/spotify.py
```diff
@@ -6,12 +6,55 @@
...
+        cache.set(key=key, namespace=namespace, val=access_token, time=ACCESS_TOKEN_EXPIRATION)
+    return access_token
+
+
+def get_spotify(url):
+    """Make a GET request to Spotify Web API.
```
gentlecat: This is not very accurate. It currently makes a GET request to whatever URL is provided, not just from spotify.com. One thing you can do here is instead accept a path with query string and append it to the base URL of the Spotify API.

critiquebrainz/frontend/external/spotify.py
```diff
@@ -6,12 +6,55 @@
...
+from base64 import b64encode
+from flask import current_app as app
 
 DEFAULT_CACHE_EXPIRATION = 12 * 60 * 60  # seconds (12 hours)
+ACCESS_TOKEN_EXPIRATION = 60 * 60 # seconds (1 hour)
 
 BASE_URL = "https://api.spotify.com/v1"
 
 
+def fetch_access_token():
```
gentlecat: Make sure to prefix functions that are not meant to be used outside of the module with `_`.

gentlecat: `get_spotify` too?

critiquebrainz/frontend/external/spotify.py
```diff
@@ -6,12 +6,57 @@
...
+        query (str): Query to the Web API.
+
+    Returns:
+        Dictionary containing the information.
+    """
+    access_token = _fetch_access_token()
+    url = BASE_URL + query
+    headers = {"Authorization": f"Bearer {access_token}"}
+
+    result = requests.get(f"{url}", headers=headers)
```
paramsingh: `f"{url}"` 

Nice

alastair: we should probably try being consistent in our format strings, whether this is `%`, `.format`, or `f""` - my opinion is that having too many isn't a good thing. If we want to slowly move from one style to another, that's OK (I'm not sure if this has already been talked about)

In this specific case, do we actually need the format string? we can just do `requests.get(url)`

gentlecat: You are right, we don't need it.

And no, we haven't discussed this. I think f-strings are new in Python 3, so we couldn't have used it before. This inconsistency might be partially my fault since at the beginning I used `%` and then started switching to `.format` since it allows to specify placeholder names explicitly.

PR #124, opened 2017-07-06
CB-244: Rating system (Schema changes)

ps2611:
- Add rating column to revision table to store rating
- Create new table avg_rating to store average rating of an entity


gentlecat:
Can you please force push these changes into #109 instead? That's where the work on this started and there's already some discussion which we should keep in one place. There's no need to create a new pull request when you change something in the old one. If you need to make changes it's better to make another commit or force push.

PR #125, opened 2017-07-06
CB-244: Rating data access functions

ps2611:
Create db/avg_rating module for operations on avg_rating

### DIFF
--- a/critiquebrainz/db/avg_rating.py
+++ b/critiquebrainz/db/avg_rating.py
@@ -4,112 +4,113 @@ from critiquebrainz.db import exceptions as db_exceptions
 
 
 def update(entity_id, entity_type):
-	"""Updates the average rating of the entity
-	
-	It does so by collecting all the review_ids for a given entity from review
-	and selecting respective revisions with latest timestamp
-	and calculating sum and count of ratings 
-	
-	Args:
-		entity_id (uuid): ID of the entity
-		entity_type (str): Type of the entity
-	"""
-	with db.engine.connect() as connection:
-		result = connection.execute(sqlalchemy.text("""
-			WITH LatestRevisions AS (
-				SELECT review_id,
-					   MAX("timestamp") created_at
-				  FROM revision
-				 WHERE review_id in ( 
-				 	       SELECT id 
-				 	         FROM review
-				 	        WHERE entity_id = :entity_id
-				 	          AND entity_type = :entity_type)
-			  GROUP BY review_id)
-			
-			SELECT SUM(rating),
-			       COUNT(rating)
-			  FROM revision
-		INNER JOIN LatestRevisions
-		        ON revision.review_id = LatestRevisions.review_id
-		       AND revision.timestamp = LatestRevisions.created_at
-		"""),{
-			"entity_id": entity_id,
-			"entity_type": entity_type,
-		})
+    """Updates the average rating of the entity
+    
+    It does so by collecting all the review_ids for a given entity from review
+    and selecting respective revisions with latest timestamp
+    and calculating sum and count of ratings 
+    
+    Args:
+        entity_id (uuid): ID of the entity
+        entity_type (str): Type of the entity
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            WITH LatestRevisions AS (
+                SELECT review_id,
+                       MAX("timestamp") created_at
+                  FROM revision
+                 WHERE review_id in ( 
+                           SELECT id 
+                             FROM review
+                            WHERE entity_id = :entity_id
+                              AND entity_type = :entity_type)
+              GROUP BY review_id)
+            
+            SELECT SUM(rating),
+                   COUNT(rating)
+              FROM revision
+        INNER JOIN LatestRevisions
+                ON revision.review_id = LatestRevisions.review_id
+               AND revision.timestamp = LatestRevisions.created_at
+        """),{
+            "entity_id": entity_id,
+            "entity_type": entity_type,
+        })
 
-		row = result.fetchone()
-		if row is None:
-			delete(entity_id, entity_type)
-			return
-	
-	#Calulate average rating and update it	
-	sum, count = row[0], row[1]
-	avg_rating = int(sum / count + 0.5)
-	with db.engine.connect() as connection:
-		connection.execute(sqlalchemy.text("""
-	   		INSERT INTO avg_rating(entity_id, entity_type, rating, count)
-				 VALUES (:entity_id, :entity_type, :rating, :count)
-	        ON CONFLICT 
-	      ON CONSTRAINT avg_rating_pkey
-	          DO UPDATE
-	                SET rating = EXCLUDED.rating, 
-	                	count = EXCLUDED.count
-	    """),{
-			"entity_id": entity_id,
-			"entity_type": entity_type,
-			"rating": avg_rating,
-			"count": count,
-	    })
+        row = result.fetchone()
+        if row is None:
+            delete(entity_id, entity_type)
+            return
+    
+    #Calulate average rating and update it  
+    sum, count = row[0], row[1]
+    avg_rating = int(sum / count + 0.5)
+    with db.engine.connect() as connection:
+        connection.execute(sqlalchemy.text("""
+            INSERT INTO avg_rating(entity_id, entity_type, rating, count)
+                 VALUES (:entity_id, :entity_type, :rating, :count)
+            ON CONFLICT 
+          ON CONSTRAINT avg_rating_pkey
+              DO UPDATE
+                    SET rating = EXCLUDED.rating, 
+                        count = EXCLUDED.count
+        """),{
+            "entity_id": entity_id,
+            "entity_type": entity_type,
+            "rating": avg_rating,
+            "count": count,
+        })
 
 
 def delete(entity_id, entity_type):
-	"""Deletes the avg_rating, given entity_id and entity_type
+    """Deletes the avg_rating, given entity_id and entity_type
 
-	Args: 
-		entity_id (uuid): ID of the entity
-		entity_type (str): Type of the entity
-	"""
-	with db.engine.connect() as connection:
-		connection.execute(sqlalchemy.text("""
-			DELETE 
-			  FROM avg_rating
-			 WHERE entity_id = :entity_id
-			   AND entity_type = :entity_type
-		"""), {
-			"entity_id": entity_id,
-			"entity_type": entity_type,
-		})
+    Args: 
+        entity_id (uuid): ID of the entity
+        entity_type (str): Type of the entity
+    """
+    with db.engine.connect() as connection:
+        connection.execute(sqlalchemy.text("""
+            DELETE 
+              FROM avg_rating
+             WHERE entity_id = :entity_id
+               AND entity_type = :entity_type
+        """), {
+            "entity_id": entity_id,
+            "entity_type": entity_type,
+        })
 
 
 def get(entity_id, entity_type):
-	"""Get average rating from entity_id
+    """Get average rating from entity_id
 
-	Args: 
-		entity_id (uuid): ID of the entity
-		entity_type (str): Type of the entity
-	Returns:
-		Dictionary with the following structure
-		{
-			"entity_id": uuid,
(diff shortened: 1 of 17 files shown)


critiquebrainz/db/avg_rating.py
```diff
@@ -0,0 +1,116 @@
...
+        {
+            "entity_id": uuid,
+            "entity_type": str("release group", "event", "place"),
+            "rating": int,
+            "count": int,
+        }
+    """
+    with db.engine.connet() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT *
```
gentlecat: Rows should be explicitly specified there.

critiquebrainz/db/avg_rating.py
```diff
@@ -0,0 +1,116 @@
...
+             WHERE entity_id = :entity_id
+               AND entity_type = :entity_type
+        """), {
+            "entity_id": entity_id,
+            "entity_type": entity_type
+        })
+
+        avg_rating = result.fetchone()
+        if not avg_rating:
+            raise db_exceptions.NoDataFoundException("No rating for the entity with ID: {id}".format(id=entity_id))
```
gentlecat: Can also specify which entity type it is.

gentlecat:
*It seems like this cannot be merged before #128.*

ps2611:
How will it be tested then?

gentlecat:
Right now? You can rebase this on top of the branch that is used in the other PR that this one depends on.

ps2611:
The three db/review_test failures will be resolved when this will be merged after PR#128.
https://ci.metabrainz.org/job/critiquebrainz/374/#showFailuresLink

gentlecat:
I'll take a look at the code soon, but you need to make sure that all checks pass before I can merge this. Otherwise it will break production setup.

ps2611:
This PR is only for db part. So frontend and ws tests will fail. They will be changed in their respective PRs.
If this PR works fine, frontend is the next step.

critiquebrainz/db/review.py
```diff
@@ -179,7 +183,7 @@ def get_count(*, is_draft=False, is_hidden=False):
         return result.fetchone()[0]
 
 
-def update(review_id, *, drafted, text, license_id=None, language=None, is_draft=None):
+def update(review_id, *, drafted, text, rating, license_id=None, language=None, is_draft=None):
```
gentlecat: There is a problem here. The idea that we had is that review is one of the following:

* text *(how it is right now)*
* rating
* text + rating

Looking at the signature of this function it seems like both text and review are mandatory (arguments). First of all, this is not going to work with the current implementation that depends on this function. It doesn't provide rating, obviously, since that part is missing. But even without accounting for that, in the future it needs to support different variations of reviews (as I mentioned before).

What I'd do here instead is make both `text` and `rating` arguments optional, then check that at least one of those is provided.

gentlecat: Same with the `create` function.

critiquebrainz/frontend/login/__init__.py
```diff
@@ -8,9 +8,9 @@
 from flask_login import LoginManager, current_user
 from flask_babel import lazy_gettext, gettext
 from critiquebrainz.data.mixins import AnonymousUser
-from critiquebrainz.db.user import User
 import critiquebrainz.db.users as db_users
 from werkzeug.exceptions import Unauthorized
+from critiquebrainz.db.user import User
```
gentlecat: Is there a reason for moving `from critiquebrainz.db.user import User` in multiple files in this PR?

gentlecat:
*To see what specifically is failing you can click on **Details** links next to checks.*

critiquebrainz/db/review.py
ps2611: IMHO, rating or text should explicitly declared to be None if a review is just text or just rating . That way, it will be easier to maintain integrity.
And, it seems I have missed to raise exception for IntegrityError at the start of update and create function. I will add that.

critiquebrainz/frontend/login/__init__.py
ps2611: It's a workaround for ImportError because of circular import. I have changed only those files which were causing problems.

critiquebrainz/db/review.py
```diff
@@ -135,9 +137,11 @@ def get_by_id(review_id):
...
         }
         votes = db_revision.votes(review["last_revision"]["id"])
-        review["votes_positive_count"] = votes["positive"]
-        review["votes_negative_count"] = votes["negative"]
-        review["popularity"] = review["votes_positive_count"] - review["votes_negative_count"]
+        review["votes"] = {
+            "positive": votes["positive"],
+            "negative": votes["negative"],
+            "popularity": votes["positive"] - votes["negative"],
+        }
```
ps2611: No, it should be done seperately. I will remove it.

critiquebrainz/db/avg_rating_test.py
```diff
@@ -0,0 +1,103 @@
...
+            rating=100,
+            user_id=self.user.id,
+            is_draft=False,
+            license_id=self.license["id"],
+        )
+        avg_rating = db_avg_rating.get(review["entity_id"], review["entity_type"])
+        self.assertEqual(avg_rating["entity_id"], review["entity_id"])
+        self.assertEqual(avg_rating["entity_type"], review["entity_type"])
+        self.assertEqual(avg_rating["rating"], review["rating"])
+        self.assertEqual(avg_rating["count"], 1)
```
gentlecat: Instead of having multiple statements you can just compare one dictionary to another once.

gentlecat: Same in other tests.

critiquebrainz/db/exceptions.py
```diff
@@ -11,3 +11,7 @@ class NoDataFoundException(DatabaseException):
 class BadDataException(DatabaseException):
     """Should be used when incorrect data is being submitted."""
     pass
+
+class IntegrityError(DatabaseException):
```
gentlecat: I'm not sure if that's actually necessary. I'd say that `BadDataException` covers this case. It's also a different level of abstraction. Integrities and constraints are involved when working directly with the database and that shouldn't concern users of the `db` package which would receive this exception.

gentlecat: Description for `BadDataException` says "Should be used when incorrect data is being submitted." To me that seems like exactly the case you use this new exception for. So that addition seems unnecessary.

critiquebrainz/db/review_test.py
```diff
@@ -31,6 +31,7 @@ def test_review_creation(self):
             entity_id="e7aad618-fa86-3983-9e77-405e21796eca",
             entity_type="release_group",
             text="Testing",
+            rating=100,
```
gentlecat: You added rating everywhere, but is it necessary? Would it be better to test different kinds of reviews that we'll be able to have now (with rating, without rating, with both text and rating)?

critiquebrainz/db/revision.py
```diff
@@ -148,23 +152,34 @@ def get_revision_number(review_id, revision_id):
     return rev_num
 
 
-def create(review_id, text):
+def create(review_id, text, rating):
```
gentlecat: Did you forget to check that either `text` or `rating` is passed to this function?

critiquebrainz/db/revision.py
```diff
@@ -148,23 +152,34 @@ def get_revision_number(review_id, revision_id):
...
         })
 
+    review = db_review.get_by_id(review_id)
+    rev_num = get_revision_number(review["id"], review["last_revision"]["id"])
+    if rev_num > 1:
+        revisions = get(review["id"], limit=2, offset=0)
+        if revisions[0]["rating"] != revisions[1]["rating"]:
+            db_avg_rating.update(review["entity_id"], review["entity_type"])
+    elif rating is not None:
+        db_avg_rating.update(review["entity_id"], review["entity_type"])
```
gentlecat: I don't think I understand this part, can you clarify what it does? It also seems like that should go into a separate function.

critiquebrainz/db/avg_rating.py
```diff
@@ -0,0 +1,118 @@
...
+                              AND entity_type = :entity_type)
+              GROUP BY review_id)
+            
+            SELECT SUM(rating),
+                   COUNT(rating)
+              FROM revision
+        INNER JOIN LatestRevisions
+                ON revision.review_id = LatestRevisions.review_id
+               AND revision.timestamp = LatestRevisions.created_at
+        """), {
```
ps2611: No. Because revisions with no rating will not make any difference to the sum and count.

critiquebrainz/db/avg_rating_test.py
ps2611: But they do not have the same structure. So they can not be compared.

critiquebrainz/db/exceptions.py
ps2611: Removed it.

critiquebrainz/db/review_test.py
ps2611: I had it done when rating argument was mandatory. I will update it.

critiquebrainz/db/avg_rating_test.py
gentlecat: Ah, I misread it.

critiquebrainz/db/revision.py
ps2611: This checks if the rating of newly created revision and earlier revision (if it exists) are equal.
It does not call the avg_rating.update if they are equal.

critiquebrainz/db/revision.py
ps2611: Even if one of text or rating is None, both of them must be passed to insert their values. And, the check if text and rating both are None is placed at the start of update and create in review. So, it works fine.

gentlecat: What if I have a review without a rating? I shouldn't need to worry about how revisions are stored when I use this function. This function might be used in other places in the future, so it is better to do the check here as well.

Arguments can always be modified however necessary before they are inserted into the database.

gentlecat: Also, I'd suggest to make both `text` and `rating` optional here too.

critiquebrainz/db/revision.py
gentlecat: So it updates the average rating only if the rating in this review has changed?

critiquebrainz/db/review_test.py
gentlecat: OK. You could also add tests to make sure that all kinds of reviews are stored properly.

critiquebrainz/db/revision.py
ps2611: Exactly.

gentlecat: Would probably be better to put a comment saying this instead at the top of the whole block. It's a bit more clear.

critiquebrainz/db/review_test.py
ps2611: All possible types of a review have been covered now.

critiquebrainz/db/revision.py
ps2611: Did that and made the statement simpler.

critiquebrainz/db/avg_rating.py
```diff
@@ -0,0 +1,117 @@
...
+            "entity_id": entity_id,
+            "entity_type": entity_type,
+            "rating": avg_rating,
+            "count": count,
+        })
+
+
+def delete(entity_id, entity_type):
+    """Deletes the avg_rating, given entity_id and entity_type
```
gentlecat: I'd also note here when we might need to delete average rating.

critiquebrainz/db/avg_rating.py
```diff
@@ -0,0 +1,117 @@
...
+             WHERE entity_id = :entity_id
+               AND entity_type = :entity_type
+        """), {
+            "entity_id": entity_id,
+            "entity_type": entity_type,
+        })
+
+
+def get(entity_id, entity_type):
+    """Get average rating from entity_id
```
gentlecat: *Get average rating of an entity.*

gentlecat:
Feel free to submit another PR with adjustments to the things I noted. Not critical.

PR #128, opened 2017-07-12
Update current usage of the term 'rating'

ps2611:
- Currently 'rating' is used as difference in +/- votes to list and determine
  popular reviews
- Changed it to 'net_votes' to avoid conflict with entity rating

I didn't change it to 'popularity' because we may change the criteria to determine popular reviews after the rating system has been implemented

### DIFF
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -309,7 +309,7 @@ def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None,
         entity_id (uuid): ID of the entity that has been reviewed.
         entity_type (str): Type of the entity that has been reviewed.
         user_id (uuid): ID of the author.
-        sort (str): Order of the returned reviews. Can be either "net_votes" (order by difference in +/- votes),
+        sort (str): Order of the returned reviews. Can be either "popularity" (order by difference in +/- votes),
                     or "created" (order by creation time), or "random" (order randomly).
         limit (int): Maximum number of reviews to return.
         offset (int): Offset that can be used in conjunction with the limit.
@@ -374,9 +374,9 @@ def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None,
         count = result.fetchone()[0]
     order_by_clause = str()
 
-    if sort == "net_votes":
+    if sort == "popularity":
         order_by_clause = """
-            ORDER BY net_votes DESC
+            ORDER BY popularity DESC
         """
     elif sort == "created":
         order_by_clause = """
@@ -386,7 +386,7 @@ def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None,
         order_by_clause = """
             ORDER BY RANDOM()
         """
-    # Note that all revisions' votes are considered in these net_votes
+    # Note that all revisions' votes are considered in this popularity
     query = sqlalchemy.text("""
         SELECT review.id,
                review.entity_id,
@@ -415,7 +415,7 @@ def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None,
                SUM(
                    CASE WHEN vote = 't' THEN 1
                    WHEN vote = 'f' THEN -1 WHEN vote IS NULL THEN 0 END
-               ) AS net_votes,
+               ) AS popularity,
                latest_revision.id as latest_revision_id,
                latest_revision.timestamp as latest_revision_timestamp,
                latest_revision.text as text,
@@ -478,9 +478,9 @@ def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None,
 def get_popular(limit=None):
     """Get a list of popular reviews.
 
-    Popularity is determined by net_votes of a particular review. net_votes is a
+    Popularity is determined by 'popularity' of a particular review. popularity is a
     difference between positive votes and negative. In this case only votes
-    from the last month are used to calculate net_votes to make results more
+    from the last month are used to calculate popularity to make results more
     varied.
 
     Args:
@@ -512,7 +512,7 @@ def get_popular(limit=None):
                            CASE WHEN vote = 't' THEN 1
                                 WHEN vote = 'f' THEN -1
                                 WHEN vote IS NULL THEN 0 END
-                       ) AS net_votes,
+                       ) AS popularity,
                        latest_revision.id AS latest_revision_id,
                        latest_revision.timestamp AS latest_revision_timestamp,
                        latest_revision.text AS text,
@@ -546,7 +546,7 @@ def get_popular(limit=None):
                       ) AS randomized_entity_ids
                     )
               GROUP BY review.id, latest_revision.id
-              ORDER BY net_votes
+              ORDER BY popularity
                  LIMIT :limit
             """), {
                 "limit": defined_limit,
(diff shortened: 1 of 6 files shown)


gentlecat:
> I didn't change it to 'popularity' because we may change the criteria to determine popular reviews after the rating system has been implemented

Popularity of a review would have nothing to do with a rating system though. One is determined by readers and another by authors.

ps2611:
Agreed. I misunderstood it as popularity of entity rather than of the review.
So, I will change 'net_votes' to 'popularity'.

gentlecat:
I made some changes to the tests, so it would be nice if you could rebase on top of the latest commit on the *master* branch. I'm also still waiting for the changes I mentioned before:

> Make sure to replace the term `rating` in other places where it's used as well. We don't want to have multiple terms describing the same thing. This will cause confusion in the future.

Tell me if you need pointers for this and when this can be done so that we are on time.

critiquebrainz/ws/review/views.py
```diff
@@ -283,10 +283,10 @@ def review_list_handler():
                 "id": "CC BY-NC-SA 3.0",
                 "info_url": "https:\/\/creativecommons.org\/licenses\/by-nc-sa\/3.0\/"
               },
-              "rating": 0,
               "source": "BBC",
               "source_url": "http:\/\/www.bbc.co.uk\/music\/reviews\/vh54",
               "text": "REVIEW TEXT GOES HERE",
+              "rating": 100, (Rating part of the review)
```
gentlecat: This value is not returned, so it shouldn't be there.

gentlecat:
There are still some usages left. Here's a list:
screenshot 2017-07-14 20 39 28

critiquebrainz/ws/review/views.py
ps2611: This is what I asked about earlier. This should be done when making changes in the ws code for rating feature.

ps2611:
The changes in db/review and db/review_test are in https://github.com/metabrainz/critiquebrainz/pull/125.
They are a part of changes in tests.

gentlecat:
> The changes in db/review and db/review_test are in #125.They are a part of changes in tests.

Why can't they be a part of this one? I think they should be.

Think about it this way: a pull request should contain a change that can be merged and deployed on it's own. We can easily avoid confusion that having multiple terms for the same thing (or same term describing different things) can cause by changing it all here. The change is not going to be that big.

critiquebrainz/ws/review/views.py
gentlecat: Right, this is not related to renaming existing "rating" usage to "popularity".

ps2611:
Understood. Making the required changes now
And should we change it in db/users as well because it was used previously too and its pretty isolated?
If yes, then change it to what?

critiquebrainz/db/review.py
```diff
@@ -415,10 +415,11 @@ def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None,
                SUM(
                    CASE WHEN vote = 't' THEN 1
                    WHEN vote = 'f' THEN -1 WHEN vote IS NULL THEN 0 END
-               ) AS rating,
+               ) AS popularity,
                latest_revision.id as latest_revision_id,
                latest_revision.timestamp as latest_revision_timestamp,
                latest_revision.text as text,
+               latest_revision.rating as rating,
```
gentlecat: Also make sure to avoid making additions to SQL queries (it's fine to change names). This should be done separately when you have an actual implementation in works.

gentlecat:
> And should we change it in db/users as well because it was used previously too and its pretty isolated?

"Rating" there is used in a completely different context. It's a rating value for images coming from Gravatar. So there's no need to change that.

critiquebrainz/db/review.py
ps2611: I have to do it for tests otherwise they will fail.

gentlecat: Which tests?

ps2611: We have to do it now for tests otherwise they will fail.

critiquebrainz/ws/review/views.py
```diff
@@ -283,10 +283,10 @@ def review_list_handler():
                 "id": "CC BY-NC-SA 3.0",
                 "info_url": "https:\/\/creativecommons.org\/licenses\/by-nc-sa\/3.0\/"
               },
+              "rating": 0,
```
gentlecat: `rating` values shouldn't be listed there if they are no longer returned though. Just replace them with `popularity`.

critiquebrainz/db/review.py
ps2611: Here https://github.com/metabrainz/critiquebrainz/pull/125/files#diff-f16a779841395de9d9362b2168bf320f

gentlecat: Then the pull request #125 would be the place to make that change.

critiquebrainz/ws/review/views.py
ps2611: Take a look at this:https://github.com/metabrainz/critiquebrainz/pull/125/files#diff-c15d6e193e4bf52d4a77707695d5f205
Rating will be returned in the context of entity rating. So, will have to change that appropriately when we work on ws for rating feature

gentlecat: I understand that this will change in the future, but this has nothing to do with changes you are making in this pull request. Forget about pull request #125 for a moment and just focus on this one.

What is the work that is being done here? Changing usage of a term "rating" to "popularity". Is this directly related to the upcoming changes that you are working on to implement the rating functionality? Yes, but this is irrelevant in the context of this pull request. This is a separate part. Don't mix pull requests when they don't need to be, you'll only make it more confusing.

critiquebrainz/db/review.py
ps2611: The changes in #125 are dependent on this PR. So its getting a bit confusing. That's the reason I asked you if I could do it in that PR as a seperate commit.

critiquebrainz/ws/review/views.py
ps2611: Okay. I will push the final changes in a bit.

critiquebrainz/db/review.py
gentlecat: One way to keep another pull request in sync with another unmerged branch is to keep rebasing the new work on top of it. Commits from another PR will show up there, but as long as it's made clear that the work in that PR depends on another it shouldn't be a problem.

In general I'd suggest to avoid a situation like that if it's confusing to keep dependent PRs in sync.

critiquebrainz/db/review.py
```diff
@@ -510,10 +510,10 @@ def get_popular(limit=None):
                            CASE WHEN vote = 't' THEN 1
                                 WHEN vote = 'f' THEN -1
                                 WHEN vote IS NULL THEN 0 END
-                       ) AS rating,
+                       ) AS popularity,
                        latest_revision.id AS latest_revision_id,
                        latest_revision.timestamp AS latest_revision_timestamp,
-                       latest_revision.text AS text
+                       latest_revision.text AS text,
```
gentlecat: Pretty sure this change is causing a syntax error.

gentlecat:
> I finally got this. We must change the current use of 'rating' before introducing the entity rating.

Exactly!

critiquebrainz/db/review.py
ps2611: Fixed it.

PR #129, opened 2017-07-13
Add support for running PyLint checks in Jenkins + some fixes

gentlecat:
This adds support for running PyLint checks in Jenkins and fixes some issues.


gentlecat:
The idea is to always keep PyLint warning count at 0. We should also start fixing ignored issues.

gentlecat:
Fixed some. Going to merge.

PR #130, opened 2017-07-14
Stop ignoring too-many-branches warning in pylint

gentlecat:


critiquebrainz/db/review.py
```diff
@@ -297,6 +297,7 @@ def create(*, entity_id, entity_type, user_id, is_draft, text,
     return get_by_id(review_id)
 
 
+# pylint: disable=too-many-branches
```
gentlecat: No good way to avoid it in this function without, perhaps, creating a helper function to generate SQL statement and argument extensions based on a set of arguments.

PR #132, opened 2017-07-18
Remove 'popularity' where its not used in correct form

ps2611:
In list_review and get_popular functions in db/review, all revisions' votes are considered in popularity which was not the case where changes are made.

### DIFF
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -59,7 +59,6 @@ def get_by_id(review_id):
             "source_url": str,
             "last_revision: dict,
             "votes": dict,
-            "popularity": int,
             "text": str,
             "created": datetime,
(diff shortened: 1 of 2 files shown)


gentlecat:
I don't really understand what this does. Can you please clarify? It looks like this change just removes that popularity output, which doesn't seem right.

ps2611:
The review.get_by_id function returns popularity as difference in +/- votes of only the latest revision. But,
at all other places, all revisions' votes are considered for popularity. So, its better to remove it.

gentlecat:
Well, I think using only votes for the latest revision is the right way for now. It might not be the best way, but that's what we have. If other places do it differently then it should be fixed not removed.

Feel free to reopen this if I'm completely misunderstanding it.

PR #133, opened 2017-07-18
Consolidate votes in a dict in db.review.get_by_id() function

ps2611:
The function is supposed to return votes of a review in votes dictionary as mentioned in the comments.
So, created that and other necessary changes elsewhere.

### DIFF
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -135,9 +135,11 @@ def get_by_id(review_id):
             "full_name": review["full_name"],
         }
         votes = db_revision.votes(review["last_revision"]["id"])
-        review["votes_positive_count"] = votes["positive"]
-        review["votes_negative_count"] = votes["negative"]
-        review["popularity"] = review["votes_positive_count"] - review["votes_negative_count"]
+        review["votes"] = {
+            "positive": votes["positive"],
+            "negative": votes["negative"],
+        }
+        review["popularity"] = review["votes"]["positive"] - review["votes"]["negative"]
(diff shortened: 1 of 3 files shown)


critiquebrainz/db/review.py
```diff
@@ -135,9 +135,11 @@ def get_by_id(review_id):
             "full_name": review["full_name"],
         }
         votes = db_revision.votes(review["last_revision"]["id"])
-        review["votes_positive_count"] = votes["positive"]
-        review["votes_negative_count"] = votes["negative"]
-        review["popularity"] = review["votes_positive_count"] - review["votes_negative_count"]
+        review["votes"] = {
+            "positive": votes["positive"],
+            "negative": votes["negative"],
+        }
```
gentlecat: I like the fact that vote values are in a sub-dict now. That seems cleaner.

critiquebrainz/db/review.py
```diff
@@ -135,9 +135,11 @@ def get_by_id(review_id):
...
+        review["popularity"] = review["votes"]["positive"] - review["votes"]["negative"]
```
gentlecat: Is it necessary to actually return this value if it's so easy to calculate it from what is already being returned? To me it seems like this actually can be removed.

ps2611: It's removed in #132

gentlecat: I closed #132 (see my comment there about why). This was done in a different context. I would agree to remove it here to avoid data duplication.

PR #135, opened 2017-07-20
CB-231: Reduce db queries for entity: place

ferbncode:
Reduces the number of db queries for entity: `place`. Using `joinedload` for `area` and `type` saves two queries. Also, added `get_entities_by_gid` which reduces the number of queries to fetch many places by `gid`.(to max 3 queries (in case of redirected entities, otherwise 1)). Also, fixes fetching entities for redirected mbids. A single `fetch_multiple_places` method now makes atmost 7 queries (4 for relationships + 3 for fetching all places(at max)) (for any number of `entity-gids`). 
I will also later update the tests for checking on the number of queries.

### DIFF
--- a/critiquebrainz/frontend/external/musicbrainz_db/utils.py
+++ b/critiquebrainz/frontend/external/musicbrainz_db/utils.py
@@ -13,5 +13,5 @@ def get_entities_by_gids(query, entity_model, redirect_model, mbids):
         entities.extend(redirected_entities)
     remaining_gids = list(set(mbids) - {entity.gid for entity in entities})
     if remaining_gids:
-        raise mb_exceptions.NoDataFoundException("Couldn't find a place with id(s): {mbids}".format(mbids=remaining_gids))
+        raise mb_exceptions.NoDataFoundException("Couldn't find an entity with id(s): {mbids}".format(mbids=remaining_gids))
     return entities
--- a/critiquebrainz/frontend/external/musicbrainz_db/utils.py
+++ b/critiquebrainz/frontend/external/musicbrainz_db/utils.py
@@ -2,6 +2,17 @@ import critiquebrainz.frontend.external.musicbrainz_db.exceptions as mb_exceptio
 
 
 def get_entities_by_gids(query, entity_model, redirect_model, mbids):
+    """Get entities using their mbids.
+
+    Args:
+        query (Query): sqlalchemy Query object.
+        entity_model (mbdata.models): Model of the target entity.
+        redirect_model (mbdata.models): Redirect Model of the target entity.
+        mbids (list): IDs of the target entities.
+
+    Returns:
+        List of objects of target entities.
+    """
     entities = query.filter(entity_model.gid.in_(mbids)).all()
     remaining_gids = list(set(mbids) - {entity.gid for entity in entities})
     if remaining_gids:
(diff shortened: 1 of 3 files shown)


gentlecat:
> I will also later update the tests for checking on the number of queries.

What do you mean by that?

critiquebrainz/frontend/external/musicbrainz_db/utils.py
```diff
@@ -0,0 +1,17 @@
+import critiquebrainz.frontend.external.musicbrainz_db.exceptions as mb_exceptions
+
+
+def get_entities_by_gids(query, entity_model, redirect_model, mbids):
```
gentlecat: Can you please add a docstring describing what this function does and what the arguments are (can use type annotations to indicate types as well)?

critiquebrainz/frontend/external/musicbrainz_db/utils.py
```diff
@@ -0,0 +1,17 @@
...
+    if remaining_gids:
+        redirects = query.session.query(redirect_model).filter(redirect_model.gid.in_(remaining_gids))
+        redirect_gids = {redirect.redirect_id: redirect.gid for redirect in redirects}
+        redirected_entities = query.filter(redirect_model.redirect.property.primaryjoin.left.in_(redirect_gids.keys())).all()
+        for entity in redirected_entities:
+            entity.gid = redirect_gids[entity.id]
+        entities.extend(redirected_entities)
+    remaining_gids = list(set(mbids) - {entity.gid for entity in entities})
+    if remaining_gids:
+        raise mb_exceptions.NoDataFoundException("Couldn't find an entity with id(s): {mbids}".format(mbids=remaining_gids))
```
gentlecat: Can simply say "*Couldn't find entities with IDs: \<list of IDs\>*" here.

ferbncode:
I was thinking of adding code for checking on the number of sql queries issued (something like https://stackoverflow.com/questions/19073099/how-to-count-sqlalchemy-queries-in-unit-tests). But, on second thought, I think it would be good to keep that code only temporarily while writing code for fetching entities. Please suggest. Thanks :)

gentlecat:
I don't think a test for the number of queries would show anything useful. Benchmarking function runtime might be a bit more useful, but that's not something I'd worry about at the moment since we don't have an automated way to monitor this.

critiquebrainz/frontend/external/musicbrainz_db/utils.py
```diff
@@ -0,0 +1,28 @@
+import critiquebrainz.frontend.external.musicbrainz_db.exceptions as mb_exceptions
+
+
+def get_entities_by_gids(query, entity_model, redirect_model, mbids):
+    """Get entities using their mbids.
+
+    Args:
+        query (Query): sqlalchemy Query object.
```
ferbncode: The `query` may be modified for using statements like `joinedload` to make use of `join` statements and reducing the number of queries to the database.
Presently, due to SQLAlchemy's lazy loading, it makes an additional query for place type (https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/frontend/external/musicbrainz_db/serialize.py#L78) and another query for place area. (https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/frontend/external/musicbrainz_db/serialize.py#L87)

ferbncode:
Apologies for the docstring. 
On second thought, users of `get_entities_by_gids` should not really care about the models for the mbid redirects and the entity model as it makes the usage a bit complex :sweat_smile:. I've updated the function to take only the entity_type and made use of keyword only arguments.

gentlecat:
*Side note:*

> Apologies for the docstring.

There's no need to apologise. Not seeing some issues, making mistakes, doing adjustments, etc. is the reason we do code reviews. There are plenty of things written about this out there, https://softwareengineering.stackexchange.com/a/213764 is the first one I found. You might know about this already, but I feel like this is important to point out since we had instances where reviews were taken personally or their purpose has been misunderstood in some other way.

PR #138, opened 2017-07-27
CB-244: Frontend for the rating system

ps2611:
Basic functionality for different types of review is up and running. Styling along with some other minor changes are to be done.


critiquebrainz/frontend/forms/review.py
```diff
@@ -16,7 +16,15 @@ def __call__(self, form, field):
...
-        if l < self.min or self.max != -1 and l > self.max:
+        if (l < self.min or self.max != -1 and l > self.max) and (l != 0):
+            raise ValidationError(self.message)
+
+class TextAndRatingBothNotEmpty(object):
+    def __init__(self, message):
+        self.message = message
+
+    def __call__(self, form, field):
+        if not field.data and not form.rating.data:
```
gentlecat: Accessing other fields in a validator doesn't seem right. `Form` class uses a `validate` method which you can override and check that either text or rating are present. Make sure to call original method with `super().validate()`.

critiquebrainz/frontend/templates/review/entity/base.html
```diff
@@ -124,3 +125,7 @@
   </div>
 </div>
 {% endblock %}
+
+{% block scripts %}
+  <script src="{{ get_static_path('scripts/rating1.js') }}" type="text/javascript"></script>
```
gentlecat: Is it supposed to be *rating.js*?

critiquebrainz/frontend/templates/review/entity/base.html
```diff
@@ -124,3 +125,7 @@
   </div>
 </div>
 {% endblock %}
+
+{% block scripts %}
```
gentlecat: Make sure to call `super()` when defining a `scripts` block.

critiquebrainz/frontend/templates/review/modify/base.html
```diff
@@ -108,4 +114,5 @@
       });
     });
   </script>
+  <script src="{{ get_static_path('scripts/rating.js') }}" type="text/javascript"></script>
```
gentlecat: JS files shouldn't be accessed directly. The way we do it here now is by building JS files in *gulpfile.js* and referencing them like this: `<script src="{{ get_static_path('rating.js') }}"></script>`. That way you can easily use React, jQuery, etc. in your scripts.

critiquebrainz/frontend/views/review.py
```diff
@@ -211,8 +214,10 @@ def create():
             return redirect(url_for('user.reviews', user_id=current_user.id))
 
         is_draft = form.state.data == 'draft'
+        if not form.text.data:
```
gentlecat: What values would these be?

critiquebrainz/frontend/forms/review.py
```diff
@@ -16,7 +16,15 @@ def __call__(self, form, field):
         if form.state.data == "draft":
             return
         l = len(field.data) if field.data else 0
-        if l < self.min or self.max != -1 and l > self.max:
+        if (l < self.min or self.max != -1 and l > self.max) and (l != 0):
```
gentlecat: Can you explain what this change does? I even forgot how the original check works.

ps2611: Originally it checked whether the length is between min and max and max is not -1. I added the check for not zero so that error is not raised when its empty. I am gonna remove it because the in- built Optional() validator works better.

critiquebrainz/frontend/forms/review.py
ps2611: I did that - https://gist.github.com/ps2611/119b0d6e073ba680034f5a78789afe34
But somehow, it doesn't get added to form.errors and gets displayed like - https://gist.github.com/ps2611/1e3443a370865b6dfa1bfede07c950bc

critiquebrainz/frontend/views/review.py
ps2611: TextAreaField in wtforms returns an empty string when no input is provided. So, it must be changed to None before inserted into the database.

gentlecat: Can you make a comment noting this? Or, even better, just compare to an empty string.

gentlecat: Same in another case below.

critiquebrainz/frontend/templates/review/entity/base.html
ps2611: Yes, a typo.

critiquebrainz/frontend/views/review.py
ps2611: Done

critiquebrainz/frontend/templates/review/modify/base.html
ps2611: Added ratingBundle in gulpfile.js

critiquebrainz/frontend/forms/review.py
```diff
@@ -43,12 +43,22 @@ class ReviewEditForm(Form):
...
     def __init__(self, default_license_id='CC BY-SA 3.0', default_language='en', **kwargs):
         kwargs.setdefault('license_choice', default_license_id)
         kwargs.setdefault('language', default_language)
         Form.__init__(self, **kwargs)
 
+    def validate(self):
+        if not super(ReviewEditForm, self).validate():
+            return False
+        if not self.text.data and not self.rating.data:
+            message = "Please provide at least one of text or rating for this review."
```
gentlecat: Does the message have to be in a separate variable?

critiquebrainz/frontend/static/scripts/rating.js
```diff
@@ -0,0 +1,157 @@
+(function ($) {
```
gentlecat: Can you explain why this is necessary here? *Sorry, I'm not that good at JS.*

critiquebrainz/frontend/static/scripts/rating.js
```diff
@@ -0,0 +1,157 @@
...
+  'use strict';
+
+  var clearClass = 'rating-clear',
```
gentlecat: I think it's just [`let`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) by default now unless `var` is actually needed.

critiquebrainz/frontend/static/scripts/rating.js
```diff
@@ -0,0 +1,157 @@
...
+  var Rating = function(input, options) {
+    var $input = this.$input = input;
+    this.options = parseOptions($input, options);
+    var $ratingEl = this.$el = createRatingEl($input, this.options);
+    $input.addClass(hiddenClass).before($ratingEl);
+    $input.attr('type', 'hidden');
+    this.highlight($input.val());
+  };
+
+  Rating.VERSION = '0.4.0';
```
gentlecat: What's this version?

critiquebrainz/frontend/views/review.py
```diff
@@ -94,11 +94,16 @@ def entity(id, rev=None):
...
-    review["text_html"] = markdown(revision['text'], safe_mode="escape")
+    if revision["text"] is None:
+        review["text_html"] = None
+    else:
+        review["text_html"] = markdown(revision['text'], safe_mode="escape")
 
     user_all_reviews, review_count = db_review.list_reviews(user_id=review["user_id"], sort="random", exclude=[review["id"]])  # pylint: disable=unused-variable
     other_reviews = user_all_reviews[:3]
-    return render_template('review/entity/%s.html' % review["entity_type"], review=review, spotify_mappings=spotify_mappings, soundcloud_url=soundcloud_url, vote=vote, other_reviews=other_reviews)
+    rating_titles = {20: "Terrible", 40: "Bad", 60: "Average", 80: "Good", 100: "Extradordinary"}
```
gentlecat: These strings should be wrapped in `gettext`.

critiquebrainz/frontend/static/scripts/rating.js
```diff
@@ -0,0 +1,157 @@
...
+
+  Rating.DEFAULTS = DEFAULTS;
+
+  Rating.prototype = {
```
gentlecat: Would it make sense to use [Classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) here?

critiquebrainz/frontend/templates/review/entity/base.html
```diff
@@ -25,9 +25,17 @@
           {% endif %}
         </em>
       {% endif %}
+      <br>
+      <br>
```
gentlecat: It's better to use CSS instead of adding line breaks for spacing.

critiquebrainz/frontend/static/scripts/rating.js
ps2611: It provides closure.

critiquebrainz/frontend/static/scripts/rating.js
ps2611: Its plugin version. I will remove it.

critiquebrainz/frontend/static/scripts/rating.js
ps2611: Since there's a closure, so the global environment won't be polluted and the code has been taken from a plugin, I think its wise to let it be the way author wrote it.

gentlecat: Don't be afraid to modify a library code if it can be improved or adjusted for the way we use it. Not every library published on the internet is perfect. :)

critiquebrainz/frontend/forms/review.py
mwiencek: The message isn't entirely grammatical. I suggest "Please provide some text or a rating for this review."

critiquebrainz/frontend/static/scripts/rating.js
mwiencek: I don't recommend forking external libraries, but I can't say much else without knowing to what extent you've modified this. Can you please link to the original source code?

If we fork this, the file *must* include a copyright and license attached.

ps2611: Original source - https://github.com/javiertoledo/bootstrap-rating-input
This plugin suited our needs the best with a few modifications. Can you please tell why are you skeptical of this kinda thing. And yes I will add the copyright.

mwiencek: It's more code we have to test (the modifications we make) and maintain (we don't have a lot of resources).

If any bug fixes are made upstream, we have to manually merge them into our fork, instead of just bumping the version in package.json.

It doesn't look like you've made that many modifications: https://gist.github.com/mwiencek/838e5abd016ce6627c23ba47befe2d2d

You don't need to fork the code to modify `DEFAULTS`. You can mutate it externally through `$.fn.rating.Constructor.DEFAULTS`.

For the rest, why not just submit a patch upstream to support what you're doing? That way other people benefit, and we don't have to maintain our own fork.

mwiencek: That's just my opinion, but I don't have much stake in this. :) Fork the code if you think that's a good idea.

gentlecat: If it's possible to just make it a dependency and modify defaults externally, I'd also prefer to do that.

ps2611: I am trying do that and also looking for some other plugin which might be more suitable.

ps2611:
Understood. Will make proper commits from now on.

critiquebrainz/frontend/static/scripts/rating.js
gentlecat: There's less than a week left for the official GSoC coding period, so I'd suggest to avoid doing very time consuming tasks such as evaluating different libraries. If this one works well enough then the only thing we need to do is clean up the code to meet our requirements and move on.

ps2611: Yes, I decided for this plugin after a lot of searching and still didn't find any other which is more suitable. I have updated the package file and working on the required changes. Will let you know at the earliest if I face any problems.

critiquebrainz/frontend/static/scripts/rating.js
```diff
@@ -1,155 +1,17 @@
...
+
+$(document).ready(function() {
+  // Set color and size of stars icon
+  $('.glyphicon-star, .glyphicon-star-empty').css({'color': '#DAA520', 'font-size': '18px'});
+  // Set color and size of delete icon
+  $('.glyphicon-remove-circle').css({'color': '#808080', 'font-size': '16px'});
+  $('.glyphicon-remove-circle').hover(function() {
+    $(this).css('color', 'red');
+    }, function() {
+    $(this).css('color', '#808080');
```
gentlecat: Why not do this in the style definitions (Less files) that we already have?

ps2611: Should this go in main.less or theme/glyphicons.less?

gentlecat: Stuff in *theme* shouldn't really be modified, so I'd say *main.less*.

ps2611:
Please don't mind the merge and conflicts commits. Ran into some problems.

gentlecat:
> The functions which have this warning are quite large.

Well, that's not good. One way to fix this would be to extract some of the operations that are done in them into separate functions.

gentlecat:
You can also add an ignore for this check on the functions and add a `TODO` comment saying that it needs improvement.

gentlecat:
Another one:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/werkzeug/wsgi.py", line 660, in __call__
    return app(environ, start_response)
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python3.6/site-packages/flask_login/utils.py", line 228, in decorated_view
    return func(*args, **kwargs)
  File "/code/critiquebrainz/frontend/views/review.py", line 304, in edit
    form.rating.data = review["rating"] // 20
TypeError: unsupported operand type(s) for //: 'NoneType' and 'int'
```

gentlecat:
I can create a review with 0 stars. Is it possible to do that in the MusicBrainz UI?

ps2611:
`I get this exception when I open a homepage when there's a review without text:`
Are you sure you have included changes from #147 and the cache is clear? Because it works perfectly fine for me.

ps2611:
> Another one:

Resolved this

gentlecat:
> Can you please check if you have included changes from #147 and the cache is clear? Because it works perfectly fine for me.

OK, seems to be working with latest changes now. But it might still be worth checking if there's any text before trying to convert it.

critiquebrainz/frontend/templates/event/entity.html
```diff
@@ -102,6 +102,13 @@ <h4 style="margin-bottom:0;">{{ _('Reviews') }}</h4>
   <div class="col-md-3">
     <h4>{{ _('Event information') }}</h4>
 
+    {% if avg_rating %}
+      <p>
+        <i class="glyphicon glyphicon-star"></i>
+        <span style="font-size: 16px;">{{ avg_rating.rating }}</span><em class="text-muted">/5 based on {{ avg_rating.count }} ratings</em>
+      </p>
+    {% endif %}
```
gentlecat: It looks like this part is the same among all templates. It would be a good idea to make it a [macro](http://jinja.pocoo.org/docs/2.9/templates/#macros).

critiquebrainz/frontend/views/place.py
```diff
@@ -35,6 +37,11 @@ def entity(id):
         limit=limit,
         offset=offset,
     )
+    try:
+        avg_rating = db_avg_rating.get(place['id'], "place")
+        avg_rating["rating"] = round(avg_rating["rating"] / 20, 1)
+    except db_exceptions.NoDataFoundException:
+        avg_rating = None
```
gentlecat: This part of code is also almost the same among different views. Can it be moved into one function that does average rating retrieval?

critiquebrainz/frontend/views/review.py
```diff
@@ -58,6 +59,8 @@ def browse():
                            page=page, limit=limit, count=count, entity_type=entity_type)
 
 
+# TODO : Refactor this function to remove PyLint warning.
```
gentlecat: In CB code we put an identifier next to a TODO comment to know who to ask if more details are needed. So something like
```python
# TODO(psolanki): Refactor this function to remove PyLint warning.
```

gentlecat: Same in the comment below.

critiquebrainz/frontend/views/review.py
```diff
@@ -94,17 +97,27 @@ def entity(id, rev=None):
...
+        review["text_html"] = markdown(revision['text'], safe_mode="escape")
 
     user_all_reviews, review_count = db_review.list_reviews(  # pylint: disable=unused-variable
         user_id=review["user_id"],
         sort="random",
         exclude=[review["id"]],
     )
     other_reviews = user_all_reviews[:3]
+    if review["rating"] is not None:
+        review["rating"] //= 20
```
gentlecat: Can this conversion be done within the `db` package? That way users of these functions don't have to worry about converting the values, they'd just use 1-5 scale.

critiquebrainz/frontend/templates/review/entity/base.html
```diff
@@ -15,19 +19,33 @@
...
 
-      <p style="word-wrap: break-word; white-space: pre-wrap;">{{ review.text_html|safe }}</p>
-
+      {% if review.rating %}
+        <div style="margin-top: 15px;">
+          <input type="number" class="rating" id="rating" value={{ review.rating }} data-readonly data-inline />
+        </div>
+      {% endif %}
+      {% if review.text %}
+        <!-- Used to display only-text review properly -->
```
gentlecat: What is used? Not sure if I understand what issue you are trying to solve there.

gentlecat:
> Please don't mind the merge and conflicts commits. Ran into some trouble.

You can use `git rebase -i origin/master` on your branch to sync the changes quicker and cleaner. Let us know if you have any questions about that.

critiquebrainz/frontend/templates/review/entity/base.html
ps2611: It is used for proper spacing above text when stars icon are not present.

critiquebrainz/frontend/views/place.py
ps2611: Added views/avg_rating for this purpose.

critiquebrainz/frontend/views/review.py
ps2611: Will keep this in mind next time.

critiquebrainz/frontend/templates/place/entity.html
```diff
@@ -93,10 +94,8 @@ <h4 style="margin-bottom:0;">{{ _('Reviews') }}</h4>
   <div class="col-md-3">
     <h4>{{ _('Place information') }}</h4>
       {% if avg_rating %}
```
gentlecat: This checking can also be moved inside the `show_avg_rating` macro. And simply pass `avg_rating` to it. Make usage as simple as possible.

critiquebrainz/frontend/views/avg_rating.py
```diff
@@ -0,0 +1,12 @@
+import critiquebrainz.db.avg_rating as db_avg_rating
+import critiquebrainz.db.exceptions as db_exceptions
+
+def get(entity_id, entity_type):
```
gentlecat: It would probably be better to just put this into `__init__.py` of the `frontend.views` package instead of creating a new module just for one utility function. I think it's expected that modules in `views` contain actual views.

critiquebrainz/frontend/views/review.py
gentlecat: I'm still interested in a fix for this. Conversion from 5-star to 100 should be done within the `db` package. Modules that use it don't need to know about how ratings are stored. It's about making the interface less complex and easier to use.

critiquebrainz/frontend/templates/review/modify/base.html
```diff
@@ -13,19 +13,25 @@
...
-              url='http://daringfireball.net/projects/markdown/syntax') }}
-      </small>
-    </ul>
-  </div>
+  <hr />
+
   <form id="review-editor" method="POST" class="form-horizontal" role="form">
     {{ form.hidden_tag() }}
     <div class="form-group">
+      <label class="col-sm-2" for="rating">{{ _('Rating for the review:') }}</label>
```
gentlecat: `for the review` part can be removed since it's pretty obvious what the rating is for.

critiquebrainz/frontend/templates/review/entity/base.html
gentlecat: OK, but this should be simplified in the future.

critiquebrainz/frontend/templates/place/entity.html
```diff
@@ -92,6 +93,11 @@ <h4 style="margin-bottom:0;">{{ _('Reviews') }}</h4>
 
   <div class="col-md-3">
     <h4>{{ _('Place information') }}</h4>
+      {% if avg_rating %}
+        <p>
```
gentlecat: What is the `<p>` for here?

critiquebrainz/frontend/templates/macros.html
```diff
@@ -121,3 +121,8 @@
       {{ spotify_player(spotify_release_group_id, spotify_mappings, spotify_show_add_message) }}
   {% endif %}
 {% endmacro %}
+
+{% macro show_avg_rating(rating, count) %}
+  <i class="glyphicon glyphicon-star"></i>
+  <span style="font-size: 16px;">{{ rating }}</span><em class="text-muted">/5 {{ _('based on') }} {{ count }} {{ _('ratings') }}</em>
```
gentlecat: Split long lines, if necessary:
```html
<span style="font-size: 16px;">{{ rating }}</span>
<em class="text-muted">/5 {{ _('based on') }} {{ count }} {{ _('ratings') }}</em>
```

gentlecat: `{{ _('based on') }} {{ count }} {{ _('ratings') }}` should be combined into one localisable string. It can be a bad idea that the order of words would be the same in all languages.

```python
_('based on %(number)s ratings', number=count)
```

critiquebrainz/frontend/views/review.py
ps2611: I have proposed these changes - https://github.com/javiertoledo/bootstrap-rating-input/pull/61
If this gets merged then there won't be any need for converting ratings to different scales anywhere.

gentlecat: That's great, but that's what we have now and the interface is not very well designed.

critiquebrainz/frontend/views/avg_rating.py
ps2611: Move the function to __init__.py

critiquebrainz/frontend/templates/review/modify/base.html
ps2611: Removed it and changed the column grid accordingly.

critiquebrainz/frontend/templates/macros.html
ps2611: Its intentional. Splitting them will create an extra space.

ps2611: Combined into one.

critiquebrainz/frontend/templates/place/entity.html
ps2611: It is needed to maintain the spacing format.

ps2611:
There are 4 new commits containing the requested changes and 4 commits at the base are part of db/rating_scale.

critiquebrainz/frontend/templates/place/entity.html
gentlecat: You should use CSS for this.

critiquebrainz/frontend/templates/review/entity/base.html
```diff
@@ -88,9 +109,13 @@
...
           <div class="row">
-            <div class="col-md-5">
-              {{ _('<a href="%(link)s">Sign in</a> to rate this review', link=url_for('login.index', next=url_for('review.entity', id=review.id))) }}
-            </div>
+            {% if review.text %}
+              <div class="col-md-5">
+               {{ _('<a href="%(link)s">Sign in</a> to rate this review', link=url_for('login.index', next=url_for('review.entity', id=review.id))) }}
+              </div>
+            {% else %}
+              <div class="col-md-5"></div>
```
gentlecat: Why keep this duplicate block definition around? It might be better to just move the condition inside it.

critiquebrainz/frontend/views/__init__.py
```diff
@@ -0,0 +1,11 @@
+import critiquebrainz.db.avg_rating as db_avg_rating
+import critiquebrainz.db.exceptions as db_exceptions
+
+def get_avg_rating(entity_id, entity_type):
+    """Retrieve avg_rating and convert rating on a scale 1-5."""
```
gentlecat: What's `avg_rating`? Average rating? There's no need to literally use variable names in docstrings.

critiquebrainz/frontend/views/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+    try:
+        avg_rating = db_avg_rating.get(entity_id, entity_type)
+    except db_exceptions.NoDataFoundException:
+        avg_rating = None
+
+    return avg_rating
```
gentlecat: Instead of assigning a variable why not just directly return values?

critiquebrainz/frontend/templates/place/entity.html
ps2611: IMO, CSS isn't necessary here. Other elements in this column (in event and release_group) also use <p>, so we get perfectly formatted column without any use of CSS and by using mere <p>.

gentlecat: There are plenty of places that are not implemented correctly. The goal is to make the code better, not to keep using bad practices. When other code has been written it wasn't reviewed properly.

PR #145, opened 2017-08-09
Fix ISE on entity pages involving MBID redirects

ferbncode:
Due to sqlalchemy's `autoflushing`, it issues an `update` query while modifying redirected entity's `gid` field and this issue is raised https://sentry.metabrainz.org/share/issue/342e32363535/. I've updated the function `get_entities_by_gids` for fixing this. Eg. https://critiquebrainz.org/release-group/820ba931-840d-45f1-97ed-0cceaff26da0

### DIFF
--- a/critiquebrainz/frontend/external/musicbrainz_db/utils.py
+++ b/critiquebrainz/frontend/external/musicbrainz_db/utils.py
@@ -1,4 +1,5 @@
 import critiquebrainz.frontend.external.musicbrainz_db.exceptions as mb_exceptions
+from critiquebrainz.frontend.external.musicbrainz_db import mb_session
 from mbdata import models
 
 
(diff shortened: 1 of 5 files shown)


critiquebrainz/frontend/external/musicbrainz_db/utils.py
```diff
@@ -47,8 +48,11 @@ def get_entities_by_gids(*, query, entity_type, mbids):
         redirects = query.session.query(redirect_model).filter(redirect_model.gid.in_(remaining_gids))
         redirect_gids = {redirect.redirect_id: redirect.gid for redirect in redirects}
         redirected_entities = query.filter(redirect_model.redirect.property.primaryjoin.left.in_(redirect_gids.keys())).all()
-        for entity in redirected_entities:
-            entity.gid = redirect_gids[entity.id]
```
alastair: Because these are ORM objects, I wonder if there is a better way of doing this operation so that we don't modify them. What about returning a dictionary of `{gid: entity}` where the `gid` could be a redirect gid, and not just `entity.gid`. This is quite similar to what you're doing in `fetch_multiple_x` anyway, and means that we're not passing around modified but not committed ORM objects.

ferbncode:
This now also solves [CB-130](https://tickets.metabrainz.org/projects/CB/issues/CB-130) (for frontend). 

- For eg. `/release-group/820ba931-840d-45f1-97ed-0cceaff26da0` will show reviews for `/release-group/bf63ed7c-c15f-4e5d-8711-c495233b7cff` (Original MBID).
- Also, adding a review for the release-group at `/release-group/820ba931-840d-45f1-97ed-0cceaff26da0` will add a review with `entity_id`: `bf63ed7c-c15f-4e5d-8711-c495233b7cff`.

So, pages `/release-group/820ba931-840d-45f1-97ed-0cceaff26da0` (additional MBID) and `/release-group/bf63ed7c-c15f-4e5d-8711-c495233b7cff` (original MBID) will be completely same. This was the behavior intended for solving CB-130. right?. Please suggest.

critiquebrainz/frontend/external/musicbrainz_db/utils.py
```diff
@@ -37,20 +37,19 @@ def get_entities_by_gids(*, query, entity_type, mbids):
         mbids (list): IDs of the target entities.
 
     Returns:
-        List of objects of target entities.
+        Dictionary of objects of target entities keyed by their mbid.
```
gentlecat: MBID

PR #146, opened 2017-08-09
CB-231: Add wrapper for fetch_multiple_places for a single MBID

ferbncode:
Added wrapper `_get_place_by_gid` for calling `fetch_multiple_places` for a single `MBID`.

### DIFF
--- a/critiquebrainz/frontend/external/musicbrainz_db/place.py
+++ b/critiquebrainz/frontend/external/musicbrainz_db/place.py
@@ -21,12 +21,16 @@ def get_place_by_id(mbid):
     key = cache.gen_key(mbid)
     place = cache.get(key)
     if not place:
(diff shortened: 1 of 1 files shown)


alastair:
This looks OK to me, however if there are no pages which use the multiple version, do we need the added complexity that it brings? (I seem to recall that I asked this previously, so I guess we decided to keep it)

ferbncode:
The `/review` [pages](https://critiquebrainz.org/review) require fetching information for multiple entities. Since, these pages make a number of requests to the web-service presently, the `fetch_multiple_*` functions are added to fetch multiple entities in less time.

alastair:
Right, I had a memory that this was the case, no problem

gentlecat:
One thing: There's no need to mention `CB-231` since it's not directly related to this change.

PR #147, opened 2017-08-11
Don't include reviews without text into the popular list

ps2611:
Modified the criteria for popular_reviews.
Consider only those reviews with text for popular_reviews. lt doesn't seem right to display reviews with only rating on the index page.

### DIFF
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -559,6 +559,7 @@ def get_popular(limit=None):
                       ORDER BY RANDOM()
                       ) AS randomized_entity_ids
                     )
+                   AND latest_revision.text IS NOT NULL
               GROUP BY review.id, latest_revision.id
               ORDER BY popularity
                  LIMIT :limit


ps2611:
I am not quite sure how to change this part of code. Need some help here. https://github.com/ps2611/critiquebrainz/blob/3a8f407426f10cee173de47ee34d700c632b497c/critiquebrainz/db/review.py#L507

gentlecat:
> I am not quite sure how to change this part of code. Need some help here. https://github.com/ps2611/critiquebrainz/blob/3a8f407426f10cee173de47ee34d700c632b497c/critiquebrainz/db/review.py#L507

Can't you check if latest revision of each review has any text?

ps2611:
I don't know how the cache works but its possible that reviews might not contain any review with text or  the number falls short of the limit. So, we need a loop but again I don't know how cache.get() works.
Also, I don't understand how the popularity factor is accounted here.

gentlecat:
Oh, I misunderstood. You linked the line with the code related to cache key generation, not the whole function. I'm not sure what the problem is then. Cache only stores data that you give it.

The docstring of the `get_popular` function provides an overview for how it works. I think the only change that needs to be done with regards to this is to include only reviews that contain text. Now that I'm thinking about it, it doesn't actually make sense to  vote on usefulness of star-only review.

ps2611:
Understood. So, this PR is complete unless I am missing something.
And, yes, voting on rating-only review doesn't make any sense. Working on it now.

gentlecat:
I strongly suggest that you finish UI work that you started and get to the API if you want to finish the project in time. There are always be things to improve, but you barely have enough time to finish what was planned.

PR #151, opened 2017-08-20
Reference MetaBrainz-wide contribution guidelines

gentlecat:
Since we now have a repository with common contribution guildelines for
MetaBrainz projects it doesn't make much sense to copy them around. It's
much easier to reference that repository and than keeping
CONTRIBUTING.md files in sync in all the projects.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,16 +1,15 @@
 # Contribution guidelines 📜
 
-**Thank you for you interest in contributing to the CritiqueBrainz project!** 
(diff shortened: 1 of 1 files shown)


gentlecat:
@mwiencek, would you mind checking the grammar? :)

mwiencek:
Here are the corrections and changes I would suggest: https://github.com/mwiencek/critiquebrainz/commit/b03b781dd20476ba9479231322ff5a00ee9a3ef0

gentlecat:
Thanks a lot!

PR #152, opened 2017-08-21
Constrain rating storage conversions within one module

ps2611:
Now, the whole CB codebase can use the scale of 1-5 for ratings.
For storage, ratings will be converted to the scale 0-100 right before inserting them into the database.

### DIFF
--- a/critiquebrainz/db/avg_rating.py
+++ b/critiquebrainz/db/avg_rating.py
@@ -116,4 +116,4 @@ def get(entity_id, entity_type):
         avg_rating = dict(avg_rating)
         avg_rating["rating"] = round(avg_rating["rating"] / 20, 1)
 
-    return (avg_rating)
+    return avg_rating
--- a/critiquebrainz/db/avg_rating.py
+++ b/critiquebrainz/db/avg_rating.py
@@ -113,5 +113,7 @@ def get(entity_id, entity_type):
         if not avg_rating:
             raise db_exceptions.NoDataFoundException("""No rating for the entity with ID: {id} and Type: {type}""".
                                                      format(id=entity_id, type=entity_type))
+        avg_rating = dict(avg_rating)
+        avg_rating["rating"] = round(avg_rating["rating"] / 20, 1)
 
-    return dict(avg_rating)
+    return (avg_rating)
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -19,6 +19,7 @@ ENTITY_TYPES = [
     "place",
     "release_group",
 ]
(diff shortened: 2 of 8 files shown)


critiquebrainz/db/avg_rating.py
```diff
@@ -113,5 +113,7 @@ def get(entity_id, entity_type):
         if not avg_rating:
             raise db_exceptions.NoDataFoundException("""No rating for the entity with ID: {id} and Type: {type}""".
                                                      format(id=entity_id, type=entity_type))
+        avg_rating = dict(avg_rating)
+        avg_rating["rating"] = round(avg_rating["rating"] / 20, 1)
 
-    return dict(avg_rating)
+    return (avg_rating)
```
gentlecat: Parentheses around a return value are unnecessary.

ps2611: Removed

critiquebrainz/db/review.py
```diff
@@ -19,6 +19,7 @@
     "place",
     "release_group",
 ]
+RATING_SCALE_1_5 = {20: 1, 40: 2, 60: 3, 80: 4, 100: 5}
```
gentlecat: Since there is a module specifically related to ratings it might be better to put these values there instead.

critiquebrainz/db/revision.py
```diff
@@ -5,6 +5,10 @@
 from critiquebrainz.db import exceptions as db_exceptions
 import sqlalchemy
 
+VALID_RATING_VALUES = [None, 1, 2, 3, 4, 5]
+RATING_SCALE_0_100 = {1: 20, 2: 40, 3: 60, 4: 80, 5: 100}
+RATING_SCALE_1_5 = {20: 1, 40: 2, 60: 3, 80: 4, 100: 5}
```
gentlecat: Again, these also belong in the rating module.

critiquebrainz/db/review.py
ps2611: Sorry, didn't get you. Which module are you talking about?

gentlecat: I guess there's no `rating` module there, but https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/db/avg_rating.py is probably the closes. In any case, a more relevant place to have these definitions.

ps2611: Will `__init__.py` be a good place for this? I know avg_rating.py is related to ratings but it doesn't seem a good fit for rating conversion scales.

gentlecat: Sure, not ideal but better.

PR #154, opened 2017-08-24
CB-244: Rating support in the API

ps2611:
Dependent on #152

### DIFF
--- a/critiquebrainz/ws/parser.py
+++ b/critiquebrainz/ws/parser.py
@@ -21,8 +21,7 @@ class Parser(object):
         _dict = cls.get_dict(src)
         if key not in _dict and not optional:
             raise MissingDataError(key)
-        else:
-            return _dict.get(key)
+        return _dict.get(key)
 
     @classmethod
     def bool(cls, src, key, optional=False):
--- a/critiquebrainz/ws/parser.py
+++ b/critiquebrainz/ws/parser.py
@@ -17,31 +17,27 @@ class Parser(object):
             return request.json
 
     @classmethod
-    def get_key(cls, src, key):
+    def get_key(cls, src, key, optional):
         _dict = cls.get_dict(src)
-        _k = _dict.get(key)
-        return _k
+        if key not in _dict and not optional:
+            raise MissingDataError(key)
+        else:
+            return _dict.get(key)
 
     @classmethod
     def bool(cls, src, key, optional=False):
-        _b = cls.get_key(src, key)
+        _b = cls.get_key(src, key, optional)
         if _b is None:
-            if optional:
-                return None
-            else:
-                raise MissingDataError(key)
+            return None
         if isinstance(_b, bool) is False:
             raise ParserError(key, 'is not bool')
         return _b
 
     @classmethod
     def string(cls, src, key, min=None, max=None, valid_values=None, optional=False):
-        _s = cls.get_key(src, key)
+        _s = cls.get_key(src, key, optional)
         if _s is None:
-            if optional:
-                return None
-            else:
-                raise MissingDataError(key)
+            return None
         _s_len = len(_s)
         if max is not None and _s_len > max:
             raise ParserError(key, 'too long (max=%d)' % max)
@@ -53,12 +49,9 @@ class Parser(object):
 
     @classmethod
     def int(cls, src, key, min=None, max=None, optional=False):
-        _i = cls.get_key(src, key)
+        _i = cls.get_key(src, key, optional)
         if _i is None:
-            if optional:
-                return None
-            else:
-                raise MissingDataError(key)
+            return None
         if _i.isdigit() is False:
             raise ParserError(key, 'NaN')
         else:
@@ -71,24 +64,18 @@ class Parser(object):
 
     @classmethod
     def uuid(cls, src, key, optional=False):
-        _u = cls.get_key(src, key)
+        _u = cls.get_key(src, key, optional)
         if _u is None:
-            if optional:
-                return None
-            else:
-                raise MissingDataError(key)
+            return None
         if not validate_uuid(_u):
             raise ParserError(key, 'not valid UUID')
         return _u
 
     @classmethod
     def uri(cls, src, key, optional=False):
-        _u = cls.get_key(src, key)
+        _u = cls.get_key(src, key, optional)
         if _u is None:
-            if optional:
-                return None
-            else:
-                raise MissingDataError(key)
+            return None
         d = urllib.parse.urlparse(_u)
         if d.scheme not in ['http', 'https'] or not d.netloc:
             raise ParserError(key, 'not valid URI')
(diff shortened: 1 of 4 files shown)


critiquebrainz/ws/review/views.py
```diff
@@ -231,21 +233,30 @@ def review_modify_handler(review_id, user):
 
     :resheader Content-Type: *application/json*
     """
-
+    # TODO(psolanki): One caveat is that client will have to pass the unmodified parameter with previous revision's value
+    #                 otherwise it will be set to None
```
ps2611: Need some help to with this. I tried some solutions but they were not promising.

ps2611: Need some help to with this. I tried some solutions but they were not promising.

ps2611:
@gentlecat Had to do it this way. For some reason, requesting a review option was not showing up for me.

gentlecat:
OK, I guess that's admin-only feature. Feel free to just cc me in a description like you've done here.

critiquebrainz/ws/review/views.py
gentlecat: Which parameter? I don't quite understand the issue.

critiquebrainz/ws/review/views.py
```diff
@@ -231,21 +233,30 @@ def review_modify_handler(review_id, user):
...
     def fetch_params():
-        text = Parser.string('json', 'text', min=REVIEW_MIN_LENGTH, max=REVIEW_MAX_LENGTH)
-        return text
+        text = Parser.string('json', 'text', min=REVIEW_TEXT_MIN_LENGTH, max=REVIEW_TEXT_MAX_LENGTH, optional=True)
+        rating = Parser.int('json', 'rating', min=REVIEW_RATING_MIN, max=REVIEW_RATING_MAX, optional=True)
+        if text is None and rating is None:
+            raise InvalidRequest(desc='Text part and rating part of a review cannot be None simultaneously')
```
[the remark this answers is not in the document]
ps2611: This seems to be a part of the issue mentioned above.

critiquebrainz/ws/review/views.py
ps2611: Suppose the review is of text+rating type. If the user wants to change just the rating and passes just the rating parameter, text will be set to None(because its optional). In order to do so, text (unmodified parameter) must be passed with previous revision's text value. 

Now this does not seem to be a very elegant solution.

gentlecat: Since that's an interface for updating reviews I don't see anything wrong without skipping updates for missing fields. If *text* is not provided then it doesn't need to be modified.

ps2611: But, what if the user wants to remove the text part from text+rating review.

critiquebrainz/ws/review/views.py
```diff
@@ -530,6 +545,8 @@ def fetch_params():
     vote = fetch_params()
     if str(review["user_id"]) == user.id:
         raise InvalidRequest(desc='You cannot rate your own review.')
+    if review['rating'] and review["text"] is None:
```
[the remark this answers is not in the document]
ps2611: Yes. Its removed.

critiquebrainz/ws/review/views.py
gentlecat: They can pass `None`.

ps2611: There's a problem with this. Since the parameters are optional, not passing them is equal to passing them with None.

gentlecat: Is there a way to check if request JSON has a field or not?

ps2611: Added this check in the parser.

ps2611:
Cleaned up some code in the parser module. It also includes the check if a field is present in the request. Updates will be skipped for the missing fields and all the other conditions remain the same.

ps2611:
With code clean-up in parser, two things are added. A check is included if the request has a certain field and a mandatory field can now have its value as `None`.

ps2611:
Are the changes in the commit https://github.com/metabrainz/critiquebrainz/pull/154/commits/c467a774b3400c3b969ef40e9a093569955e52d7 final? Should I update the docs accordingly?

critiquebrainz/ws/parser.py
```diff
@@ -17,31 +17,27 @@ def get_dict(cls, src):
...
 
     @classmethod
-    def get_key(cls, src, key):
+    def get_key(cls, src, key, optional):
         _dict = cls.get_dict(src)
-        _k = _dict.get(key)
-        return _k
+        if key not in _dict and not optional:
+            raise MissingDataError(key)
+        else:
```
gentlecat: `else` here looks unnecessary.

critiquebrainz/ws/review/views.py
```diff
@@ -8,16 +8,18 @@
...
 from brainzutils import cache
 
 review_bp = Blueprint('ws_review', __name__)
 
-REVIEW_MAX_LENGTH = 100000
-REVIEW_MIN_LENGTH = 25
+REVIEW_TEXT_MAX_LENGTH = 100000
+REVIEW_TEXT_MIN_LENGTH = 25
+REVIEW_RATING_MAX = 5
+REVIEW_RATING_MIN = 1
```
gentlecat: These constants should be defined in the database-related package.

critiquebrainz/ws/review/views.py
```diff
@@ -233,19 +235,36 @@ def review_modify_handler(review_id, user):
     """
 
     def fetch_params():
-        text = Parser.string('json', 'text', min=REVIEW_MIN_LENGTH, max=REVIEW_MAX_LENGTH)
-        return text
+        try:
+            text = Parser.string('json', 'text', min=REVIEW_TEXT_MIN_LENGTH, max=REVIEW_TEXT_MAX_LENGTH)
+        except MissingDataError:
+            text = 'same'                        # Assign temporary value which indicates no modification
```
gentlecat: This doesn't seem right. What if the review text is actually `same`, then the comparisons afterwards are going to break. Sure, it doesn't match the constraints for the text, so it's currently not possible to write a review like that. But this might change in the future and in any case that seems very hacky way of doing things.

critiquebrainz/ws/review/views.py
```diff
@@ -233,19 +235,36 @@ def review_modify_handler(review_id, user):
...
     if str(review["user_id"]) != user.id:
         raise AccessDenied
-    text = fetch_params()
+    text, rating = fetch_params()
+    if text == 'same':
+        text = review['text']
+    if rating == 0:
+        rating = review['rating']
+    if (text == review['text']) and (rating == review['rating']):
+        raise InvalidRequest(desc='Either text or rating should be edited to update the review')
```
gentlecat: I think we discussed this, but this shouldn't be a requirement. Users don't need to think about that.

critiquebrainz/ws/review/views.py
```diff
@@ -233,19 +235,36 @@ def review_modify_handler(review_id, user):
...
+        return text, rating
 
     review = get_review_or_404(review_id)
     if review["is_hidden"]:
         raise NotFound("Review has been hidden.")
     if str(review["user_id"]) != user.id:
         raise AccessDenied
-    text = fetch_params()
+    text, rating = fetch_params()
+    if text == 'same':
```
gentlecat: Yeah, this is not good.

critiquebrainz/ws/review/views.py
```diff
@@ -233,19 +235,36 @@ def review_modify_handler(review_id, user):
...
+        text = review['text']
+    if rating == 0:
```
gentlecat: Same thing with rating, really. What happens if we decide to change the range to support `0`?

gentlecat:
> Should I update the docs accordingly?

Documentation updates for endpoints that are changing should be here as well.

critiquebrainz/ws/review/views.py
ps2611: Agreed. This is not subtle. I added some changes which are much cleaner than this. Don't know why I couldn't think of such a simple solution earlier.

critiquebrainz/ws/review/views.py
ps2611: Ok. If none of the parameters are modified, it won't throw an exception and won't make a new revision as well.

PR #155, opened 2017-08-24
CB-231: Retrieve data from database directly for browse review pages

ferbncode:
Modifed `jinja2` filter for fetching entities by id and added code for retrieving data directly from database for browse review (`/review`) pages.

### DIFF
--- a/critiquebrainz/frontend/external/musicbrainz_db/entities.py
+++ b/critiquebrainz/frontend/external/musicbrainz_db/entities.py
@@ -10,37 +10,37 @@ def get_multiple_entities(entities):
         entites: List of tuples containing the entity_id and the entity_type.
 
     Returns:
-        Dictionary containing the entities keyed by their mbid.
+        Dictionary containing the basic information related to the entites.
+        {
+            "id": uuid,
+            "name/title": str,
+        }
+        Information related to the artists of release groups and the
+        coordinates of the places is also included.
     """
     entities_info = {}
     release_group_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'release_group', entities)]
     place_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'place', entities)]
     event_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'event', entities)]
-    if release_group_mbids:
-        release_groups = fetch_multiple_release_groups(
-            release_group_mbids,
-            includes=['artists'],
-        )
-        entities_info.update(release_groups)
-    if place_mbids:
-        places = fetch_multiple_places(
-            place_mbids,
-        )
-        entities_info.update(places)
-    if event_mbids:
-        events = fetch_multiple_events(
-            event_mbids,
-        )
-        entities_info.update(events)
+    entities_info.update(fetch_multiple_release_groups(
+        release_group_mbids,
+        includes=['artists'],
+    ))
+    entities_info.update(fetch_multiple_places(
+        place_mbids,
+    ))
+    entities_info.update(fetch_multiple_events(
+        event_mbids,
+    ))
     return entities_info
 
 
 def get_entity_by_id(id, type='release_group'):
     """A wrapper to call the correct get_*_by_id function."""
     if type == 'release_group':
-        entity = get_release_group_by_id(id)
+        entity = get_release_group_by_id(str(id))
     elif type == 'place':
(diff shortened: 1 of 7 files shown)


critiquebrainz/frontend/external/musicbrainz_db/entities.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def get_multiple_entities(entities):
+    """Fetch multiple entities using their mbids.
+
+    Args:
+        entites: List of tuples containing the entity_id and the entity_type.
+
+    Returns:
+        Dictionary containing the entities keyed by their mbid.
```
gentlecat: Can you also mention what the "entity" in the output actually is? Perhaps describing the output for each type is unnecessary, but at least mentioning what kind of information is expected there and what that depends on might be nice.

critiquebrainz/frontend/external/musicbrainz_db/entities.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    entities_info = {}
+    release_group_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'release_group', entities)]
+    place_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'place', entities)]
+    event_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'event', entities)]
+    if release_group_mbids:
```
gentlecat: This and other checks below seem redundant. If an empty list is passed we expect an empty output anyway, right?

critiquebrainz/frontend/external/musicbrainz_db/entities.py
```diff
@@ -0,0 +1,46 @@
...
+        release_groups = fetch_multiple_release_groups(
+            release_group_mbids,
+            includes=['artists'],
+        )
+        entities_info.update(release_groups)
```
gentlecat: Also, you can just chain calls. No need to assign a variable here I think.

critiquebrainz/frontend/external/musicbrainz_db/entities.py
```diff
@@ -0,0 +1,46 @@
+from critiquebrainz.frontend.external.musicbrainz_db.release_group import fetch_multiple_release_groups, get_release_group_by_id
+from critiquebrainz.frontend.external.musicbrainz_db.place import fetch_multiple_places, get_place_by_id
+from critiquebrainz.frontend.external.musicbrainz_db.event import fetch_multiple_events, get_event_by_id
+
+
+def get_multiple_entities(entities):
+    """Fetch multiple entities using their mbids.
```
gentlecat: MBIDs

critiquebrainz/frontend/external/musicbrainz_db/entities.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Args:
+        entites: List of tuples containing the entity_id and the entity_type.
```
gentlecat: Why `entity_id` and `entity_type` and not simply `entity ID` and `entity type`? It would make sense to, for example, reference dictionary keys exactly by name, but in this case I don't see any point.

critiquebrainz/frontend/external/musicbrainz_db/entities.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Returns:
+        Dictionary containing the basic information related to the entites.
```
gentlecat: entities

critiquebrainz/frontend/views/test/test_review.py
```diff
@@ -1,8 +1,29 @@
...
 from critiquebrainz.frontend.testing import FrontendTestCase
 import critiquebrainz.db.review as db_review
 from critiquebrainz.db.user import User
 import critiquebrainz.db.users as db_users
 import critiquebrainz.db.license as db_license
+import critiquebrainz.frontend.external.musicbrainz_db.release as mb_release
+from flask import current_app
+
+
+def mock_get_entity_by_id(id, type='release_group'): # pylint: disable=unused-argument
```
gentlecat: Why is `id` unused?

ferbncode: `id` was unused as all tests in `test_review.py` use same entity ID for each type. I have fixed this as there might be other entities which may be tested in future. Thanks for pointing it out.

PR #158, opened 2017-09-05
CB-231: Retrieve release group data directly from db in mapping.py

ferbncode:
Changed way of fetching release group info in `mapping.py`. Also fixes https://sentry.metabrainz.org/share/issue/342e32383839/


alastair:
Can you also add a test which will fail before this patch is applied?

ferbncode:
This patch actually doesn't fix the sentry error :sweat_smile:. It just changes the way of fetching release group info in `mapping.py` by retrieving data directly from the database. The sentry error results from deletion of the release group from the MusicBrainz database whose reviews exist in CritiqueBrainz. I've opened a ticket https://tickets.metabrainz.org/browse/CB-269 for this.

paramsingh:
What is the status of this PR? If we've opened CB-269 for the sentry error, I think this is okay to review?

PR #168, opened 2017-12-11
CB-253: Replace Administrator strings with Moderator

TheJYKoder:
Replaced instances of 'administrator' with 'moderator'

### DIFF
--- a/critiquebrainz/frontend/login/__init__.py
+++ b/critiquebrainz/frontend/login/__init__.py
@@ -43,7 +43,7 @@ def admin_view(f):
     @wraps(f)
     def decorated(*args, **kwargs):
         if not current_user.is_admin():
-            raise Unauthorized(lazy_gettext('You must be an administrator to view this page.'))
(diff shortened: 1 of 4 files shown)


anshuman73:
Also, I'd prefer you squash the commits and give them a generic commit message like "Changed administrator to moderator" rather than the messages you have now for each of the edited files - It's slightly more explaining in nature :)

anshuman73:
Also, although not so important, considering the users never see this, But can you make the changes in the config files as well ? It'll probably help to cause less confusion to developers as well :)

Here - https://github.com/metabrainz/critiquebrainz/blob/master/custom_config.py.example#L54
and 
here - https://github.com/metabrainz/critiquebrainz/blob/master/default_config.py#L48

anshuman73:
Hi, please also edit the PR title to "CB-253: Replace Administrator strings with Moderator"

You should be able to see a small edit button (pencil-shaped) right beside the title of this Pull Request, click that and edit it please :)

PR #164, opened 2017-12-07
CB-55: Add Label for Frequently Used/Other Languages in Review create page

IsaacZanoria:
thanks to samj1912 for the help in the irc

PR #170, opened 2018-01-06
CB-263: Update publication date of draft reviews on CritiqueBrainz when they get published

IsaacZanoria:


ferbncode:
@IsaacZanoria Please make a branch off the current master (and then add commits related to the issue only). It seems like you made a branch from PR #164.

PR #171, opened 2018-01-08
CB-263: Update publication date of draft reviews on CritiqueBrainz when they get published

IsaacZanoria:
add new column (publish_time) to review table which is only updated when a review is published without saving as a draft or changed from a draft to a published review.  add to get_by_id() so that it gets put in the returned dictionary, and change html tempaltes to use the value from the new key


critiquebrainz/db/review.py
```diff
@@ -220,6 +222,10 @@ def update(review_id, *, drafted, text=None, rating=None, license_id=None, langu
     if is_draft is not None:
         if not drafted and is_draft:  # If trying to convert published review into draft
             raise db_exceptions.BadDataException("Converting published reviews back to drafts is not allowed.")
+        if drafted and (not is_draft):
```
naiveaiguy: Won't it be impossible for `not is_draft` to be `True` here?

critiquebrainz/db/review.py
```diff
@@ -322,7 +334,7 @@ def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None,
                  offset=None):
     """Get a list of reviews.
 
-    This function provides several filters that can be used to select a subset of reviews.
+    This function provides several  filters that can be used to select a subset of reviews.
```
anshuman73: Hi, it seems a space was added here by mistake

IsaacZanoria: oh, yeah, you're right, I'll remove that right away

IsaacZanoria: sorry about making more problems with one of the commits; I'm on a school computer and not having fun using the in-browser editor

admin/sql/create_tables.sql
```diff
@@ -49,17 +49,18 @@ ALTER TABLE oauth_token ADD CONSTRAINT oauth_token_access_token_key UNIQUE (acce
...
+    id              UUID         NOT NULL DEFAULT uuid_generate_v4(),
+    entity_id       UUID         NOT NULL,
+    entity_type     entity_types NOT NULL,
+    user_id         UUID         NOT NULL,
+    edits           INTEGER      NOT NULL,
+    is_draft        BOOLEAN      NOT NULL,
+    is_hidden       BOOLEAN      NOT NULL,
+    license_id      VARCHAR      NOT NULL,
+    language        VARCHAR(3)   NOT NULL,
+    publish_time    TIMESTAMP,
```
ferbncode: Let us make that `publish_date` to be more clear. :)

critiquebrainz/db/review.py
```diff
@@ -83,6 +84,7 @@ def get_by_id(review_id):
                    review.language,
                    review.source,
                    review.source_url,
+                   review.publish_time,
```
ferbncode: Since we are adding another column to the table (for publish date), it is important to remove the subquery for creating `created_time` and the corresponding entry of `created_time.created` in `SELECT`, as that part is now unnecessary.

critiquebrainz/db/review.py
ferbncode: I think there is no need of the parentheses. 
```
if drafted and not is draft:
    ... 
```

critiquebrainz/db/review.py
ferbncode: Also, for adding the `Last Revised: ` information, you can make use of the `revision.timestamp`, so I would suggest it to rename the column as `last_revision_timestamp`.

ferbncode:
@IsaacZanoria Since it is a `schema-change` PR, you need to add a file in `admin/schema_changes` (`9.sql`) for altering table `review`.

ferbncode:
@IsaacZanoria There are more changes to be made for this PR to be merged:
+ Update the `SCHEMA_VERSION` variable in the `db` module.
+ Update `_TABLES` in `data/dump_manager.py` to include column `published_on`. Also, later the CB dumps need to be updated for this schema change.

ferbncode:
@brainzbot test this please

ferbncode:
@IsaacZanoria See the Jenkins errors to fix the tests (`Pytest`)

admin/schema_changes/9.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE review ADD COLUMN published_on TIMESTAMP;
```
ferbncode: There is no need of the `BEGIN` and `COMMIT` statements here.

critiquebrainz/frontend/views/__init__.py
```diff
@@ -1,6 +1,7 @@
 import critiquebrainz.db.avg_rating as db_avg_rating
 import critiquebrainz.db.exceptions as db_exceptions
 
+
```
ferbncode: This is unnecessary. :)

IsaacZanoria: sorry, ive seen a pep8 warning about "2 blank lines expected" in tests, and figured i'd just fix it

critiquebrainz/db/review_test.py
```diff
@@ -196,7 +196,7 @@ def test_list_reviews(self):
         self.assertEqual(count, 1)
         self.assertEqual(len(reviews), 1)
 
-        reviews, count = db_review.list_reviews(sort="created")
+        reviews, count = db_review.list_reviews(sort="")
```
ferbncode: I think you should put `published_on` as an argument here. This would test if the reviews are fetched successfully if ordered by publish date.

critiquebrainz/db/review.py
```diff
@@ -93,23 +94,12 @@ def get_by_id(review_id):
                    "user".show_gravatar,
                    "user".musicbrainz_id,
                    "user".is_blocked,
-                   created_time.created,
                    license.full_name,
                    license.info_url
               FROM review
               JOIN revision ON revision.review_id = review.id
               JOIN "user" ON "user".id = review.user_id
               JOIN license ON license.id = license_id
```
ferbncode: Please make sure that the query is right. I see that the `WHERE` clause is now missing. You can fix the query by adding a `WHERE` clause for the `review_id`. :)

critiquebrainz/db/review_test.py
IsaacZanoria: sorry, i already have the "published_on" argument set up, i think i was in the middle of adding it here and got distracted, and ended up leaving it blank

ferbncode:
@brainzbot retest this please

ferbncode:
I think there is a bug in `pylint` version, which is the reason of all the new `pylint` errors (https://github.com/PyCQA/pylint/issues/1609). Could you please try updating the `pylint` version to `1.8.1`. It should fix some `pylint` errors. The rest of the PR looks good :)

ferbncode:
Don't worry about the `pylint` fixes. It seems a version error and should be solved in a different PR. 

Also, just saw your query on IRC regarding sorting on `published_on` in the `db_review.list_reviews` function. There is an argument `include_drafts` in the function set to `False` that should take care of our unpublished reviews. :)

critiquebrainz/db/review_test.py
ferbncode: No problem :)

paramsingh:
The update script that accompanies this should also fill out the correct values in the new column added for all existing rows.

admin/schema_changes/9.sql
```diff
@@ -0,0 +1 @@
...
```
paramsingh: We'd also want to fill up the `published_on` column with values for old reviews. 

And this should be inside a transaction.

paramsingh:
@brainzbot add to whitelist

paramsingh:
@ferbncode, this is ready for review, would you mind taking a look at the last two commits I made?

PR #172, opened 2018-01-18
CB-272: Update pylint version and fix the `Unused variable '__class__'` error for unit tests.

ferbncode:
Presently, this https://github.com/PyCQA/pylint/issues/1609 issue causes the tests to fail with the present version of `pylint`.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -9,7 +9,7 @@
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=missing-docstring,protected-access,invalid-name,global-statement,fixme,too-many-locals,redefined-builtin,broad-except,line-too-long,too-many-arguments,too-many-instance-attributes,redefined-outer-name,attribute-defined-outside-init,anomalous-backslash-in-string,too-many-statements,too-few-public-methods,redefined-argument-from-local,too-many-return-statements,too-many-function-args,redefined-argument-from-local,too-many-public-methods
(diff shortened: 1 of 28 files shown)


pylintrc
```diff
@@ -9,7 +9,7 @@
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=missing-docstring,protected-access,invalid-name,global-statement,fixme,too-many-locals,redefined-builtin,broad-except,line-too-long,too-many-arguments,too-many-instance-attributes,redefined-outer-name,attribute-defined-outside-init,anomalous-backslash-in-string,too-many-statements,too-few-public-methods,redefined-argument-from-local,too-many-return-statements,too-many-function-args,redefined-argument-from-local,too-many-public-methods
+disable=missing-docstring,protected-access,invalid-name,global-statement,fixme,too-many-locals,redefined-builtin,broad-except,line-too-long,too-many-arguments,too-many-instance-attributes,redefined-outer-name,attribute-defined-outside-init,anomalous-backslash-in-string,too-many-statements,too-few-public-methods,redefined-argument-from-local,too-many-return-statements,too-many-function-args,redefined-argument-from-local,too-many-public-methods,inconsistent-return-statements, keyword-arg-before-vararg
```
paramsingh: Are we sure these rules need to be ignored? I think this could be fixed in the code instead maybe?

paramsingh: Are we sure these rules need to be ignored? I think this could be fixed in the code instead maybe?

paramsingh: As discussed, I've opened a ticket for this (https://tickets.metabrainz.org/browse/CB-273). I'm gonna merge this considering all builds are broken right now, and we can continue fixing CB-273 later.

PR #174, opened 2018-02-11
CB-276: Initialize db and create tables in test container before running tests

ferbncode:
As decided in discussions [here](https://github.com/metabrainz/critiquebrainz/pull/164#issuecomment-354597289), it is good to create db tables in test container before running tests (and not just rely on pytest's  `setUp()` method .


ferbncode:
@brainzbot retest this please

PR #177, opened 2018-02-12
Change container name in Testing

Prakriti-nith:
This pull request is for the bug fix in the documentation.
If the user runs `$ docker logs -f critiquebrainz_web_test_1`, following error occurs:
Error: No such container: critiquebrainz_web_test_1


ferbncode:
@brainzbot add to whitelist

paramsingh:
@prakriti-nith instead of running tests in the background and then running `docker logs`, I think just telling people to run `docker-compose -f docker/docker-compose.test.yml up --build` would be better

PR #178, opened 2018-02-13
CB-225: Add Sign in button on error pages

Prakriti-nith:
This pull request is the bug fix on error pages. 
It adds the missing sign in button on error pages.
Dropdown menu is purposely not provided as it is mentioned in the comments (https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/frontend/templates/navbar.html#L19) that while loading, user info might fail and page will not render.

### DIFF
--- a/critiquebrainz/frontend/templates/navbar.html
+++ b/critiquebrainz/frontend/templates/navbar.html
@@ -32,6 +32,10 @@
               </ul>
             </li>
           {% endif %}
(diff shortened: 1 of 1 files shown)


Prakriti-nith:
screenshot from 2018-02-14 15-04-08
It adds the sign in button on navigation bar on error pages

PR #162, opened 2017-12-05
CB-253: Changed administrator to moderator

saidsellai:
I couldn't figure out the Babel/pybabel although I tried extracting/compiling the .mo and .pot files

PR #161, opened 2017-12-03
changed administrator to moderator for consistency

saidsellai:
GCI task

PR #179, opened 2018-02-13
CB-253: Change moderator(s) to administrator(s)

Prakriti-nith:
As mentioned in the issue "Admin" and "moderator" were used interchangeably.
So I have changed all the occurances of moderator/moderators to administrator/administrators.


paramsingh:
@prakriti-nith, we have 4 patches for this bug. Would you mind taking a look at the other ones and seeing if you've missed any places? I'd like to get the bug fixed in this one.

paramsingh:
#168 #161 #162

PR #181, opened 2018-02-16
CB-252: Make whole row selectable in search results

dpmittal:
In this patch, the user does not need to check the radio button, by clicking on the row itself the corresponding radio button gets checked.

### DIFF
--- a/critiquebrainz/frontend/templates/search/selector.html
+++ b/critiquebrainz/frontend/templates/search/selector.html
@@ -173,6 +173,11 @@
           alert({{ _("Failed to load more search results!") | tojson }});
         });
       }
+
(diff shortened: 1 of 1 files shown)


paramsingh:
@brainzbot add to whitelist

paramsingh:
@ferbncode review this please?

critiquebrainz/frontend/templates/search/selector_results.html
```diff
@@ -66,3 +66,8 @@
 
 {% endif %}
 {% endfor %}
+<script type="text/javascript">
```
ferbncode: I think it would be best to move this code to `selector.html` in the `script` block. Also, please add a comment to explain why we are doing this.

dpmittal:
@ferbncode made the suggested changes :)

PR #182, opened 2018-02-20
CB-280: Fix Spotify mapping page suggestions

ferbncode:
Fix missing suggestions after caching and added a test for checking cached spotify results. Since the same list object was passed to the function `get_multiple_albums` (in `spotify.py`), `album_ids` were deleted from the original object and cached results showed no suggestions instead.

### DIFF
--- a/critiquebrainz/frontend/external/spotify_test.py
+++ b/critiquebrainz/frontend/external/spotify_test.py
@@ -12,6 +12,18 @@ class FakeSpotifyResponse:
 
     @classmethod
     def fromSpotifyIds(cls, spotify_ids):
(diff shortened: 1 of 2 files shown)


critiquebrainz/frontend/external/spotify_test.py
```diff
@@ -9,6 +10,13 @@ def __init__(self, query):
     def json(self):
         return dict(url=self.url)
 
+    @classmethod
+    def fromSpotifyIds(cls, spotify_ids):
```
paramsingh: Should probably document the structure that this method should return

PR #183, opened 2018-02-20
CB-55: Add OptGroupSelectField for frequently used languages

man-of-code:
Added `OptGroupSelectField` from `wtforms-components` for separating frequently used languages.

### DIFF
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -40,6 +40,22 @@ def to_dict(review, confidential=False):
     return review
 
 
+def get_top_languages():
+    """
+        List of the top ten languages (only the language codes) in CritiqueBrainz.
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+               SELECT review.language
+                 FROM review
+             GROUP BY review.language
+             ORDER BY COUNT(review.id) DESC
+                LIMIT 10
+            """))
+        languages = result.fetchall()
+        return [language[0] for language in languages]
+
+
 def get_by_id(review_id):
     """Get a review by its ID.
 
--- a/critiquebrainz/frontend/forms/review.py
+++ b/critiquebrainz/frontend/forms/review.py
@@ -1,11 +1,12 @@
 from flask_wtf import Form
 from flask_babel import lazy_gettext, Locale
-from wtforms import TextAreaField, RadioField, SelectField, BooleanField, StringField, validators, IntegerField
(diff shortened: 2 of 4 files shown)


ferbncode:
@brainzbot add to whitelist

critiquebrainz/db/review.py
```diff
@@ -40,6 +40,22 @@ def to_dict(review, confidential=False):
     return review
 
 
+def get_top_languages():
+    """
+        List of the top ten languages (only the language codes) in CritiqueBrainz.
+    """
```
ferbncode: Make sure that the docstring matches the guidelines: https://github.com/metabrainz/guidelines/blob/master/Python.md#docstrings

critiquebrainz/frontend/forms/review.py
```diff
@@ -20,13 +21,32 @@ def __call__(self, form, field):
...
 
 
-# Loading supported languages
-languages = []
-for language_code in supported_languages:
-    try:
-        languages.append((language_code, Locale(language_code).language_name))
-    except UnknownLocaleError:
-        languages.append((language_code, pycountry.languages.get(iso639_1_code=language_code).name))
+def get_languages():
```
ferbncode: It would be great if you could add a docstring for this function too.

ferbncode:
The tests are failing because the `create_app()` function is called in `python manage.py init_db --skip-create-db --test-db` command in test container which fails to recognize `db` (which the `get_top_language()` function is using). We need to figure out a way to fix this in tests.

We can also change the way `frequently_used_languages` is populated. We can maintain a script for updating frequently used languages (like in MB: https://github.com/metabrainz/musicbrainz-server/blob/master/admin/SetLanguageFrequencies)

paramsingh:
@ferbncode: Seems like MB has a cron job running daily to calculate language frequencies: https://github.com/metabrainz/musicbrainz-server/blob/master/admin/cron/daily.sh#L87, we could do something like that too.

man-of-code:
Okay, I'll add the `doctsrings` according to the guidelines. Thanks for reviewing @ferbncode .

ferbncode:
I think the method for calculating `frequently_used_languages` should be similar to MusicBrainz. @man-of-code , it would be awesome if you could write a script that executes the exact same query, updates cache (store the result in a namespace 'frequent-languages'). Then we can add that script to cron, so that it gets updated every day (or week).

man-of-code:
@ferbncode I'll get to it, right away.

paramsingh:
Closing this for now, we can do a different PR for calculating frequently used languages first.

PR #186, opened 2018-03-02
CB-269: Handle reviewed entities deleted in MB database

ferbncode:
Todo:

- [x] Docstrings
- [x] Some Tests

### DIFF
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -637,7 +637,7 @@ def reviewed_entities(*, entity_ids, entity_type):
         entity_ids: List of ID(s) of the entities.
         entity_type: Type of the entities.
     Returns:
-        Dictionary stating whether an entity has been reviewed keyed by the entity ID.
+        List of entity ID(s) that have a review in the database.
     """
     with db.engine.connect() as connection:
         results = connection.execute(sqlalchemy.text("""
@@ -650,7 +650,4 @@ def reviewed_entities(*, entity_ids, entity_type):
             "entity_ids": tuple(entity_ids),
         })
         reviewed_ids = [str(row[0]) for row in results.fetchall()]
-    reviewed = {}
-    for entity_id in entity_ids:
-        reviewed[entity_id] = True if entity_id in reviewed_ids else False
-    return reviewed
+    return reviewed_ids
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -628,3 +628,29 @@ def distinct_entities():
               FROM review
         """))
         return {row[0] for row in results.fetchall()}
+
+
+def reviewed_entities(*, entity_ids, entity_type):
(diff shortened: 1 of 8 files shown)


critiquebrainz/db/review.py
```diff
@@ -628,3 +628,29 @@ def distinct_entities():
               FROM review
         """))
         return {row[0] for row in results.fetchall()}
+
+
+def reviewed_entities(*, entity_ids, entity_type):
```
paramsingh: This function name is a little misleading, we should either rename it to `check_if_entities_reviewed` or something along those lines or keep the name and make it return a list of the gids of those entities which have been reviewed (which we're doing in `get_entities_by_gid` right now), I kinda prefer the second method.

critiquebrainz/frontend/external/musicbrainz_db/tests/release_group_test.py
```diff
@@ -78,3 +79,19 @@ def test_fetch_browse_release_groups(self):
             }
         ])
         self.assertEqual(release_groups[1], 2)
+
+    def test_unknown_release_group(self):
+        mb_release_group.get_entities_by_gids = MagicMock()
+        mb_release_group.get_entities_by_gids.return_value = {
```
paramsingh: If we mock the result of `get_entities_by_gid`, do any of the changes in the function even get tested?

critiquebrainz/db/review.py
```diff
@@ -628,3 +628,29 @@ def distinct_entities():
...
+            SELECT entity_id
+              FROM review
+             WHERE entity_type = :entity_type
+               AND entity_id IN :entity_ids
+        """), {
+            "entity_type": entity_type,
+            "entity_ids": tuple(entity_ids),
+        })
+        reviewed_ids = [str(row[0]) for row in results.fetchall()]
+    reviewed = {}
```
paramsingh: we could make this a one liner with a dict comprehension.

 ```reviewed = {entity_id: True if entity_id in reviewed_ids else False for entity_id in entity_ids}```

^^ this might not be entirely correct but you get the idea.

critiquebrainz/db/review.py
paramsingh: Also, we should add a test for this function as well.

ferbncode: I've added a test for the function `reviewed_entities`.

critiquebrainz/frontend/external/musicbrainz_db/tests/release_group_test.py
ferbncode: For improving this, I've mocked the result of the `mbdata` queries. 
I've also mocked the `reviewed_entities` function (I think it is okay to do so, since we now have a test for it in `db/review_test.py`). 
The test now handles the changes made to the function `get_entities_by_gid` as well as tests all the serialization stuff as before.

PR #188, opened 2018-03-07
Add script develop.sh

man-of-code:
This patch adds the script `develop.sh` for building and running the server up.
Copy paste from LB.

### DIFF
--- a/develop.sh
+++ b/develop.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+if [[ ! -d "docker" ]]; then
+    echo "This script must be run from the top level directory of the critiquebrainz-server source."
+    exit -1
+fi
+
+# invoke docker, but ungrep noisy influx log messages that I can't turn off
+docker-compose -f docker/docker-compose.dev.yml build && docker-compose -f docker/docker-compose.dev.yml up


develop.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+if [[ ! -d "docker" ]]; then
+    echo "This script must be run from the top level directory of the critiquebrainz-server source."
+    exit -1
+fi
+
+# invoke docker, but ungrep noisy influx log messages that I can't turn off
```
alastair: this is a listenbrainz-specific comment (which is out of date anyway). It shouldn't be here.

develop.sh
```diff
@@ -0,0 +1,9 @@
...
+docker-compose -f docker/docker-compose.dev.yml build && docker-compose -f docker/docker-compose.dev.yml up
```
alastair: we should have a `-p` flag here to call the docker-compose project `critiquebrainz` instead of the default in this case which will be `docker` (the name of the directory that the docker compose file is in)

alastair:
We took a step back and considered how we want develop.sh to work in MetaBrainz projects: https://tickets.metabrainz.org/browse/LB-326
I'm going to close this so that we can concentrate on getting a great script going that we can also use in CB

PR #190, opened 2018-03-16
Improve the setup docs a bit

paramsingh:
Added the command to convert sample dumps into bz2. And some other minor changes.


paramsingh:
Thanks for the review (and the help!)

PR #191, opened 2018-03-16
Fix cron jobs for data dumps

paramsingh:
Also add a new job for copying data over to the
ftp server

### DIFF
--- a/docker/prod/cron/jobs
+++ b/docker/prod/cron/jobs
@@ -1,12 +1,14 @@
+PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin
+PYTHONPATH=/usr/local/lib/python36.zip:/usr/local/lib/python3.6:/usr/local/lib/python3.6/lib-dynload:/usr/local/lib/python3.6/site-packages:/usr/local/lib/python3.6/site-packages/messybrainz-99.0.0.dev0-py3.6.egg
+
 # Database backup creation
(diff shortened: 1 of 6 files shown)


docker/prod/cron/jobs
```diff
@@ -1,12 +1,14 @@
...
+15 00 * * * /usr/local/bin/python manage.py dump public -l /data/dumps/tsv -r >> /var/log/public_dump_create.log 2>&1
 
 # JSON dump creation
-20 00 * * * cd /code/ && python3 manage.py dump json -l /data/dumps/json -r
+20 00 * * * /usr/local/bin/python manage.py dump json -l /data/dumps/json -r >> /var/log/json_dump_create.log 2>&1
 
 # Copy everything over...
-# TODO: Fix this
-#45 00 * * * rsync -av --delete-after -e "ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" /data/dumps/ ftp-misc@taz.mb:/media/ftp/root/pub/musicbrainz/critiquebrainz
+45 00 * * * /code/admin/rsync-dump-files.sh
```
mayhem: I think we need to start a SysWiki document about when which services are syncing to FTP -- or something to know when to do something so that we don't accidentally overload one poor server due to copy 🍝 .

paramsingh: https://github.com/metabrainz/syswiki/pull/17

PR #195, opened 2018-03-27
Change Dockerfile and push.sh and add stuff for CB Beta

paramsingh:
Add new uwsgi.ini and service files, use these according to deploy_env in the Dockerfile, make changes in push.sh.

### DIFF
--- a/docker/push.sh
+++ b/docker/push.sh
@@ -9,7 +9,7 @@
 cd "$(dirname "${BASH_SOURCE[0]}")/../"
 
 ENV=${1:-beta}
-TAG=${1:-beta}
+TAG=${2:-beta}
 docker build -t metabrainz/critiquebrainz:$TAG \
         --build-arg GIT_COMMIT_SHA=$(git rev-parse HEAD) \
         --build-arg DEPLOY_ENV=$ENV .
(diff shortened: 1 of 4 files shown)


docker/push.sh
```diff
@@ -1,13 +1,16 @@
...
-#   $ ./push.sh [tag]
+#   $ ./push.sh [env] [tag]
 
 cd "$(dirname "${BASH_SOURCE[0]}")/../"
 
-TAG_PART=${1:-latest}
-docker build -t metabrainz/critiquebrainz:$TAG_PART --build-arg GIT_COMMIT_SHA=$(git rev-parse HEAD) .
-docker push metabrainz/critiquebrainz:$TAG_PART
+ENV=${1:-beta}
+TAG=${1:-beta}
```
mayhem: You'll want to make TAG look at a second argument in order to give containers version tags. See LB.

docker/push.sh
```diff
@@ -1,13 +1,16 @@
...
+docker build -t metabrainz/critiquebrainz:$TAG \
+        --build-arg GIT_COMMIT_SHA=$(git rev-parse HEAD) \
```
mayhem: Oh, clever. We should steal this for LB too!

PR #197, opened 2018-03-28
CB-292: Show a table for stats on the home page

paramsingh:
Looks better than just the text currently present
on the site.


paramsingh:
screenshot_2018-03-29_02-10-01

PR #199, opened 2018-03-28
CB-286: Add tests for dump manager functions

ferbncode:
Added tests for `dump_manager.py` functions. The tests fail presently. `reset_db()` fails and closes connection to the db. This needs to be fixed.

PR #200, opened 2018-04-03
Refactor public function in dump_manager

ferbncode:
Refactor public function in `dump_manager.py` (Removing a lot of duplicate code) and using transaction for copying tables.

### DIFF
--- a/critiquebrainz/data/dump_manager.py
+++ b/critiquebrainz/data/dump_manager.py
@@ -205,13 +205,9 @@ def public(location, rotate=False):
     print("Creating public database dump...")
     time_now = datetime.today()
 
-    connection = db.engine.raw_connection()
-    cursor = connection.cursor()
-
     # Creating a directory where all dumps will go
     dump_dir = os.path.join(location, time_now.strftime('%Y%m%d-%H%M%S'))
     create_path(dump_dir)
-
     temp_dir = tempfile.mkdtemp()
 
     # Preparing meta files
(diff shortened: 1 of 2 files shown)


paramsingh:
Do tests in #199 pass after these changes are applied?

critiquebrainz/data/dump_manager.py
```diff
@@ -221,18 +217,51 @@ def public(location, rotate=False):
...
+
+    if rotate:
+        print("Removing old dumps (except two latest)...")
+        remove_old_archives(location, "[0-9]+-[0-9]+", is_dir=True)
+
+    print("Done!")
+
+
+def create_base_archive(dump_dir, temp_dir):
+    """Creates a dump of all license-independent information: (users, license)."""
```
paramsingh: This docstring should be improved. Add information about what this function takes as arguments and what it returns.

critiquebrainz/data/dump_manager.py
```diff
@@ -221,18 +217,51 @@ def public(location, rotate=False):
...
+    shutil.rmtree(temp_dir)  # Cleanup
+
+    if rotate:
+        print("Removing old dumps (except two latest)...")
+        remove_old_archives(location, "[0-9]+-[0-9]+", is_dir=True)
+
+    print("Done!")
+
+
+def create_base_archive(dump_dir, temp_dir):
```
paramsingh: I feel like it would be better for readability and decoupling if we created a `temp_dir` in this function instead of passing it.

critiquebrainz/data/dump_manager.py
```diff
@@ -241,83 +270,74 @@ def public(location, rotate=False):
...
 
-    # REVIEWS
-    # Archiving review tables (review, revision)
 
-    # 1. COMBINED
-    # Archiving all reviews (any license)
-    REVISION_COMBINED_SQL = """
-        SELECT {columns} FROM revision JOIN review
+def create_reviews_archive(dump_dir, temp_dir, license_id=None):
+    """Creates a dump of reviews filtered on the given license_id, their revisions and
```
paramsingh: This docstring should be improved as well. You should also add what happens when `license_id=None`.

critiquebrainz/frontend/views/review.py
```diff
@@ -19,7 +20,6 @@
 import critiquebrainz.db.review as db_review
 import critiquebrainz.db.moderation_log as db_moderation_log
 from critiquebrainz.frontend.external.musicbrainz_db.entities import get_multiple_entities, get_entity_by_id
-from langdetect import detect
```
paramsingh: This change doesn't seem related to anything in this PR.

critiquebrainz/data/dump_manager.py
```diff
@@ -221,18 +217,51 @@ def public(location, rotate=False):
...
+        with db.engine.connect() as connection:
+            with connection.begin() as transaction:
+                cursor = connection.connection.cursor()
+                try:
+                    with open(os.path.join(base_archive_tables_dir, 'user_sanitised'), 'w') as f:
+                        cursor.copy_to(f, '"user"', columns=('id', 'created', 'display_name', 'musicbrainz_id'))
+                    with open(os.path.join(base_archive_tables_dir, 'license'), 'w') as f:
+                        cursor.copy_to(f, 'license', columns=_TABLES["license"])
+                except Exception:
+                    print('Error while copying tables')
```
paramsingh: You should log the error as well. Also, the message should be something like `error while copying tables during creation of base archive` so that it is easier to understand where the error took place.

PR #201, opened 2018-04-03
CB-284: Add an option to reset tables and types to manage.py init_db

dpmittal:
Reinitialize database using ``--force`` command.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -101,7 +101,7 @@ def init_db(skip_create_db=False, test_db=False, force=False):
     click.echo("Initializing the database...")
 
     if force:
-        click.echo("Dropping exisiting tables and types...")
+        click.echo("Dropping existing tables and types...")
         data_utils.drop_tables()
         data_utils.drop_types()
         click.echo("Done!")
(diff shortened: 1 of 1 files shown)


manage.py
```diff
@@ -88,8 +88,10 @@ def clear_memcached():
               help="Skip database creation step.")
 @click.option("--test-db", "-t", is_flag=True,
               help="Initialize the test database.")
+@click.option("--force", "-f", is_flag=True,
+              help="Drop existing database and user.")
```
ferbncode: This doesn't drop the existing database and the user. Using this option, according to this implementation, drops all tables and types.

manage.py
```diff
@@ -98,6 +100,12 @@ def init_db(skip_create_db=False, test_db=False):
     """
     click.echo("Initializing the database...")
 
+    if force:
+        click.echo("Dropping Database")
```
ferbncode: This needs to be fixed too. Also, this is a temporary fix. The database and the user must be dropped using the `postgres` user (as in LB).

PR #202, opened 2018-04-05
CB-291: Call for action to MB for adding missing Cover arts. (on review and release group pages)

ferbncode:
missing-cover-art


paramsingh:
I talked to reo on IRC and the gist was that linking exact release-groups or releases would lead to bad cover art in MB from people who google cover art. The alternative is to link to the our guide on how to add cover art to MB.

https://chatlogs.metabrainz.org/brainzbot/metabrainz/2018-04-13/?msg=4159695&page=1

PR #203, opened 2018-04-22
CB-295: Add tables to support comments on reviews in CB

paramsingh:
Add two tables, very similar to how it's done for reviews.


paramsingh:
Open questions:

* Do we keep votes in the already existing `vote` table or create a new `comment_vote` table?
* Similarly, is the `comment_revision` table needed really?

PR #209, opened 2018-05-06
CB-295: Show comments on reviews and add ability to add comments.

paramsingh:
There's lots of features to add still, but I think that this is work enough to release.


paramsingh:
Future work: [CB-303](https://tickets.metabrainz.org/browse/CB-303), [CB-304](https://tickets.metabrainz.org/browse/CB-304), [CB-305](https://tickets.metabrainz.org/browse/CB-305),[CB-306](https://tickets.metabrainz.org/browse/CB-306)

paramsingh:
Good point, I missed that. Do you think it should be a part of this PR or another ticket? I don't think it would be much work for me to implement it.

PR #216, opened 2018-10-18
CB-298: Add transaction support for json dumps

code-master5:
Make JSON Dumps in a transaction.

### DIFF
--- a/critiquebrainz/data/dump_manager.py
+++ b/critiquebrainz/data/dump_manager.py
@@ -144,43 +144,46 @@ def json(location, rotate=False):
     current_app.json_encoder = DumpJSONEncoder
 
     print("Creating new archives...")
-    for license in db_license.list_licenses():
-        safe_name = slugify(license["id"])
-        with tarfile.open(os.path.join(location, "critiquebrainz-%s-%s-json.tar.bz2" %
-                                       (datetime.today().strftime('%Y%m%d'), safe_name)), "w:bz2") as tar:
-            temp_dir = tempfile.mkdtemp()
-            license_dir = os.path.join(temp_dir, safe_name)
-            create_path(license_dir)
+    with db.engine.begin() as connection:
+        for license in db_license.get_licenses_list(connection):
+            safe_name = slugify(license["id"])
+            with tarfile.open(os.path.join(location, "critiquebrainz-%s-%s-json.tar.bz2" %
+                                           (datetime.today().strftime('%Y%m%d'), safe_name)), "w:bz2") as tar:
+                temp_dir = tempfile.mkdtemp()
+                license_dir = os.path.join(temp_dir, safe_name)
+                create_path(license_dir)
 
-            # Finding entities that have reviews with current license
-            entities = db_review.distinct_entities()
-            for entity in entities:
-                entity = str(entity)
-                # Creating directory structure and dumping reviews
-                dir_part = os.path.join(entity[0:1], entity[0:2])
-                reviews = db_review.list_reviews(entity_id=entity, license_id=license["id"], limit=None)[0]
-                if reviews:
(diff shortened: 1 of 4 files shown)


ferbncode:
@brainzbot add to whitelist

critiquebrainz/db/review.py
```diff
@@ -613,22 +626,32 @@ def delete(review_id):
         db_avg_rating.update(review["entity_id"], review["entity_type"])
 
 
-def distinct_entities():
+def distinct_entities(*, trans_conn=None):
```
paramsingh: I'm not entirely happy with the `if-else` conditions here. Can we do something like this? Do the meat of the stuff in a new function which takes a connection as a parameter, change the original functions to use the new one and use the new one directly in the dump code?

```
def get_distinct_entities(connection):
     query = sqlalchemy.text("""
        SELECT DISTINCT entity_id
          FROM review
    """)
    results = connection.execute(query)
    return {row[0] for row in results.fetchall()}
   
def distinct_entities():
    with db.engine.connect() as connection:
        return get_distinct_entities(connection)
```

paramsingh: This would also mean that we don't _have_ to worry about breaking existing code.

paramsingh: yes, so we'd just use `get_distinct_entities(transaction)` in the `json()` function in `dump_manager.py`

code-master5: @paramsingh it would work for functions where only `connection` object has to be passed. What about functions where multiple arguments have to be passed and all these arguments are used to formulate the query?

For example, have a look at [list_reviews()](https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/db/review.py#L319). The `list_review()` function is being called by `json()` function by supplying multiple arguments which are then used to formulate a query. Therefore, if we make a new `get_list_reviews(connection)` function, we cannot formulate query inside it without all the arguments of `list_reviews()` function.

paramsingh: Why not make a get_reviews(connection, *, inc_drafts...) and then call it directly in the json function and the list_reviews function

critiquebrainz/data/dump_manager.py
```diff
@@ -144,43 +144,48 @@ def json(location, rotate=False):
...
-
-            shutil.rmtree(temp_dir)  # Cleanup
-
-    if rotate:
-        print("Removing old sets of archives (except two latest)...")
-        remove_old_archives(location, "critiquebrainz-[0-9]+-[-\w]+-json.tar.bz2",
-                            is_dir=False, sort_key=os.path.getmtime)
-
-    print("Done!")
+    with db.engine.connect() as connection:
```
paramsingh: just use `with db.engine.begin() as connection` here.

critiquebrainz/db/review.py
```diff
@@ -316,33 +319,9 @@ def create(*, entity_id, entity_type, user_id, is_draft, text=None, rating=None,
...
-        inc_drafts (bool): True if reviews marked as drafts should be included, False if not.
-        inc_hidden (bool): True if reviews marked as hidden should be included, False if not.
-        exclude (list): List of reviews (their IDs) to exclude from results.
-
-    Returns:
-        Tuple with two values:
-        1. list of reviews as dictionaries,
-        2. total number of reviews that match the specified filters.
-    """
+def get_reviews_list(connection, *, inc_drafts=False, inc_hidden=False, entity_id=None,
```
paramsingh: you should add documentation here for why we're passing the `connection` here.

code-master5: @paramsingh Done!

code-master5:
@paramsingh Done!

PR #217, opened 2018-10-23
CB-311: Temporary pytest failure workaround

code-master5:
It's a temporary workaround for [CB-311](https://tickets.metabrainz.org/projects/CB/issues/CB-311).
Will be solved after the issue [here](https://github.com/PyCQA/pylint/issues/2563).


SPH73:
I got the following error when trying the workaround:

app.jinja_env().filters['filter_name'] = filtername
TypeError: 'Environment' object is not callable

PR #218, opened 2018-10-25
CB-298: Add transaction support for public dumps

code-master5:
Make public dumps in a single transaction.

### DIFF
--- a/critiquebrainz/data/dump_manager.py
+++ b/critiquebrainz/data/dump_manager.py
@@ -209,31 +209,35 @@ def public(location, rotate=False):
     meta_files_dir = tempfile.mkdtemp()
     prepare_meta_files(meta_files_dir, time_now=time_now)
 
-    # BASE ARCHIVE
-    # Contains all license independent data (licenses, users)
-    base_archive_path = create_base_archive(
-        location=dump_dir,
-        meta_files_dir=meta_files_dir,
-    )
-    print(base_archive_path)
-
-    # 1. COMBINED
-    # Archiving all reviews (any license)
-    review_dump_path = create_reviews_archive(
-        location=dump_dir,
-        meta_files_dir=meta_files_dir,
-    )
-    print(review_dump_path)
-
-    # 2. SEPARATE
-    # Creating separate archives for each license
-    for license in db_license.list_licenses():
-        review_dump_path = create_reviews_archive(
+    with db.engine.begin() as connection:
+        # BASE ARCHIVE
+        # Contains all license independent data (licenses, users)
+        base_archive_path = create_base_archive(
+            connection,
             location=dump_dir,
             meta_files_dir=meta_files_dir,
-            license_id=license['id'],
+        )
+        print(base_archive_path)
+
+        # 1. COMBINED
+        # Archiving all reviews (any license)
+        review_dump_path = create_reviews_archive(
+            connection,
+            location=dump_dir,
+            meta_files_dir=meta_files_dir,
         )
         print(review_dump_path)
+
+        # 2. SEPARATE
+        # Creating separate archives for each license
+        for license in db_license.get_licenses_list(connection):
+            review_dump_path = create_reviews_archive(
+                connection,
+                location=dump_dir,
+                meta_files_dir=meta_files_dir,
+                license_id=license['id'],
+            )
+            print(review_dump_path)
 
     shutil.rmtree(meta_files_dir)  # Cleanup
     if rotate:
@@ -269,10 +273,11 @@ def add_meta_files(tarfile, meta_files_dir):
     tarfile.add(os.path.join(meta_files_dir, 'SCHEMA_SEQUENCE'), arcname='SCHEMA_SEQUENCE')
 
 
-def create_base_archive(*, location, meta_files_dir=None):
+def create_base_archive(connection, *, location, meta_files_dir=None):
     """Creates a dump of all license-independent information: (users, license).
 
     Args:
+        connection (sqlalchemy.engine.Connection): an sqlalchemy connection to the database for executing database queries
         location: Path of the directory where the archive needs to be created.
         meta_files_dir (optional): Path of the directory containing the meta files to be copied
             into the archive (TIMESTAMP and SCHEMA_VERSION). If not specified, the meta files are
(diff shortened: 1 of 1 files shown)


critiquebrainz/data/dump_manager.py
```diff
@@ -364,21 +367,18 @@ def create_reviews_archive(*, location, meta_files_dir=None, license_id=None):
...
+            with open(os.path.join(reviews_tables_dir, 'review'), 'w') as f:
+                cursor.copy_to(f, REVIEW_SQL)
+
+            with open(os.path.join(reviews_tables_dir, 'revision'), 'w') as f:
+                cursor.copy_to(f, REVISION_SQL)
+
+            with open(os.path.join(reviews_tables_dir, 'avg_rating'), 'w') as f:
+                cursor.copy_to(f, "(SELECT {columns} FROM avg_rating)".format(columns=", ".join(_TABLES["avg_rating"])))
+        except Exception:
+            print("Error while copying tables during the creation of the reviews archive")
```
paramsingh: If you're catching the exception here, you'll have to rollback the transaction manually because otherwise it'll move out of the with block with no error and commit the transaction

Or just raise the exception back

critiquebrainz/data/dump_manager.py
```diff
@@ -269,10 +273,11 @@ def add_meta_files(tarfile, meta_files_dir):
     tarfile.add(os.path.join(meta_files_dir, 'SCHEMA_SEQUENCE'), arcname='SCHEMA_SEQUENCE')
 
 
-def create_base_archive(*, location, meta_files_dir=None):
+def create_base_archive(connection, *, location, meta_files_dir=None):
     """Creates a dump of all license-independent information: (users, license).
 
     Args:
+        connection (object): for supporting execution withing a transaction
```
paramsingh: What type of object? And there's a typo

critiquebrainz/data/dump_manager.py
```diff
@@ -288,17 +293,14 @@ def create_base_archive(*, location, meta_files_dir=None):
...
-                    print('Error while copying tables during creation of the base archive.')
-                    transaction.rollback()
+        cursor = connection.connection.cursor()
+        try:
+            with open(os.path.join(base_archive_tables_dir, 'user_sanitised'), 'w') as f:
+                cursor.copy_to(f, '"user"', columns=('id', 'created', 'display_name', 'musicbrainz_id'))
+            with open(os.path.join(base_archive_tables_dir, 'license'), 'w') as f:
+                cursor.copy_to(f, 'license', columns=_TABLES["license"])
+        except Exception:
+            print('Error while copying tables during creation of the base archive.')
```
paramsingh: The exception should be raised here too

critiquebrainz/data/dump_manager.py
code-master5: @paramsingh I've checked this manually. Whether you catch the exception or not, `__exit__()` function will still be called. Since here, the `__exit__()` function itself contains the `transaction.rollback()` we don't need to call it explicitly.

critiquebrainz/data/dump_manager.py
code-master5: @paramsingh How should I write the type of object? Is there any sample which you can point out?

critiquebrainz/data/dump_manager.py
paramsingh: Why would it call the `transaction.rollback` if no exception is raised and the exit is a normal exit? 



```
with db.engine.begin() as transaction:
     pass
```
would probably do a transaction.commit

```
with db.engine.begin() as transaction:
     raise Exception
```
should rollback

the code where we're catching the exception and just logging it is equivalent to the first case?

critiquebrainz/data/dump_manager.py
```diff
@@ -314,11 +316,12 @@ def create_base_archive(*, location, meta_files_dir=None):
         return " + %s/cbdump.tar.bz2" % location
 
 
-def create_reviews_archive(*, location, meta_files_dir=None, license_id=None):
+def create_reviews_archive(connection, *, location, meta_files_dir=None, license_id=None):
     """Creates a dump of reviews filtered on the given license_id, their revisions and
        the avg. rating tables.
 
     Args:
+        connection (object): for supporting execution within a transaction
```
paramsingh: everything is an object in python so the type hint here is not very useful, you should be more specific. :)

critiquebrainz/data/dump_manager.py
paramsingh: I think it'll be `sqlalchemy.engine.Connection` or `sqlalchemy.engine.Transaction` here

critiquebrainz/data/dump_manager.py
code-master5: @paramsingh Sorry, my bad. Now I get it. Let's just raise the exception in that case, rather that explicitly writing rollback. What you say?

paramsingh: Right, okay. We should just be careful that the Exception reaches the toplevel `with` block for sure.

critiquebrainz/data/dump_manager.py
code-master5: @paramsingh something like `object returned by db.engine.begin()` ?

critiquebrainz/data/dump_manager.py
code-master5: Decided to use nested transactions instead.

critiquebrainz/data/dump_manager.py
```diff
@@ -288,17 +293,16 @@ def create_base_archive(*, location, meta_files_dir=None):
...
+
+        cursor = connection.connection.cursor()
+        try:
+            with open(os.path.join(base_archive_tables_dir, 'user_sanitised'), 'w') as f:
+                cursor.copy_to(f, '"user"', columns=('id', 'created', 'display_name', 'musicbrainz_id'))
+            with open(os.path.join(base_archive_tables_dir, 'license'), 'w') as f:
+                cursor.copy_to(f, 'license', columns=_TABLES["license"])
+        except Exception as e:
+            print('Error "{}" occurred while copying tables during creation of the base archive!'.format(e))
+            raise Exception
```
paramsingh: if you do just `raise` here, it would raise the specific exception that was raised, which would be more helpful for debugging.

```suggestion
            raise 
```

critiquebrainz/data/dump_manager.py
```diff
@@ -364,22 +369,20 @@ def create_reviews_archive(*, location, meta_files_dir=None, license_id=None):
...
+                cursor.copy_to(f, REVIEW_SQL)
+
+            with open(os.path.join(reviews_tables_dir, 'revision'), 'w') as f:
+                cursor.copy_to(f, REVISION_SQL)
+
+            with open(os.path.join(reviews_tables_dir, 'avg_rating'), 'w') as f:
+                cursor.copy_to(f, "(SELECT {columns} FROM avg_rating)".format(columns=", ".join(_TABLES["avg_rating"])))
+        except Exception as e:
+            print("Error {} occurred while copying tables during the creation of the reviews archive!".format(e))
+            raise Exception
```
paramsingh:
```suggestion
            raise
```

code-master5: @paramsingh Done!

PR #220, opened 2018-10-28
CB-320: Show commenter's name over comment, instead of reviewer's name

code-master5:


paramsingh:
nice catch, could we add a test for this?

Maybe add a review, add a comment and then request the page and assert that the name of the commenter is present in the html served.

code-master5:
@paramsingh Added tests.

PR #221, opened 2018-10-30
CB-321: No empty comments

code-master5:
Changed SQL and Comment Form to reject empty comments.

### DIFF
--- a/admin/schema_changes/13.sql
+++ b/admin/schema_changes/13.sql
@@ -1,0 +1,8 @@
+BEGIN;
+
+ALTER TABLE "comment_revision"
+  ALTER COLUMN text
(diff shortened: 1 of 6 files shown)


code-master5:
@paramsingh Please review this.

admin/schema_changes/13.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
+
+ALTER TABLE "comment_revision"
```
paramsingh: Generally, we'd have to think about what we want to do with empty comments already in the db before we add a constraint, but I checked and there aren't any empty comments (or any comments whatsoever) in the db so we should be okay.

PR #223, opened 2018-11-03
CB-299: Wait for consul config

code-master5:
Added code in webserver init file to wait for consul config file.


paramsingh:
Please don't force push into a single commit, it makes testing PRs locally harder, just add new commits and if something needs to be rebased or merged, I'll do it while merging.

PR #224, opened 2018-11-13
CB-323: Display error if unable to report incorrect spotify mapping

code-master5:


code-master5:
@paramsingh I've changed the mbspotify requests to use `params` as we had discussed. Check now.

PR #244, opened 2019-02-01
CB-326: Hide and Unhide button for moderators on entity review page

code-master5:
* If a review is not hidden, `Hide this review` button is shown to the moderator. Similarly, if the review is hidden, `Unhide this review` button is shown.
* A user cannot hide or unhide their review even if the user is a moderator. In this case, only some other moderator can take these actions (and is shown hide/unhide button).

### DIFF
--- a/admin/schema_changes/14.sql
+++ b/admin/schema_changes/14.sql
@@ -11,4 +11,3 @@ ALTER TABLE review ADD CONSTRAINT published_on_null_for_drafts_and_not_null_for_
     CHECK ((is_draft = 't' AND published_on IS NULL) OR (is_draft = 'f' AND published_on IS NOT NULL));
 
 COMMIT;
-
--- a/critiquebrainz/frontend/templates/review/entity/base.html
+++ b/critiquebrainz/frontend/templates/review/entity/base.html
@@ -45,7 +45,7 @@
         {% endif %}
       {% endif %}
 
-      {% if current_user.is_authenticated and current_user != review.user and current_user.is_admin() %}
+      {% if current_user.is_authenticated and current_user.is_admin() %}
         <hr />
         <div>
           <a href="{{ url_for('review.unhide' if review.is_hidden else 'review.hide', id=review.id) }}"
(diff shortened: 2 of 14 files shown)


critiquebrainz/frontend/templates/review/entity/base.html
```diff
@@ -45,6 +45,25 @@ <h3>{{ review_credit(review, user_picture_size=24) }}</h3>
         {% endif %}
       {% endif %}
 
+      {% if current_user.is_authenticated and current_user != review.user and current_user.is_admin() %}
```
paramsingh: admins should be able to hide their own reviews as well. It won't be needed in real life, but it'd make testing easier.

code-master5: @paramsingh Ok.

code-master5:
@paramsingh I've added some significant changes to this PR. Review now!

paramsingh:
>I've added some significant changes to this PR. 

A description of what the changes are is very helpful.

code-master5:
@paramsingh sorry for not describing the changes. Changes are as follows -

* DB changes - added `unhide_review` and `unblock_user` into `action_types` enum.
* Defined `AdminActions` enum (python 3), instead of explicit variables in `db/moderation_log.py` module. Essentially, a python based enum is used to represent DB enum.
* Updated tests for `moderation_log` module.
* `review/unhide` and `user/unblock` endpoints now accept POST request to carry out DB
changes. This is done to keep the codebase in conformance with REST principles.
* Edited `review` and `user` templates to remove code redundancy. Also added cases for handling
`review unhide` and `user unblock` actions.
* Wrote test cases for `user` view almost from scratch, in order to make it more like other tests. Added tests for handling `user unblock` case.

admin/schema_changes/14.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TYPE action_types ADD VALUE 'unhide_review' AFTER 'hide_review';
```
paramsingh: This should be inside a transaction.

PR #245, opened 2019-02-08
CB-342: Show comment stats in user profile

code-master5:


critiquebrainz/db/users.py
```diff
@@ -532,6 +532,51 @@ def get_reviews(user_id, from_date='1-1-1970'):
     return [dict(row) for row in rows]
 
 
+def get_comments(user_id, from_date='1-1-1970'):
+    """Get comments on reviews by a user from a specified time.
+
+    Args:
+        user_id(uuid): ID of the user.
+        from_date(datetime): Date from which comments on reviews submitted by user are to be returned.
```
paramsingh: the docstring says this is a datetime object, but you pass a string as the default.

code-master5:
@paramsingh Done. Review now! I also rebased this with current master.

critiquebrainz/db/users.py
```diff
@@ -532,6 +532,51 @@ def get_reviews(user_id, from_date='1-1-1970'):
     return [dict(row) for row in rows]
 
 
+def get_comments(user_id, from_date='1-1-1970'):
+    """Get comments on reviews by a user from a specified time.
+
+    Args:
+        user_id(uuid): ID of the user.
+        from_date(datetime/str): Date from which comments on reviews submitted by user are to be returned.
```
paramsingh: i would prefer it if the type of this was singular, either datetime or string. Having variables of multiple types does not lead to good things.

paramsingh: Sorry for being unclear on this.

code-master5: Since we're finally using these parameters into SQL queries, `datetime` object gets essentially converted to `str(datetime)`. Therefore, it should not be of much concern IMO. But still, I can change defaults to `datetime` type for more consistency.

paramsingh: I would prefer it to be datetime by default.

PR #247, opened 2019-02-17
CB-344: Make RG review header use join phrases

reosarevok:
https://tickets.metabrainz.org/browse/CB-344

### DIFF
--- a/critiquebrainz/frontend/templates/review/entity/release_group.html
+++ b/critiquebrainz/frontend/templates/review/entity/release_group.html
@@ -20,7 +20,10 @@
       {% set artist = [] %}
       {% for credit in release_group['artist-credit'] %}
         {% if credit.artist %}
-          {% do artist.append('<a href="%s">'|safe % url_for('artist.entity', mbid=credit.artist.id) + credit.artist.name + '</a>'|safe) %}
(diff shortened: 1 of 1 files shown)


critiquebrainz/frontend/templates/review/entity/release_group.html
```diff
@@ -20,7 +20,10 @@ <h2 id="title">
       {% set artist = [] %}
       {% for credit in release_group['artist-credit'] %}
         {% if credit.artist %}
-          {% do artist.append('<a href="%s">'|safe % url_for('artist.entity', mbid=credit.artist.id) + credit.artist.name + '</a>'|safe) %}
+          {% do artist.append('<a href="%s">'|safe % url_for('artist.entity', mbid=credit.artist.id) ~ credit.artist.name ~ '</a>'|safe) %}
```
reosarevok: As far as I can tell ~ is preferred to + (and this makes it consistent with the RG page header)

reosarevok:
This and https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/frontend/templates/release_group/entity.html#L12 should probably be turned into a function and shared rather than entered twice, but I'm not sure how Jinja uses functions, so for now I just made them work in the same way.

paramsingh:
@brainzbot add to whitelist

PR #251, opened 2019-03-05
CB-305: Allow users to edit comments

code-master5:
Changes/additions summary:
* Separate template for comments
* Added JavaScript for dynamically displaying comment edit form when a user clicks on the comment edit button.
* Added `comment.edit` view

### DIFF
--- a/critiquebrainz/frontend/static/styles/main.less
+++ b/critiquebrainz/frontend/static/styles/main.less
@@ -418,6 +418,11 @@ a#edit-review { margin-top: 20px; }
       }
     }
   }
+
+  // comments
+  .comment-header {
+    margin: 15px auto;
+  }
 }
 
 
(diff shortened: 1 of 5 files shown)


paramsingh:
i wonder if we should wait until we integrate react and then do this over with that. This JS is not really ideal imo?

Opinions?

code-master5:
@paramsingh You're right that the `js` is pretty messy here.

But the problem is that the `comment` section is actually a part of the `reviews` page. It would take us time to get there if we're trying to convert pages to support react. We should not wait till then to add this feature IMO. We should instead delay comments being moderatable. That's why I didn't send PR for that.

What do you think?

code-master5:
@paramsingh @ferbncode please find some time to review and merge this (or suggest changes if any).

code-master5:
@ferbncode @paramsingh Any reasons for not reviewing or merging this PR?

code-master5:
comment-edit

@paramsingh  Works fine for me! Still, I've added a line break between username and comment update form. Are you sure that you [build the static files](https://critiquebrainz.readthedocs.io/intro.html#building-static-files)?

paramsingh:
whoops

PR #253, opened 2019-03-18
CB-349: Gulp to webpack

code-master5:
Summary of changes:

* Updated `package.json` with packages to support Webpack, replaced `npm-shrinkwrap.json` with `package-lock.json`.
* Removed `gulp` and added `webpack`. Added plugins to convert `less` to `css`, clean build directory and generate `manifest` file.
* Updated `static_manager.py` to read from newer `manifest` file.
* Updated `.babelrc` with newer babel env.
* Updated `node` to version `10.x` in Dockerfile. Also updated command to build scripts and styles.


alastair:
> Do you know what the reason is not to compile the static assets on build in dev environment
We can't do this during development because we mount the working directory into the web container as a volume, and this overrides any commands that we run during build. This is the same way that we do static assets for build in AB

MonkeyDo:
OK, thanks @alastair . Personally I don't mind running npm commands on my machine.

webpack.config.js
```diff
@@ -0,0 +1,63 @@
...
+    main: path.resolve(__dirname, 'critiquebrainz/frontend/static/scripts/main.js'),
+    common: path.resolve(__dirname, 'critiquebrainz/frontend/static/scripts/common.js'),
+    leaflet: path.resolve(__dirname, 'critiquebrainz/frontend/static/scripts/leaflet.js'),
+    rating: path.resolve(__dirname, 'critiquebrainz/frontend/static/scripts/rating.js'),
+    spotify: path.resolve(__dirname, 'critiquebrainz/frontend/static/scripts/spotify.js'),
+    "wysiwyg-editor": path.resolve(__dirname, 'critiquebrainz/frontend/static/scripts/wysiwyg-editor.js'),
+  },
+  output: {
+    path: path.resolve(__dirname, 'critiquebrainz/frontend/static/build/'),
+    filename: '[name].[hash].js',
```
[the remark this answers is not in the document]
code-master5: @MonkeyDo Thanks for pointing this out.

code-master5:
@MonkeyDo Thanks for suggesting improvements. I've made the changes.

MonkeyDo:
Look good, nice job !

paramsingh:
In LB, We do live compiling by adding a seperate container that has webpack build --watch running inside and the static assets are moved out of the volume. I'd like a similar setup here.

code-master5:
@paramsingh please review now! I have added the changes.

webpack.config.js
```diff
@@ -0,0 +1,69 @@
...
+    entry: {
+      'main': path.resolve(frontEndStaticDir, 'scripts/main.js'),
+      'common': path.resolve(frontEndStaticDir, 'scripts/common.js'),
+      'leaflet': path.resolve(frontEndStaticDir, 'scripts/leaflet.js'),
+      'rating': path.resolve(frontEndStaticDir, 'scripts/rating.js'),
+      'spotify': path.resolve(frontEndStaticDir, 'scripts/spotify.js'),
+      'wysiwyg-editor': path.resolve(frontEndStaticDir, 'scripts/wysiwyg-editor.js'),
+    },
+    output: {
+      path: path.resolve(__dirname, 'critiquebrainz/frontend/static/build/'),
```
alastair: could this be path.resolve(frontEndstaticDir, build) in order to not duplicate path?

Dockerfile.webpack
```diff
@@ -0,0 +1,8 @@
+FROM node:10.15-alpine
```
alastair: Does the builder have to be a separate docker image? Typically we try to share our images as much as possible. The Dockerfile for CB also performs `npm install`, so as far as I can tell we should be able to build the static files using a container based on that image

docker/docker-compose.dev.yml
```diff
@@ -50,3 +51,12 @@ services:
...
       - "5430:5432"
+
+  static_builder:
+    build:
+      context: ..
+      dockerfile: Dockerfile.webpack
+    command: npm run dev
+    volumes:
+      - ..:/code:z
+      - ../critiquebrainz/frontend/static:/static:z
```
alastair: why is this needed? I don't see a reference to `/static` in the webpack configuration file, so I don't think it has any effect

admin/compile_resources.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
```
alastair: I don't see this file used anywhere in this repository. Is it needed?

Dockerfile.webpack
code-master5: @alastair Well, I just tried to take inspiration from the new setup in LB. :sweat_smile: @paramsingh @MonkeyDo what do you guys think about this?

docker/docker-compose.dev.yml
code-master5: Done!

webpack.config.js
code-master5: Done!

Dockerfile.webpack
MonkeyDo: alastair is absolutely right! If the BD docker image can run npm commands, then there's no need to use another image.

For LB, there was previously nothing using node so the image did not contain node.

MonkeyDo:
@code-master5 : After discussing it with alastair, removing CleanWebpackPlugin and running the asset building once before starting the services does the trick (at least until a better solution is found). I've added some commits to document that.

code-master5:
@MonkeyDo @alastair Thanks for your guidance! I'm sorry I could not follow up more actively because of some personal stuff. :sweat_smile:

code-master5:
@paramsingh @ferbncode please find some time to review and merge this (or suggest changes if any).

code-master5:
@ferbncode @paramsingh Any reasons for not reviewing or merging this PR?

alastair:
Thanks, merged into master!

code-master5:
Thanks @alastair ! :bow:

PR #255, opened 2019-04-23
CB-352: Fix sorting in review blueprint's browse route

spellew:


spellew:
Updated commit to fix similar issue in WS docstring and valid values list

PR #259, opened 2019-05-24
CB-353: ORM query for getting tags on an entity fails after schema change in the musicbrainz_db

spellew:


spellew:
Updated `mbdata` and added a test related to the issue.

spellew:
Fixed pep8 and pylint issues.

PR #261, opened 2019-06-09
CB-270: Implemented the reviewal of additional entities (works, recordings, labels, artist) [WIP]

spellew:


ferbncode:
@spellew I understand that the pull request contains code for all the entities: work, recordings, labels, etc and using BU in CB? Could you maybe separate the changes into different pull requests? According to your (modified) timeline, we're to attend one feature at a time. But, since you've already opened pull request for all entities, I'd say that just split this pull request into multiple ones:

1. Using BU in CB for existing entities
2. Support for labels
3. Support for works
4. Support for recordings
5. Support for artist 

I suggest to do that as changes might differ for different entities, especially on the front-end side.

This way we can progress on tasks one by one by working on one of your pull requests and do releases as we progress in the GSoC coding period. If we try to release everything at once, some things might not work and it makes that very hard to see any outcome. Please split the pull requests.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/metabrainz/brainzutils-python.git@v1.10.1
+git+git://github.com/spellew/brainzutils-python.git@additional-entity-support#egg=brainzutils
```
ferbncode: @spellew There was a release made after we merged your pull request. You can use release version: `v1.12.0` for BU for these changes.

PR #263, opened 2019-06-16
CB-270: Implemented the reviewal of artists

spellew:


ferbncode:
Let's start this as well. Could you fix the conflicts?

spellew:
Merge conflict resolved.

ferbncode:
This is already good. If we could support coverart, it would be awesome, but it seems that the coverart API supports only release and release-groups at the moment. 

However, do you think you can add a tab or list of existing reviews for an artist. 
image
For example, for a release-group, a user can see the existing reviews as well. 
image. You can come up with different ideas as to how to structure the page, and we can discuss and decide on one :)

spellew:
@ferbncode So something like this?

Screenshot_2019-07-26 Fugees - CritiqueBrainz(2)
Screenshot_2019-07-26 Fugees - CritiqueBrainz(1)
Screenshot_2019-07-26 Fugees - CritiqueBrainz

ferbncode:
@spellew The third one looks good. We can limit to 5 reviews to show on artist page if there are many reviews. But the third one looks good, being outside the discography and very similar to release group pages.. What do you think?

spellew:
@ferbncode I also like the third one, but I'm not sure if users having to scroll down to view the discography is a problem or not? If we limit it to five, we'd have arrows on the left or right, allowing users to view the next or previous reviews?

ferbncode:
@spellew I think there could be a link "See all reviews for"Artist"" which would list all the remaining reviews on the same page? If we find its not what users expect, we can come back again to fix it.

spellew:
@ferbncode That sounds good to me.

spellew:
@ferbncode So something like this?

http://localhost/artist/e8414012-4a1c-4ad4-be5e-fc55294e28cc
Screenshot_2019-07-27 Lauryn Hill - CritiqueBrainz

http://localhost/artist/e8414012-4a1c-4ad4-be5e-fc55294e28cc?release_type=album&reviews=all
Screenshot_2019-07-27 Lauryn Hill - CritiqueBrainz(1)

ferbncode:
@spellew That looks good IMO :)

spellew:
> @spellew That looks good IMO :)

Changes have been pushed.

critiquebrainz/frontend/views/artist.py
```diff
@@ -25,29 +25,42 @@ def entity(mbid):
     # Note that some artists might not have a list of members because they are not a band
     band_members = _get_band_members(artist)
 
+    reviews_limit = 5
```
ferbncode: Let's make it configurable as a constant at the beginning of the file (and `release_group_limit` as well) :)

spellew: Can I get an example of what this would look like? I've attempted placing the constants at the beginning of the file, under the imports, but I keep getting `UnboundLocalError: local variable referenced before assignment`.

ferbncode: You can place then in `__init__.py` and then import the variables?

spellew: Updated.

ferbncode:
@paramsingh When you get time, could you merge this and deploy to beta.critiquebrainz.org? Thanks in advance! :)

paramsingh:
I've deployed it on beta, thanks for all the great work!

PR #264, opened 2019-06-16
CB-359: Implemented the reviewal of labels

spellew:


ferbncode:
@spellew When you have time, please update this PR to fix conflicts too. Thanks! :)

spellew:
@ferbncode Resolved.

critiquebrainz/frontend/views/label.py
```diff
@@ -0,0 +1,42 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
```
ferbncode: I don't think we need a license here.

critiquebrainz/frontend/views/label.py
```diff
@@ -0,0 +1,42 @@
...
+from flask import Blueprint, render_template
+# from flask import request
+# from flask_login import current_user
+from werkzeug.exceptions import NotFound
+from flask_babel import gettext
+import critiquebrainz.frontend.external.musicbrainz_db.label as mb_label
+import critiquebrainz.frontend.external.musicbrainz_db.exceptions as mb_exceptions
+# import critiquebrainz.db.review as db_review
+# from critiquebrainz.frontend.forms.rate import RatingEditForm
+# from critiquebrainz.frontend.views import get_avg_rating
```
ferbncode: Let's remove the commented imports

critiquebrainz/frontend/views/label.py
ferbncode: (Also for other remaining pull requests)

ferbncode:
The pull request is good, but let's try to
1. Fix the conflicts
2. Show some data on the label page for the Label. (Atleast the name, maybe the annotations and the releases too). See, for example: https://musicbrainz.org/label/5899271b-a77a-4081-93a9-0bd05b9b4180

spellew:
@ferbncode Updated the entity page to display some data.

ferbncode:
@spellew Right now, the information is okay, One thing I noticed for artists and labels is missing rating which is also part of the proposal. Please update this PR for including ratings. I'll add review for the code later today :+1:

critiquebrainz/frontend/external/musicbrainz_db/label.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    key = cache.gen_key(mbid)
+    label = cache.get(key)
+    if not label:
+        label = fetch_multiple_labels(
+            [mbid],
+            includes=['artist-rels', 'url-rels'],
+        ).get(mbid)
+        cache.set(key=key, val=label, time=DEFAULT_CACHE_EXPIRATION)
+    return label_rel.process(label)
```
ferbncode: Do you have the complete musicbrainz_db imported in your local setup or just the sample dumps? (I am curious if the same `process` works for labels? We will find out when this is deployed to beta, but please keep in mind as we have potential duplicated code we can reduce :)

critiquebrainz/frontend/views/label.py
```diff
@@ -0,0 +1,60 @@
...
+    if request.args.get('reviews') == "all":
+        label_reviews_limit = None
+
+    if current_user.is_authenticated:
+        my_reviews, my_count = db_review.list_reviews(
+            entity_id=label['id'],
+            entity_type='label',
+            user_id=current_user.id,
+        )
+        my_review = my_reviews[0] if my_count else None
```
ferbncode: What happens if I have a review already for a label. It should be passed to the `render_template` to display the option of `Edit Review` rather than `Write Review`. See `place.py` or `release_group.py` for example.

ferbncode: Also, other important variable used in the template like `current_user`.

critiquebrainz/frontend/external/musicbrainz_db/label.py
ferbncode: Actually, let's try to remove duplication of this code in a separate PR in some way. (I'll create a ticket for this)

critiquebrainz/frontend/templates/macros.html
```diff
@@ -53,6 +53,13 @@
         {{ _('Artist') }}
       </span>
     {% endif %}
+  {% elif entity_type == 'label'  %}
+    <img src="{{ get_static_path('images/placeholder_place.svg') }}" {{ attributes }} />
```
ferbncode: `placeholder_disc.svg` would be much relating to labels, imho.

spellew:
@ferbncode I was planning on showing the releases, but I'm worried about time constraints, and if that can be achieved while we're working on other entities.

critiquebrainz/frontend/views/label.py
spellew: All the other important variables too? I think I only passed the ones that we're currently using, to the render template.

ferbncode:
@spellew Yes, that is my plan to do it after all the entities as an enhancement (Right now, I'll just create a ticket before merging this)

critiquebrainz/frontend/views/label.py
ferbncode: See the `base.py` (the variable `current_user` and `my_review` do play a role there.

critiquebrainz/frontend/external/musicbrainz_db/label.py
spellew: I was looking at this and I think it can be simplified. I believe I manually downloaded and imported the dumps.

ferbncode: :+1:

critiquebrainz/frontend/templates/macros.html
spellew: Updated.

critiquebrainz/frontend/views/label.py
spellew: I've added current_user and my_review those seem to be the only variables used outside the label template.

paramsingh:
Did you create a ticket already, @ferbncode ?

ferbncode:
@paramsingh I'll make sure to create it today.

admin/schema_changes/16.sql
```diff
@@ -1 +1,2 @@
 ALTER TYPE entity_types ADD VALUE 'artist' AFTER 'place';
+ALTER TYPE entity_types ADD VALUE 'label' AFTER 'artist';
```
paramsingh: I just saw this, this should be in a different script so I can run it on the main db. Generally, each PR should have a single schema change script.

ferbncode: Ah, right. @spellew Please add schema change scripts to `admin/schema_changes` (One for `artists` and one for `labels`)

ferbncode: Thank you @paramsingh for pointing that out :)

ferbncode: And for fixing it too.

PR #269, opened 2019-07-14
CB-356: Implemented handling of deleted entities

spellew:


critiquebrainz/frontend/external/musicbrainz_db/utils.py
```diff
@@ -45,44 +22,27 @@ def unknown_entity(entity_gid, entity_type):
...
 
-def get_entities_by_gids(*, query, entity_type, mbids):
-    """Get entities using their MBIDs.
-
-    An entity can have multiple MBIDs. This function may be passed another
-    MBID of an entity, in which case, it is redirected to the original entity.
-
-    Note that the query may be modified before passing it to this
-    function in order to save queries made to the database.
+def deleted_entities_to_unknown(*, entities, entity_type, mbids):
```
ferbncode: Let's call it `map_deleted_mb_entities_to_unknown`

critiquebrainz/frontend/external/musicbrainz_db/artist.py
```diff
@@ -15,9 +16,14 @@ def get_artist_by_id(mbid):
     key = cache.gen_key(mbid)
     artist = cache.get(key)
     if not artist:
-        artist = fetch_multiple_artists(
+        multiple_artists = fetch_multiple_artists(
             [mbid],
             includes=['artist-rels', 'url-rels'],
+        )
+        artist = deleted_entities_to_unknown(
```
ferbncode: We should do this, only if the returned artist in the call to `fetch_multiple_artists` returns empty dictionary. That makes the code much easy to read and intuitive.

critiquebrainz/frontend/external/musicbrainz_db/entities.py
```diff
@@ -25,15 +26,23 @@ def get_multiple_entities(entities):
     release_group_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'release_group', entities)]
     place_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'place', entities)]
     event_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'event', entities)]
-    entities_info.update(fetch_multiple_release_groups(
-        release_group_mbids,
-        includes=['artists'],
+    multiple_release_groups = fetch_multiple_release_groups(release_group_mbids, includes=['artists'])
+    multiple_places = fetch_multiple_places(place_mbids)
+    multiple_events = fetch_multiple_events(event_mbids)
+    entities_info.update(map_deleted_mb_entities_to_unknown(
```
ferbncode: This can be in a loop using python zip, what do you think?

ferbncode: Scratch that. It would not be very readable.

critiquebrainz/frontend/external/musicbrainz_db/artist.py
```diff
@@ -15,9 +16,17 @@ def get_artist_by_id(mbid):
     key = cache.gen_key(mbid)
     artist = cache.get(key)
     if not artist:
-        artist = fetch_multiple_artists(
+        multiple_artists = fetch_multiple_artists(
```
ferbncode: One thing that bothers me is the name `multiple_artists` or `multiple_*` since we pass only a single mbid, I'd recommend  having a function `fetch_multiple_artists` (critiquebrainz version), that in turns calls the bu version and returns unknown if not found. (Similar for others)

critiquebrainz/frontend/external/musicbrainz_db/entities.py
ferbncode: This wouldn't be very readable, ignore previous comment.

critiquebrainz/frontend/external/musicbrainz_db/artist.py
ferbncode: That would avoid code duplication, for example, in `entities.py`

spellew: Updated code and fixed a bug I found in utils.py

PR #289, opened 2020-01-05
Upgrade to python 3.8

ferbncode:
Upgrade python version to 3.8. I realized that the CB pylint and flake8 tests show some warnings related to python 3.8 changes. I'll create a ticket for fixing them by updating with new `pylint` version (not released yet imo).


ferbncode:
The latest version of `flake8` doesn't [support](https://bbs.archlinux.org/viewtopic.php?id=251088) `Python3.8`, so we'll have to wait before merging this :sweat:

ferbncode:
@brainzbot retest this please

ferbncode:
@brainzbot retest this please

alastair:
I've update this PR to rebase it against the webpack update, and to include @mayhem's patch to upgrade the dev environment to postgres 12. 
I also updated some places where the py 3.8 and postgres 12 upgrade were missed, and improved the way that we use the postgres image in order to remove some unnecessary docker images and config files.
I improved the docker setup by adding `--no-cache-dir` to the pip installation.

@ferbncode what's the status of flake8 with python 3.8? If it's working now then I'm happy to merge this unless anyone has some other feedback on my changes.

ferbncode:
@alastair I updated `flake8` to the correct version a few days back and fixed the errors/warnings reported by it. It should be working as expected. 

Thanks for updating the PR and adding changes regarding upgrading development environment (and docker changes). :) 
There is one test failing, I'll try to fix it. I believe we can merge this after fixing the failing test. (with the following error):
```
/bin/sh: 1: pg_dump: not found
```

alastair:
whoops, that missing pg_dump will be due to me removing the postgres clients from some containers. We can add it back in

PR #293, opened 2020-07-31
Setup updates

alastair:
When following the setup instructions some things didn't work

* Postgres docker image now doesn't do trust auth by default, set the env variable
* New musicbrainz db test image didn't include bzip2 (https://github.com/metabrainz/musicbrainz-server/pull/1621)

Also add `develop.sh` to make more consistent with other brainzes, and update installation documentation to include it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,11 @@ CritiqueBrainz is a repository for Creative Commons licensed reviews.
 
 ## Documentation
 
-**Documentation is available at https://critiquebrainz.readthedocs.org/.**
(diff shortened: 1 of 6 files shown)


alastair:
@brainzbot add to whitelist

alastair:
I rebased this to add the javascript and python 3.8 updates, and added a test.sh (CB-281)

alastair:
We should add some notes to the developers documentation about how to use test.sh

PR #298, opened 2020-09-08
CB-373: "Write a review" link in footer doesn't work

amCap1712:
Do not raise BadRequest if both entity_id and entity_type are missing

### DIFF
--- a/critiquebrainz/frontend/templates/event/entity.html
+++ b/critiquebrainz/frontend/templates/event/entity.html
@@ -15,7 +15,7 @@
     </h2>
 
     {% if not my_review %}
-      <a id="write-review" href="{{ url_for('review.create', place=id) }}"
+      <a id="write-review" href="{{ url_for('review.create', entity_type='event', entity_id=id) }}"
          role="button" class="btn btn-success pull-right">
         <span class="glyphicon glyphicon-asterisk"></span> {{ _('Write a review') }}
       </a>
--- a/critiquebrainz/frontend/views/review.py
+++ b/critiquebrainz/frontend/views/review.py
@@ -201,20 +201,16 @@ def revisions_more(id):
     return jsonify(results=template, more=(count - offset - RESULTS_LIMIT) > 0)
 
 
-# TODO(psolanki): Refactor this function to remove PyLint warning.
-# pylint: disable=too-many-branches
-@review_bp.route('/write', methods=('GET', 'POST'))
+@review_bp.route('/write/<entity_type>/<entity_id>/', methods=('GET', 'POST'))
+@review_bp.route('/write/')
 @login_required
-def create():
-    entity_type = None
-    for supported_type in ENTITY_TYPES:
-        if entity_id := request.args.get(supported_type):
-            entity_type = supported_type
-            break
-
-    if not entity_id:
+def create(entity_type=None, entity_id=None):
+    if not (entity_id or entity_type):
         flash.info(gettext("Please choose an entity to review."))
         return redirect(url_for('search.selector', next=url_for('.create')))
+
+    if entity_type not in ENTITY_TYPES:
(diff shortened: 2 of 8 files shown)


critiquebrainz/frontend/views/review.py
```diff
@@ -212,10 +212,6 @@ def create():
             entity_type = supported_type
             break
 
-    if not (entity_id or entity_type):
```
alastair, line 215: I think that this check is useful, because in CB you can't review all types of entities in MB.
I'd like to keep this check, but perhaps we could improve the way that we format the URL, e.g. instead of /write?artist=id, what if we had `/write/artist/id`? This would make it easier to get the variable for the entity type, and would fix a bug that exists where you can add 2 different entity types as parameters!

We could do something like:
```py
@route('/write/<entity_type>/<entity_id>/)
def create(entity_type, entity_id):
    if entity_type not in ENTITY_TYPES:
        raise BadRequest
    if not entity_id:
        redirect
    ... other checks
```

In this case I'd also like to see tests for these checks, and we should also have a redirect endpoint that converts old-style urls into new-style

alastair (changes requested):

critiquebrainz/frontend/templates/event/entity.html
```diff
@@ -15,7 +15,7 @@ <h2 class="pull-left">
     </h2>
 
     {% if not my_review %}
-      <a id="write-review" href="{{ url_for('review.create', place=id) }}"
+      <a id="write-review" href="{{ url_for('review.create', entity_type='event', entity_id=id) }}"
```
alastair, line 18: whoops, good catch here. It seems that this has happened in a number of places (e.g. #297). Would you be able to have another detailed look and ensure that it isn't incorrect in any other place?

critiquebrainz/frontend/views/review.py
```diff
@@ -201,25 +201,17 @@ def revisions_more(id):
...
     if not (entity_id or entity_type):
-        logging.warning("Unsupported entity type")
-        raise BadRequest("Unsupported entity type")
-
-    if not entity_id:
         flash.info(gettext("Please choose an entity to review."))
         return redirect(url_for('search.selector', next=url_for('.create')))
 
+    if entity_type not in ENTITY_TYPES:
+        raise BadRequest("Unsupported entity type")
```
alastair, line 213: Can we make this error nicer? Something like "You can't write reviews about this type of entity"

alastair:
This looks better, thanks.

I also made two additional suggestions - 
> In this case I'd also like to see tests for these checks, and we should also have a redirect endpoint that converts old-style urls into new-style

I had a look at the tests in test_review and I couldn't find any tests that check these situations. It's also important that we have a permanent redirect set up for old-style urls to new urls (and tests for those). Don't worry about checking if an entity exists with the old-style urls, just redirect all `?artist=id` urls to `/artist/id` and do the check in the new-style URL

amCap1712:
I have added the redirection for query urls to the new scheme and a test for the same. Can you tell which checks are missing the tests ? I have purposely not added checks for when the user is blocked and when entity is not present in the db because they require changes out of the scope of this PR.

amCap1712:
I have added the commit to fix CB-377 here. I checked the codebase and found that this is the only unhandled instance (there is one more instance for a release endpoint as well which I'll explore but others are handled). 

Initially, I was confused into thinking that this problem not being handled correctly because of an issue with Flask debug toolbar. The issue was a bit weird and interesting. It gives an Internal Server Error if a `NoDataFOundException` is thrown even if there is `except` clause to handle it. However, if it works perfectly if something like `except BaseException` is used. But if I print the type of the exception caught in the `BaseException` except block, it is `NoDataFoundException` indeed.

alastair:
🥳  thanks!

PR #300, opened 2020-09-12
CB-361: Group member list shows None if relationship has end date and no start date

amCap1712:
Do not display None if start date not present

### DIFF
--- a/critiquebrainz/frontend/templates/artist/entity.html
+++ b/critiquebrainz/frontend/templates/artist/entity.html
@@ -10,17 +10,21 @@
         {{ member.name }}
       </a>
       {% if member.periods %}
-        {% if not member.ended and member.periods and member.periods[0][0] %}
-          <span class="text-muted">
-            ({{ 'since ' + member.periods[0][0] }})
-          </span>
-        {% else %}
           <span class="text-muted">
           ({% for period in member.periods -%}
-            {{ period[0] + '-' + period[1] + ', ' * (not loop.last) }}
+              {% if not loop.last %}
+                {{ period[0] + '-' + period[1] + ', '}}
+              {% else %}
+                {% if not member.ended and period[0] %}
+                    {{ 'since ' + period[0] }}
+                {% elif not period[0] and period[1] %}
+                    {{ 'until ' + period[1] }}
+                {% else %}
+                    {{ period[0] + '-' + period[1] }}
+                {% endif %}
+              {% endif %}
           {%- endfor %})
           </span>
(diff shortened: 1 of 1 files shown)


critiquebrainz/frontend/templates/artist/entity.html
```diff
@@ -48,7 +52,7 @@ <h2 class="pull-left">
   </h2>
 
   {% if not my_review %}
-    <a id="write-review" href="{{ url_for('review.create', artist=id) }}"
+    <a id="write-review" href="{{ url_for('review.create', entity_type='artist', entity_id=id) }}"
```
alastair, line 55: this change seems unrelated to this bug (is it from #298?). Please make sure that you don't include unrelated changes in commits.

critiquebrainz/frontend/templates/artist/entity.html
```diff
@@ -10,17 +10,21 @@
...
       {% if member.periods %}
-        {% if not member.ended and member.periods and member.periods[0][0] %}
-          <span class="text-muted">
-            ({{ 'since ' + member.periods[0][0] }})
-          </span>
-        {% else %}
           <span class="text-muted">
           ({% for period in member.periods -%}
-            {{ period[0] + '-' + period[1] + ', ' * (not loop.last) }}
+              {% if not loop.last %}
```
alastair, line 15: can you include a screenshot of each of these examples? (time period with _since_, time period with _until_, something with more than 1 period, and something with more than one period, and one of the since/until examples).
What would be great is if we could construct a `member.periods` list and render this section, checking the resulting html. I'm not sure if there is a method in jinja or flask-test to check the contents of an html template while ignoring additional whitespace. This could be interesting to look into.

critiquebrainz/frontend/templates/artist/entity.html
amCap1712: Yeah right. I'll take care of it from now on.

critiquebrainz/frontend/templates/artist/entity.html
amCap1712: I have attached a screenshot. I couldn't get one for the until because I only have the sample dump. I'll try to get a full dump and try again but it'll take some time. Also, I could not understand the rest of your comment specifically this part
>  if we could construct a member.periods list and render this section, checking the resulting html.
Do you mean that this processing should be done in the python file instead of the template or something else ?
Screenshot from 2020-09-17 02-26-42

PR #301, opened 2020-09-14
CB-374: Entity selection tabs don't change when clicked

amCap1712:
The current UI components need to rewritten to work correctly with Bootstrap 4. Hence, using Bootstrap 3 till then.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "pre-dev": "webpack --mode development"
   },
   "dependencies": {
-    "bootstrap": "4.3.1",
+    "bootstrap": "^3.4.1",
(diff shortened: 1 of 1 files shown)


alastair:
The latest bootstrap 3 release is 3.4.1. Why did you downgrade to 3.3?

amCap1712:
I'll test with 3.4.1 again. I had changed it to 3.3.7 because that is what the production was on.

alastair:
The git history of package.json shows that it was upgraded explicitly for a security upgrade, so it's not a good idea to stay with the current production version. The latest 3.x should be compatible, so I see no reason why it won't work.

alastair:
Thanks!

PR #302, opened 2020-09-18
CB-368: Submitting a review, clicking back, and submitting again causes error

amCap1712:
Redirect user to edit page if draft review exists

### DIFF
--- a/critiquebrainz/frontend/views/review.py
+++ b/critiquebrainz/frontend/views/review.py
@@ -228,12 +228,15 @@ def create(entity_type=None, entity_id=None):
         return redirect(url_for('user.reviews', user_id=current_user.id))
 
     # Checking if the user already wrote a review for this entity
-    reviews, count = db_review.list_reviews(user_id=current_user.id, entity_id=entity_id)
+    reviews, count = db_review.list_reviews(user_id=current_user.id, entity_id=entity_id, inc_drafts=True)
     review = reviews[0] if count != 0 else None
 
     if review:
-        flash.error(gettext("You have already published a review for this entity"))
-        return redirect(url_for('review.entity', id=review["id"]))
+        if review['is_draft']:
+            return redirect(url_for('review.edit', id=review['id']))
+        else:
+            flash.error(gettext("You have already published a review for this entity"))
+            return redirect(url_for('review.entity', id=review["id"]))
 
     if current_user.is_review_limit_exceeded:
         flash.error(gettext("You have exceeded your limit of reviews per day."))
--- a/critiquebrainz/frontend/views/review.py
+++ b/critiquebrainz/frontend/views/review.py
@@ -228,12 +228,14 @@ def create(entity_type=None, entity_id=None):
         return redirect(url_for('user.reviews', user_id=current_user.id))
 
     # Checking if the user already wrote a review for this entity
-    reviews, count = db_review.list_reviews(user_id=current_user.id, entity_id=entity_id, inc_drafts=True)
+    reviews, count = db_review.list_reviews(user_id=current_user.id, entity_id=entity_id, inc_drafts=True, inc_hidden=True)
     review = reviews[0] if count != 0 else None
 
     if review:
         if review['is_draft']:
             return redirect(url_for('review.edit', id=review['id']))
+        elif review['is_hidden']:
+            return redirect(url_for('review.entity', id=review['id']))
         else:
             flash.error(gettext("You have already published a review for this entity"))
             return redirect(url_for('review.entity', id=review["id"]))
--- a/critiquebrainz/frontend/views/test/test_review.py
+++ b/critiquebrainz/frontend/views/test/test_review.py
@@ -104,6 +104,14 @@ class ReviewViewsTestCase(FrontendTestCase):
         response = self.client.get("/review/%s" % review["id"])
         self.assert404(response, "Drafts shouldn't be publicly visible.")
 
+    def test_draft_redirect(self):
+        review = self.create_dummy_review(is_draft=True)
+        self.temporary_login(self.user)
+        response = self.client.get('/review/write/{}/{}/'
+                                   .format(review["entity_type"], review["entity_id"]))
+        redirect_url = urlparse(response.location)
+        self.assertEquals(redirect_url.path, url_for('review.edit', id=review['id']))
+
     def test_missing_review(self):
         response = self.client.get("/review/aef06569-098f-4218-a577-b413944d9493")
         self.assert404(response)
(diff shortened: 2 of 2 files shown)


critiquebrainz/frontend/views/review.py
```diff
@@ -226,12 +226,15 @@ def create():
         return redirect(url_for('user.reviews', user_id=current_user.id))
 
     # Checking if the user already wrote a review for this entity
-    reviews, count = db_review.list_reviews(user_id=current_user.id, entity_id=entity_id)
+    reviews, count = db_review.list_reviews(user_id=current_user.id, entity_id=entity_id, inc_drafts=True)
```
alastair, line 229: What about hidden reviews here? the default of `inc_hidden` is still False.

critiquebrainz/frontend/views/review.py
```diff
@@ -226,12 +226,15 @@ def create():
...
     review = reviews[0] if count != 0 else None
 
     if review:
-        flash.error(gettext("You have already published a review for this entity!"))
-        return redirect(url_for('review.entity', id=review["id"]))
+        if review['is_draft']:
```
alastair, line 233: This absolutely needs a test

alastair (changes requested):

critiquebrainz/frontend/views/review.py
amCap1712: Yeah, right. I was wondering how we should handle those. If I am right, hidden reviews are reviews that are not public but visible to the author. But since we are using the user id, the hidden review will be from the user so we probably redirect the user to their review ? Does that make sense to do ?

critiquebrainz/frontend/views/review.py
amCap1712: Yeah, sure. Will add one.

critiquebrainz/frontend/views/review.py
alastair: yes, redirecting to the review with a message sounds like a good idea.
You should work out what this flag means. Is it set if someone flags a review as bad, or is that something else? Who can set a review as hidden? An admin, or only the author?

amCap1712: Here is what I figured out. Moderators can hide reviews. Any user including the author and excluding moderators gets a 403 error if they try to access a hidden review.

critiquebrainz/frontend/views/review.py
```diff
@@ -226,12 +226,17 @@ def create():
...
+    reviews, count = db_review.list_reviews(user_id=current_user.id, entity_id=entity_id, inc_drafts=True, inc_hidden=True)
     review = reviews[0] if count != 0 else None
 
     if review:
-        flash.error(gettext("You have already published a review for this entity!"))
-        return redirect(url_for('review.entity', id=review["id"]))
+        if review['is_draft']:
+            return redirect(url_for('review.edit', id=review['id']))
+        elif review['is_hidden']:
+            return redirect(url_for('review.entity', id=review['id']))
```
alastair, line 236: What does this mean if a user tries to write a review for an item that has been hidden by an admin?
It means that they'll try and go to (for example) /write/artist/[id], which will redirect them to /review/[reviewid], which will then raise this alert, right? https://github.com/metabrainz/critiquebrainz/blob/159d934ff7886ab33cd72e89f6e1cf0f26fe9999/critiquebrainz/frontend/views/review.py#L77-L81

critiquebrainz/frontend/views/test/test_review.py
```diff
@@ -101,6 +101,11 @@ def test_draft_review(self):
         response = self.client.get("/review/%s" % review["id"])
         self.assert404(response, "Drafts shouldn't be publicly visible.")
 
+        data = {review["entity_type"]: review["entity_id"]}
+        response = self.client.get('/review/write/', query_string=data)
+        redirect_url = urlparse(response.location)
+        self.assertEquals(redirect_url.path, url_for('review.edit', id=review['id']))
```
alastair, line 107: I still don't see a test that redirects /review/write?artist=1234 -> /review/write/artist/1234 in the case that no review exists yet.

alastair: oh wait, that's the one I asked for in https://github.com/metabrainz/critiquebrainz/pull/298#issuecomment-691968954, sorry!

critiquebrainz/frontend/views/review.py
amCap1712: AFAIU, items are not hidden by admin but rather `review`s are. So, the case added above is executed when a person wrote a review for some item which was subsequently hidden by the admin and then they try to write a review again. This redirect will take them to the review page where they'll be shown the warning you linked.

critiquebrainz/frontend/views/test/test_review.py
amCap1712: Yup, and that test is also added at https://github.com/metabrainz/critiquebrainz/pull/298/files#diff-68c7d6f53d1f78aada84956c9906f4e7R143.

alastair:
@brainzbot test this please

critiquebrainz/frontend/views/test/test_review.py
```diff
@@ -312,3 +320,20 @@ def test_hide_unhide(self):
             follow_redirects=True,
         )
         self.assertIn("Review is not hidden.", str(response.data))
+
+    def test_hide_redirect(self):
```
alastair, line 324: I think that we should split this test into specific tests for each of these two things

* Have a user object (instead of `self.hacker`, for example `self.moderator`) that makes `is_admin` return True
* Have a specific view test that makes a post to /review/x/hide, and then reads the review from the database to check that the hidden flag is True
* Have a separate test that creates a hidden review directly (using create_dummy_review), and then tries to GET to /review/write/x/y and check that the redirect url (and any specific popup messages) are correct.

amCap1712: I think the first two parts are already covered in this [test_hide_unhide](https://github.com/metabrainz/critiquebrainz/blob/d9d787bd3b6d30b8f894019a6159d44d00d46482/critiquebrainz/frontend/views/test/test_review.py#L254).

amCap1712: I wanted to use `create_dummy_review` but it does not take a `is_hidden` parameter. But that is not much an issue, as there is a `set_hidden_state` so that is easy to fix.

alastair: This looks much better, thanks

alastair (approved):

alastair:
This has a conflict now that I've merged some other PRs, can you resolve it?

PR #265, opened 2019-06-16
CB-270: Implemented the reviewal of recordings

spellew:

PR #303, opened 2020-09-22
CB-270: Implemented the reviewal of recordings 

amCap1712:
This PR builds on spellew's work in #265 with aim to sync it with latest codebase and get it merged.

JIRA: CB-270


amCap1712:
The tests passed locally but I am unable to view the website locally because probably the sql schema changes need to applied somehow and I am no aware how to do so.

alastair:
You should be able to run `./develop.sh psql` to get a postgres shell, and then just run the contents of 18.sql

amCap1712:
The PR adds the feature to review recordings. Necessary changes including database schema changes, UI to search, review and browse a recording have been added and also the backend changes to support the same.

In the original PR, according to the last comment some tests were failing. However, after the rebase, all tests passed. However, this PR does not add any new tests to test reviewing the recordings. Some other entities are also missing the tests. I'll add some tests to this PR otherwise it is ready to be merged.

alastair (changes requested):
I'm just marking this PR as "Request changes" for tracking purposes. I'll review it again once you write the tests that you suggested.

critiquebrainz/frontend/external/musicbrainz_db/recording.py
```diff
@@ -0,0 +1,22 @@
...
+
+    Args:
+        mbid (uuid): MBID(gid) of the recording.
+    Returns:
+        Dictionary containing the recording information
+    """
+    key = cache.gen_key(mbid)
+    recording = cache.get(key)
+    if not recording:
+        recording = fetch_multiple_recordings(
```
alastair, line 17: any particular reason why we use the fetch_multiple method instead of the already existing `get_recording_by_mbid` in BU?

critiquebrainz/frontend/templates/entity_review.html
```diff
@@ -7,6 +7,8 @@
     {{ _('%(artist)s', artist = '<b>'|safe + entity.name | default(_('[Unknown artist]')) + '</b>'|safe) }}
   {% elif review.entity_type == 'label' %}
     {{ _('%(label)s', label = '<b>'|safe + entity.name | default(_('[Unknown label]')) + '</b>'|safe) }}
+  {% elif review.entity_type == 'recording' %}
+    {{ _('%(recording)s', recording = '<b>'|safe + entity.name | default(_('[Unknown recording]')) + '</b>'|safe) }}
```
alastair, line 11: why is the bold tag inside the translated block? can't we just do 

    <b>{{entity.name}}</b> 

?
Additionally, I don't know why we're marking the entity name itself as translateable, that probably isn't required.
What is the "Unknown x" default for? How can you make a review for an entity that doesn't exist?

critiquebrainz/frontend/templates/macros.html
```diff
@@ -67,6 +67,13 @@
...
       </span>
     {% endif %}
+  {% elif entity_type == 'recording'  %}
+    <img src="{{ get_static_path('images/placeholder_disc.svg') }}" {{ attributes }} />
+    {% if overlay_type %}
+      <span class="entity-type recording">
+        {{ _('Recording') }}
+      </span>
+    {% endif %}
   {% else %} {# release-group #}
```
alastair, line 77: it'd be nice to work out how to remove this fall-through else here, and require this to be called with entity_type="release-group" in those situations

critiquebrainz/frontend/templates/recording/entity.html
```diff
@@ -0,0 +1,140 @@
...
+
+{% block title %}
+  {{ _('Recording "%(name)s" by %(artist)s', name=recording.name, artist=recording['artist-credit-phrase']) }}
+  - CritiqueBrainz
+{% endblock %}
+
+{% block content %}
+<div class="clearfix">
+  <h2 class="pull-left">
+    {% set artist = [] %}
```
alastair, line 12: I think we can do this a bit better, can we do something like 

```
{{recording.name}} _(by) 
{% for credit in recording['artists'] %}
{% if credit.name %}
    <a href="{{url_for('artist.entity', mbid=credit.id)}}">{{credit.name}}</a>
    {% if credit.join_phrase %}{{credit.join_phrase}}{% endif %}
{% endif %}
```

What does it mean when a credit doesn't have a name and we just append it? (line 20)

critiquebrainz/frontend/templates/recording/entity.html
```diff
@@ -0,0 +1,140 @@
...
+            <th>{{ _('Published on') }}</th>
+            <th>{{ _('Votes (+/-)') }}</th>
+          </tr>
+        </thead>
+        <tbody>
+          {% for review in reviews %}
+            <tr data-review-id="{{ review.id }}">
+              <td>
+                <a href="{{ url_for('review.entity', id=review.id) }}">
+                  {{ _('by %(reviewer)s', reviewer='<img class="avatar" src="%s&s=16" /> '|safe % review.user.avatar + review.user.display_name) }}
```
alastair, line 61: again, don't put html inside a variable, unless this is a best practise for translation

critiquebrainz/frontend/templates/recording/entity.html
```diff
@@ -0,0 +1,140 @@
...
+                </a>
+              </td>
+              <td>{{ review.published_on | date }}</td>
+              <td>{{ review.votes_positive_count }}/{{ review.votes_negative_count }}</td>
+            </tr>
+          {% endfor %}
+        </tbody>
+      </table>
+      <ul class="pagination">
```
alastair, line 70: can you open a ticket to abstract out this pagination code so that we don't have the same thing on all pages?

critiquebrainz/frontend/templates/review/entity/recording.html
```diff
@@ -0,0 +1,20 @@
...
+
+{% block title %}
+  {% set recording_title = recording.name | default(_('[Unknown recording]')) %}
+  {{ _('Review of "%(recording)s" by %(user)s', recording=recording_title, user=review.user.display_name) }} - CritiqueBrainz
+{% endblock %}
+
+{% block entity_title %}
+  <h2 id="title">
+    {% if recording %}
+      {% set recording_name = '<a href="%s">' | safe % url_for('recording.entity', id=review.entity_id) ~ recording.name ~ '</a>'|safe %}
```
alastair, line 13: don't put html inside strings

critiquebrainz/frontend/templates/search/results.html
```diff
@@ -101,5 +101,29 @@
...
+      </td>
+      <td>
+        {% if result['length'] %}
+          {{ result['length'] | track_length_ms }}
+        {% else %}
+          -
+        {% endif %}
+      </td>
+      <td>
+        <a href="{{ url_for('artist.entity', mbid=result['artist-credit'][0].artist.id) }}">{{ result['artist-credit'][0]['artist']['name'] or '-' }}</a>
```
alastair, line 117: This isn't strictly correct, it'd be nice to be able to render the full artist credit here. As we have to do this elsewehre, perhaps we should have a method that can do it for us

critiquebrainz/frontend/templates/search/results.html
```diff
@@ -101,5 +101,29 @@
...
+      </td>
+      <td>
+        {% if result['release-list'] %}
+          <a href="{{ url_for('release.entity', id=result['release-list'][0].id) }}">{{ result['release-list'][0].title or '-' }}</a>
```
alastair, line 121: why are we only showing 1 release? (I agree that this could get problematic with some recordigns). Maybe we could just show the "most correct" release here? (whatever that means)

critiquebrainz/frontend/templates/search/selector_results.html
```diff
@@ -131,6 +131,34 @@
     </td>
     <td>
       <input type="radio" name="label" value="{{ result['id'] }}">
+      <input type="radio" name="artist" value="{{ result['id'] }}">
```
alastair, line 134: I opened CB-375 about this
How much do these results differ from `search/results.html`? Are we able to merge them into a single file?

critiquebrainz/frontend/templates/user/reviews.html
```diff
@@ -60,6 +60,22 @@
                   {% endif %}
                   )
                 {% endif %}
+
+              {%- elif review.entity_type == 'recording' -%}
+                <strong>{{ entity.name | default(_('[Unknown recording]')) }}</strong>
+                - {{ entity['artist-credit-phrase'] | default(_('[Unknown artist]')) }}
+
+              {%- elif review.entity_type == 'artist' -%}
```
alastair, line 68: were these items previously missing?

alastair (changes requested):
A bunch of odd stuff here, but it's clear that a lot of it has been duplicated from existing code. Where it makes sense, we should improve the way we present this data, especially in the templates.
Where it's too much work to make modifications in this PR, we should open tickets to improve the code generally

critiquebrainz/frontend/templates/search/selector_results.html
amCap1712: I think the files are mostly similar and we can merge them together.

critiquebrainz/frontend/external/musicbrainz_db/recording.py
amCap1712: I think these functions got moved to BU but this code was not updated.

critiquebrainz/frontend/templates/user/reviews.html
amCap1712: Yeah, I checked the `git blame` seems so.

critiquebrainz/frontend/templates/search/results.html
amCap1712: I agree but what should be the "most correct" release?

critiquebrainz/frontend/templates/recording/entity.html
amCap1712: Done. https://tickets.metabrainz.org/browse/CB-406

critiquebrainz/frontend/external/musicbrainz_db/entities.py
```diff
@@ -31,6 +33,7 @@ def get_multiple_entities(entities):
     release_group_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'release_group', entities)]
     artist_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'artist', entities)]
     label_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'label', entities)]
+    recording_mbids = [entity[0] for entity in filter(lambda entity: entity[1] == 'recording', entities)]
```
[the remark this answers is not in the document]
amCap1712: Fixed.

critiquebrainz/frontend/external/musicbrainz_db/recording.py
amCap1712: Fixed.

critiquebrainz/frontend/templates/entity_review.html
amCap1712: Fixed. I have removed the Unknown defaults as well.

critiquebrainz/frontend/templates/macros.html
amCap1712: I checked it is being called with `release_group` already elsewhere so simply changing else to elif was enough.

critiquebrainz/frontend/templates/recording/entity.html
amCap1712: Done.

critiquebrainz/frontend/templates/recording/entity.html
amCap1712: Fixed.

critiquebrainz/frontend/templates/search/results.html
amCap1712: Fixed.

critiquebrainz/frontend/templates/review/entity/recording.html
```diff
@@ -0,0 +1,18 @@
...
+
+{% block title %}
+  {% set recording_title = recording.name | default(_('[Unknown recording]')) %}
+  {{ _('Review of "%(recording)s" by %(user)s', recording=recording_title, user=review.user.display_name) }} - CritiqueBrainz
+{% endblock %}
+
+{% block entity_title %}
+  <h2 id="title">
+    {% if recording %}
+      <a href="{{ url_for('recording.entity', id=review.entity_id) }}">recording.name</a>
```
jdaok, line 13:
```suggestion
      <a href="{{ url_for('recording.entity', id=review.entity_id) }}">{{ recording.name }}</a>
```

So it displays the recordings actual name!

critiquebrainz/frontend/templates/search/results.html
amCap1712: CB-418

PR #270, opened 2019-07-19
CB-356: Implemented handling of deleted entities

spellew:
Set unknown_entities_for_missing flag to True for fetch_multiple_* calls

PR #304, opened 2020-09-22
CB-356: Implemented handling of deleted entities

amCap1712:
This supersedes #270.

JIRA: CB-356

### DIFF
--- a/critiquebrainz/frontend/external/musicbrainz_db/artist.py
+++ b/critiquebrainz/frontend/external/musicbrainz_db/artist.py
@@ -18,6 +18,7 @@ def get_artist_by_id(mbid):
         artist = fetch_multiple_artists(
             [mbid],
             includes=['artist-rels', 'url-rels'],
+            unknown_entities_for_missing=True,
         ).get(mbid)
         cache.set(key=key, val=artist, time=DEFAULT_CACHE_EXPIRATION)
     return artist_rel.process(artist)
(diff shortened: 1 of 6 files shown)


amCap1712:
The last comments on the PR make it unclear what the current state is. I'll try to test it and see if it works or more changes are required.

alastair:
I've been thinking about this in more detail, in general I'm not a fan of the `unknown_entities_for_missing` flag and the `NoDataFoundException` exception.
Although it's more work, I think that we should refactor BU to remove these flags and use the following behaviour:

* Bulk get methods (e.g. `get_entities_by_gids` or `fetch_multiple_artists`) take a list of ids, and return a dictionary of `{id: entity}` only for those ids which exist in the database. Update the documentation to make this clear
* individual get methods (e.g. `get_artist_by_id`) can raise NoDataFoundException if there is no item. I'm not convinced either way if we should raise an exception or just return None.

This will probably require significant changes to BU and to CB, but I think that it will simplify much of this work. We should do a review of the existing code to see if there is anywhere that calls the `multiple` load methods, but expect an exception to be raised, but I don't think there will be any. I'll open a BU ticket.

amCap1712:
Makes sense to me. I'll work on this ASAP. The CB website homepage is down for 2 or more days and I almost sure that the reason is a merged or deleted entity.

PR #305, opened 2020-09-22
CB-380: "Write a review" error messages don't give any context

amCap1712:
There are two things about the issue in question:-
1) Not a valid choice issue. This occurs because of the default value of the Flask Validator. However, the specific message that should be displayed for the validation failure has an extra Not a Valid Choice message. I checked the WTF forms docs and they [recommend ](https://wtforms.readthedocs.io/en/2.3.x/validators/#wtforms.validators.DataRequired) that `InputRequired` be used unless there is a specific reason to use `DataRequired`. Using, `InputRequired` also fixes the issue.
However, there are other places in the codebase where `DataRequired` is used. Do we want to change those to `InputRequired` as well ?

2) Changing the _exclamation_ to _period_ is simple. But a [livegrep](http://livegrep.metabrainz.org/search/livegrep?q=file%3A.py%20%22%5B%5Cw%5Cs%5D%2B!%22&fold_case=auto&regex=true&context=true&repo%5B%5D=metabrainz%2Fcritiquebrainz) shows _!_ are used at quite a few places. Do we want to change them as well, if yes which ones ?

JIRA: CB-380


alastair:
Let's open a ticket for replacing `DataRequired` -> `InputRequired`.

Let's not worry about changing all of the !'s for now. Many of the search results that you linked are in tests, so we don't need to change those. If we're editing some part of the code and see a ! that doesn't make sense we can slowly change them at that time. While we're at it can we remove the ! from "You need to choose a license!"?  My preference is for these error messages to not have a . at the end of them

PR #308, opened 2020-09-23
CB-296: `init_postgres` function in `manage.py` doesn't work for the development environment

amCap1712:
I was unable to find where this is done in the Docker but according to the comment on the ticket this fix seemed right.

JIRA: CB-296

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -115,7 +115,6 @@ def init_db(skip_create_db=False, test_db=False, force=False):
         db_uri = frontend.create_app().config['SQLALCHEMY_DATABASE_URI']
 
     if not skip_create_db:
-        init_postgres(db_uri)
         create_extension(db_uri)
 
     click.echo("Creating tables... ", nl=False)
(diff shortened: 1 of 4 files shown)


alastair (approved):
Perfect, thanks.

The database is automatically initialized with docker based on these environment variables: https://github.com/metabrainz/critiquebrainz/blob/dd20a4125010b81744ad9f2182daa84ec869c024/docker/docker-compose.dev.yml#L11-L14

amCap1712:
Ok! Understood now. Thanks

manage.py
```diff
@@ -115,7 +115,6 @@ def init_db(skip_create_db=False, test_db=False, force=False):
         db_uri = frontend.create_app().config['SQLALCHEMY_DATABASE_URI']
 
     if not skip_create_db:
-        init_postgres(db_uri)
         create_extension(db_uri)
```
alastair, line 118: I just noticed that we also create the extension with this entrypoint mount:
https://github.com/metabrainz/critiquebrainz/blob/master/docker/docker-compose.dev.yml#L18

This means that we can also remove `create_extension` and `skip_create_db`

alastair (changes requested):

manage.py
alastair: check for uses in the code of `--skip-create-db`, there are some in the test scripts, and also in the documentation

alastair (approved):

PR #309, opened 2020-09-23
CB-376: Breaking change in API (created => published_on)

amCap1712:
Check for the created parameter and map it to to published_on for backward compatibility.

JIRA: CB-376

### DIFF
--- a/critiquebrainz/ws/review/views.py
+++ b/critiquebrainz/ws/review/views.py
@@ -349,10 +349,14 @@ def review_list_handler():
         entity_type = Parser.string('uri', 'entity_type', valid_values=ENTITY_TYPES, optional=True)
 
     user_id = Parser.uuid('uri', 'user_id', optional=True)
-    # TODO: "rating" sort value is deprecated and needs to be removed.
-    sort = Parser.string('uri', 'sort', valid_values=['popularity', 'published_on', 'rating'], optional=True)
+    sort = Parser.string('uri', 'sort', valid_values=['popularity', 'published_on', 'rating', 'created'], optional=True)
+
+    # "rating" and "created" sort values are deprecated and but allowed here for backward compatibility
+    if sort == 'created':
+        sort = 'published_on'
     if sort == 'rating':
         sort = 'popularity'
+
     limit = Parser.int('uri', 'limit', min=1, max=50, optional=True) or 50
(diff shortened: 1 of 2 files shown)


critiquebrainz/ws/review/views.py
```diff
@@ -350,7 +350,10 @@ def review_list_handler():
 
     user_id = Parser.uuid('uri', 'user_id', optional=True)
     # TODO: "rating" sort value is deprecated and needs to be removed.
```
alastair, line 352: Can we combine these two comments together?

critiquebrainz/ws/review/views.py
```diff
@@ -350,7 +350,10 @@ def review_list_handler():
...
-    sort = Parser.string('uri', 'sort', valid_values=['popularity', 'published_on', 'rating'], optional=True)
+    sort = Parser.string('uri', 'sort', valid_values=['popularity', 'published_on', 'rating', 'created'], optional=True)
```
alastair, line 353: I had a quick look in critiquebrainz/ws/review/views_test.py and don't see any tests that check these parameters (or if they're rewritten correctly). Do you think that they are necessary?

alastair (commented):
This looks good, but I think we should have some tests for these parameters too.

The other comment on the ticket was about the documentation not showing the correct parameters, but I've since rebuilt the docs and they appear to be up to date now.

alastair:
@brainzbot test this please

critiquebrainz/ws/review/views.py
amCap1712: I'll add the tests. In general, I feel CB is missing a lot of tests. Many of the bug fixes we did could have been easily caught by simple tests and fixed before merging. Moving forward I think we should add tests for at least the changes we do. Later, we can try to add tests for features that already present in the code base but do not have tests.

critiquebrainz/ws/review/views.py
amCap1712: Will do.

critiquebrainz/ws/review/views.py
alastair: This is a great way of working. When fixing something, leave the code in a better state than when you arrived. This could include cleaning up some comments, fixing duplicated code, doing something in a way that is easier to understand, or writing tests. Keep this in mind especially when you're fixing these simple tickets.

PR #310, opened 2020-09-23
CB-381: Dump importer doesn't reset sequence values after importing items

amCap1712:
I was unable to test this fix because it requires the full MB data dump as well to display the website correctly. Also, the ticket does not explicitly mention it but I assumed that it meant to reset the revision table ID field because it of type SERIAL while other id columns have type uuid.


alastair (approved):
This looks like the correct fix, thanks. Yes you are right - this only has to be done on fields that are SERIAL, so I think that just the revision table is correct.
I'll do a test import before merging.

PR #311, opened 2020-09-25
CB-382: You can sometimes create a review without a revision

amCap1712:
It might be possible that there may be other uses or potential use case
for creating a revision in a separate transaction. Hence, added the
connection as an optional parameter. To enforce the creation in the
same transaction we can make the parameter mandatory.

JIRA: CB-382

### DIFF
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -240,7 +240,7 @@ def update(review_id, *, drafted, text=None, rating=None, license_id=None, langu
         updated_info["review_id"] = review_id
         with db.engine.connect() as connection:
             connection.execute(query, updated_info)
-            db_revision.create(review_id, text, rating, sql_connection=connection)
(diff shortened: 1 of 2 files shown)


alastair:
Because a revision is only created when making a new review or updating a review, and both of them require a transaction, I don't think we need the connection parameter to be optional. I think it's fine to make it required.

amCap1712:
Yeah, right. Makes sense. I was trying to be defensive to avoid break existing code just in case. I'll instead do a thorougher check of the code base and make the parameter required.

alastair (changes requested):
marking as request changes for your further work

amCap1712:
Something weird is happening with the tests. Running `sudo ./test.sh` leads to 5 test failures with the reason that connection to database is closed. However, running `sudo ./develop.sh test up` causes 9 tests to fail with different reasons.

alastair:
@brainzbot test this please

PR #312, opened 2020-09-25
CB-372: Error when trying to submit a rating twice

amCap1712:
The issue was caused due to reviews that do not have a revision. #311 fixes the issue by creating the review and revision in the same transaction. This PR  improves the frontend check for this.

Jira: CB-372

### DIFF
--- a/critiquebrainz/frontend/views/event.py
+++ b/critiquebrainz/frontend/views/event.py
@@ -55,7 +55,7 @@ def entity(id):
             entity_type='event',
             user_id=current_user.id
         )
-        my_review = my_reviews[0] if my_count else None
+        my_review = my_reviews[0] if my_reviews else None
     else:
         my_review = None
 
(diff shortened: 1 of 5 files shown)


alastair:
Can you double-check what the behaviour is if there is only a review row (and no revision), and a user tries to write/update a review?
I have a feeling that [this check](https://github.com/metabrainz/critiquebrainz/blob/dd20a4125010b81744ad9f2182daa84ec869c024/critiquebrainz/frontend/views/review.py#L231) will cause the create page to be loaded, instead of the edit page, and therefore it will try and create a new review instead of updating the existing one.
I'm not really sure if this comment is more related to this PR or #311

amCap1712:
Yeah, right. We can probably add a check here to single out when revision is missing but I think we should rather handle it as a part of #311. We should somehow inspect the database and create the missing revisions.

alastair:
I checked, and there are no reviews that have no revision. Therefore I think that this change is enough. thanks

alastair (approved):

alastair (approved):

PR #314, opened 2020-10-05
Miscellaneous Bug Fixes 

amCap1712:
1) Bugfix for showing work title in search results
2) Due to some reason, the language selector for writing a review was using pycountry as a fallback if babel could not resolve the Locale. However, while displaying the language name on the view review page only babel was being used. pycountry is not added as a fallback in this case as well. This fixes CB-393.

### DIFF
--- a/critiquebrainz/frontend/forms/utils.py
+++ b/critiquebrainz/frontend/forms/utils.py
@@ -1,0 +1,9 @@
+import pycountry
+from babel.core import UnknownLocaleError, Locale
+
+
+def get_language_name(language_code):
+    try:
+        return Locale(language_code).language_name
+    except UnknownLocaleError:
+        return pycountry.languages.get(iso639_1_code=language_code).name
--- a/critiquebrainz/frontend/__init__.py
+++ b/critiquebrainz/frontend/__init__.py
@@ -3,6 +3,7 @@ import os
 import sys
 from time import sleep
 
+import pycountry
 from brainzutils.flask import CustomFlask
 from flask import send_from_directory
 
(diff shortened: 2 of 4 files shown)


amCap1712:
Regarding language support, I found [this](https://github.com/metabrainz/critiquebrainz/blob/d1c1c175209ec78bcced1dbfd5bd64a46be2d1f4/critiquebrainz/db/review.py#L29) in `db/review.py`. I think we should not be using `flask-babel to resolve the language. What do you suggest ?

alastair:
Can you please update the description of this PR to include any ticket numbers if valid, and a brief description of what the bug was and what you fixed?

critiquebrainz/frontend/forms/utils.py
```diff
@@ -0,0 +1,9 @@
+import pycountry
+from babel.core import UnknownLocaleError, Locale
+
+
+def get_language_name(language_code):
```
alastair, line 5: This should have some documentation that explains what it does, and why there are the two different versions. Include some examples of how it works in each case, and why the first version might fail.

We should also have some tests, which specifically check our `no` bug.

Does it make sense to put this in `forms/utils.py`? I'm just thinking because we load it in `frontend/__init__.py`. Maybe there's a more generic frontend utils package where we could put it.

amCap1712: Yeah, actually regarding this I think `babel` or `pycountry` should be removed from this part. I do not think we should be doing using both.

amCap1712: The issue arises from https://github.com/metabrainz/critiquebrainz/blob/2d7a0c1958b5bd221b4340efee30123b4bcb66a7/critiquebrainz/db/review.py#L29-L32. Here, the language codes are stored in a list from pycountry. While getting the language name, we try first babel then pycountry. I think we should drop the `babel` part from this method.

PR #315, opened 2020-10-06
CB-383: Move tests to separate packages

amCap1712:
CB-383


alastair (changes requested):
You named these modules `tests`, but other places in CB and our other projects use `test`. Could you change to that instead?

Is the babel update to requirements.txt related to this change?

amCap1712:
No, that babel change does not belong here. It was an error and an oversight on my part. Apologies!!

I have changed the directory name to `test`.

PR #316, opened 2020-10-08
CB-388: Reduce scope of mocks

amCap1712:
CB-388

### DIFF
--- a/critiquebrainz/frontend/views/test/test_mapping.py
+++ b/critiquebrainz/frontend/views/test/test_mapping.py
@@ -89,7 +89,7 @@ class SpotifyMappingViewsTestCase(FrontendTestCase):
         get_release_group_by_id.reset_mock()
         get_release_group_by_id.return_value = self.test_release_group
         response = self.client.get("/mapping/6b3cd75d-7453-39f3-86c4-1441f360e121")
-        logging.error(str(response))
+        logging.error(str(response.data))
         self.assert200(response)
         self.assertIn("No mappings", str(response.data))
 
@@ -129,7 +129,7 @@ class SpotifyMappingViewsTestCase(FrontendTestCase):
         search.side_effect = ExternalServiceException
         response = self.client.get("/mapping/spotify/add",
                                    query_string={"release_group_id": "6b3cd75d-7453-39f3-86c4-1441f360e121"})
-        logging.error(str(response))
+        logging.error(str(response.data))
         self.assertStatus(response, 503)
 
         # test when response has no albums for given id
--- a/critiquebrainz/frontend/views/test/test_mapping.py
+++ b/critiquebrainz/frontend/views/test/test_mapping.py
@@ -15,9 +15,7 @@
 # You should have received a copy of the GNU General Public License along
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
-import logging
 from unittest import mock
-
 from flask import url_for
 
 import critiquebrainz.db.users as db_users
@@ -86,10 +84,9 @@ class SpotifyMappingViewsTestCase(FrontendTestCase):
         self.assert404(response, "Can't find release group with a specified ID.")
 
         # test for release group with no mappings
-        get_release_group_by_id.reset_mock()
+        get_release_group_by_id.reset_mock(side_effect=True)
         get_release_group_by_id.return_value = self.test_release_group
         response = self.client.get("/mapping/6b3cd75d-7453-39f3-86c4-1441f360e121")
-        logging.error(str(response.data))
         self.assert200(response)
         self.assertIn("No mappings", str(response.data))
 
@@ -124,16 +121,15 @@ class SpotifyMappingViewsTestCase(FrontendTestCase):
         self.assertIn("Only existing release groups can be mapped to Spotify!", str(response.data))
 
         # test Spotify service unavailable
-        get_release_group_by_id.reset_mock()
+        get_release_group_by_id.reset_mock(side_effect=True)
         get_release_group_by_id.return_value = self.test_release_group
         search.side_effect = ExternalServiceException
         response = self.client.get("/mapping/spotify/add",
                                    query_string={"release_group_id": "6b3cd75d-7453-39f3-86c4-1441f360e121"})
-        logging.error(str(response.data))
         self.assertStatus(response, 503)
 
         # test when response has no albums for given id
-        search.reset_mock()
+        search.reset_mock(side_effect=True)
         search.return_value = self.test_spotify_search_response
 
         get_multiple_albums.return_value = {}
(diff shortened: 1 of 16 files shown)


critiquebrainz/db/review_test.py
```diff
@@ -211,6 +207,7 @@ def test_list_reviews(self):
         self.assertEqual(count, 1)
         self.assertEqual(len(reviews), 1)
 
+    @mock.patch('brainzutils.cache.get', Mock(return_value=None))
```
alastair, line 210: this is a different syntax to the other uses of `mock.patch`. Can we use the other version, where you pass an argument into the method?

critiquebrainz/frontend/external/spotify_test.py
```diff
@@ -63,76 +61,86 @@ def setUp(self):
...
+        requests_get.return_value = self.ok_response
+        fetch_spotify_token.return_value = "valid-token"
         self.assertDictEqual(spotify._get("test-query"), self.some_albums_response)
-        spotify._fetch_access_token.assert_called_with(refresh=False)
+        fetch_spotify_token.assert_called_with(refresh=False)
 
         # test when _fetch_access_token returns expired token
-        requests.get = MagicMock(return_value=self.unauth_response)
-        spotify._fetch_access_token = MagicMock(return_value="expired-token")
+        requests_get.return_value = self.unauth_response
```
alastair, line 74: if we use the mock twice in a single test, we should use `.reset_mock()` on it before setting new values and asserting that it is called

critiquebrainz/frontend/views/test/test_mapping.py
```diff
@@ -78,64 +74,73 @@ def setUp(self):
             }
         }
 
-    def test_spotify_list(self):
+    @mock.patch('critiquebrainz.frontend.external.musicbrainz_db.release_group.get_release_group_by_id')
+    @mock.patch('critiquebrainz.frontend.external.spotify.get_multiple_albums')
+    @mock.patch('critiquebrainz.frontend.external.mbspotify.mappings')
+    def test_spotify_list(self, get_mappings, get_spotify_albums, get_release_groups):
```
alastair, line 80: just to make it a bit easier to follow, can we try and keep method names the same as the method that it's mocking? (in this case, `mappings, get_multiple_albums, get_release_group_by_id`)

alastair: in the case where we have to mock a method with the same name in 2 different packages, we could prefix it with part of the package name if needed

critiquebrainz/frontend/views/test/test_mapping.py
```diff
@@ -144,31 +149,40 @@ def test_spotify_add(self):
         self.assertIn("Test Album", str(response.data))
         self.assertIn("Test Artist", str(response.data))
 
-    def test_spotify_confirm(self):
+    @mock.patch('critiquebrainz.frontend.external.musicbrainz_db.release_group.get_release_group_by_id')
+    @mock.patch('critiquebrainz.frontend.external.spotify.get_multiple_albums')
+    @mock.patch('critiquebrainz.frontend.external.mbspotify.add_mapping')
+    @mock.patch('critiquebrainz.frontend.external.spotify.get_album')
+    @mock.patch('critiquebrainz.frontend.external.spotify.search')
```
alastair, line 156: are all of these mocks used in each of the tests that are in this method? I think it might be more explicit if we have separate methods for each, and we only add mocks for the methods that we know we need to mock.

However, if you want to check `assert_not_called` on these mocks, perhaps this is ok.

critiquebrainz/frontend/views/test/test_rate.py
```diff
@@ -47,8 +44,10 @@ def setUp(self):
             full_name="Test License.",
         )
 
-    def test_rate(self):
+    @mock.patch('critiquebrainz.frontend.external.musicbrainz_db.release_group.get_release_group_by_id')
+    def test_rate(self, get_release_group):
         self.temporary_login(self.reviewer)
+        get_release_group.return_value = self.test_entity
```
alastair, line 50: would be nice to `assert_called_with` this mock, and other mocks in similar tests

critiquebrainz/frontend/views/test/test_review.py
```diff
@@ -29,6 +29,7 @@ def mock_get_entity_by_id(id, type='release_group'):
         }
 
 
+@mock.patch('brainzutils.musicbrainz_db.release.browse_releases', Mock(return_value=None))
```
alastair, line 32: I would prefer to explicitly mock this on all methods where it's needed, instead of at a class level

critiquebrainz/frontend/views/test/test_review.py
```diff
@@ -264,21 +264,23 @@ def test_place_review_pages(self):
...
-    def test_hide_unhide(self):
+    @mock.patch('critiquebrainz.db.user.User.is_admin')
+    def test_hide_unhide(self, is_user_admin):
         # create a review by self.user and check it's not hidden
         review = self.create_dummy_review()
         self.assertEqual(review["is_hidden"], False)
 
         # make self.hacker as current user
         self.temporary_login(self.hacker)
+        is_user_admin.return_value = False
```
alastair, line 275: It's be better to test this by adding the username to the admin list in the config, instead of mocking this methods. 
With CB-385 this will become easier.

alastair (changes requested):
This is mostly good, just a few small changes. 

All of the tests still pass?

critiquebrainz/frontend/views/test/test_review.py
amCap1712: Sure, for now let's leave this the way it is for now. I'll add a note to CB-385 to fix this as a part of it.

critiquebrainz/frontend/views/test/test_rate.py
amCap1712: Can you explain what you mean by similar tests ?

critiquebrainz/frontend/views/test/test_mapping.py
amCap1712: Yeah, I checked and removed some unneeded mocks after your review but these four are needed and and cannot be removed. We could break this into parts but at least one test requires all 4 mocks.

critiquebrainz/frontend/views/test/test_rate.py
amCap1712: Also, this was mock turned out to be unneeded so I removed it.

critiquebrainz/frontend/views/test/test_mapping.py
amCap1712: Also, this is a part of the mapping code. I hope this will soon get removed once we add BrainzPlayer.

alastair:
@amCap1712 I rebased this against master to fix the failing tests that I fixed. I think that we should do it again after we merge #323 so that we can see if there are any new lint warnings introduced by this change. I think I see a few of them, but they're mixed up with the other ones that I fixed.

It looks like the failing tests are new, and we should look at those. I'm happy to take over this PR if you're busy. Let me know what you think of #323 and I'll get on it.

alastair:
@brainzbot test this please

alastair:
This is working for me now, I had a bit of trouble with tests on jenkins - the timeout to the musicbrainz db is about 6 minutes, and it takes a bit longer than this for the sample database to load when the CI machine is under load. I'm not sure what the best solution to this is, it's annoying to have to download and import this data for every single test run, but I don't think there's an easier way to make this load faster, unless we a) mount postgres data directory into the host (causes problems if we run two tests at the same time, makes it harder to reset the data), or b) push a db image that already includes this data (would be large, and would need manual management to push new versions with new data)

amCap1712:
@brainzbot retest this please

amCap1712:
@brainzbot retest this please

amCap1712:
The CI is passing now. However, there is still an issue with the database not being up while running tests locally. Could it be due to some difference in what `tesh.sh` and the Jenkins CI test script doing?

PR #318, opened 2020-10-16
CB-396: Add webservice parameter to return reviews with only text or ratings or both

AbhinavOhri:
CB-396

### DIFF
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -333,7 +333,7 @@ def create(*, entity_id, entity_type, user_id, is_draft, text=None, rating=None,
 # pylint: disable=too-many-branches
 def get_reviews_list(connection, *, inc_drafts=False, inc_hidden=False, entity_id=None,
                      entity_type=None, license_id=None, user_id=None, language=None,
-                     exclude=None, sort=None, limit=20, offset=None):
+                     exclude=None, sort=None, limit=20, offset=None, review_type=None):
     """
         helper function for list_reviews() that extends support for execution within a transaction by directly receiving the
         connection object
@@ -488,12 +488,18 @@ def get_reviews_list(connection, *, inc_drafts=False, inc_hidden=False, entity_i
                 "created": row.pop("user_created"),
             })
 
+        if review_type == 'rating':
+            rows = list(filter(lambda x: x['rating'], rows))
+        elif review_type == 'review':
+            rows = list(filter(lambda x: x['text'], rows))
+        count = len(rows)
+
     return rows, count
 
 
 def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None, entity_type=None,
                  license_id=None, user_id=None, language=None, exclude=None,
-                 sort=None, limit=20, offset=None):
+                 sort=None, limit=20, offset=None,review_type=None):
     """Get a list of reviews.
 
     This function provides several filters that can be used to select a subset of reviews.
@@ -520,7 +526,7 @@ def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None, entity_t
     with db.engine.connect() as connection:
         return get_reviews_list(connection, inc_drafts=inc_drafts, inc_hidden=inc_hidden, entity_id=entity_id,
                                 entity_type=entity_type, license_id=license_id, user_id=user_id,
-                                language=language, exclude=exclude, sort=sort, limit=limit, offset=offset)
+                                language=language, exclude=exclude, sort=sort, limit=limit, offset=offset, review_type=review_type)
 
 
 def get_popular(limit=None):
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -383,23 +383,28 @@ def get_reviews_list(connection, *, inc_drafts=False, inc_hidden=False, entity_i
     if filterstr:
         filterstr = " WHERE " + filterstr
 
+    latest_revision_query = """
+        JOIN (
+            revision
+            JOIN (
+                SELECT review.id AS review_uuid,
+                       MAX(timestamp) AS latest_timestamp
+                  FROM review
+                  JOIN revision ON review.id = review_id
+              GROUP BY review.id
+            ) AS latest
+              ON latest.review_uuid = revision.review_id
+             AND latest.latest_timestamp = revision.timestamp
+            ) AS latest_revision
+          ON review.id = latest_revision.review_id
+    """
+
     query = sqlalchemy.text("""
         SELECT COUNT(*)
           FROM review
-           JOIN (
-                revision
-                JOIN (
-                    SELECT review.id AS review_uuid,
-                           MAX(timestamp) AS latest_timestamp
-                      FROM review
-                      JOIN revision ON review.id = review_id
-                  GROUP BY review.id
-                  ) AS latest
-                  ON latest.review_uuid = revision.review_id
-                 AND latest.latest_timestamp = revision.timestamp
-               ) AS latest_revision ON review.id = latest_revision.review_id
+            {latest_revision_query}
             {filterstr}
-        """.format(filterstr=filterstr))
+        """.format(filterstr=filterstr, latest_revision_query=latest_revision_query))
 
     result = connection.execute(query, filter_data)
     count = result.fetchone()[0]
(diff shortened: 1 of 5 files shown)


alastair:
@brainzbot test this please

critiquebrainz/ws/review/views.py
```diff
@@ -387,6 +389,12 @@ def review_list_handler():
             'reviews': reviews,
             'count': count,
         }, namespace=REVIEW_CACHE_NAMESPACE)
+    if review_type:
```
alastair, line 392: I think that we should pass this parameter into `db_review.list_reviews`, instead of post-filtering it in python.

Keep in mind that reviews can contain either a text review, or a numerical rating, or both. This means that if you want to filter ratings, we should explicitly check `where rating is not null`. Note that in your check here, you would remove all items that have a rating of 0!

critiquebrainz/ws/review/views_test.py
```diff
@@ -5,7 +5,7 @@
 import critiquebrainz.db.users as db_users
 from critiquebrainz.db.user import User
 from critiquebrainz.ws.testing import WebServiceTestCase
-
+import logging
```
alastair, line 8: not needed

critiquebrainz/ws/review/views.py
```diff
@@ -337,6 +337,7 @@ def review_list_handler():
     :query limit: results limit, min is 0, max is 50, default is 50 **(optional)**
     :query offset: result offset, default is 0 **(optional)**
     :query language: language code (ISO 639-1) **(optional)**
+    :query review_type: ``all`` or ``text`` or ``rating`` **(optional)**
```
alastair, line 340: I think that we can be a bit more descriptive in this documentation. At the moment, it's unclear when it should be used. How about something like:

    :query review_type: ``review`` or ``rating`` (optional). If set, only return reviews which have a text review, or a rating.

I think that we can remove the `all` option. This is the same as not adding the parameter. I also think that we should rename text-> review

Check also the documentation for "language code", I think this should be a little bit clearer to explain that it will filter only reviews with the given language code.
See also the documentation for "entity_id" - this no longer refers to a release group.
The documentation for entity_type should have a link to a list in the documentation that shows a list of all reviewable entities. We do this in listenbrainz:
https://github.com/metabrainz/listenbrainz-server/blob/e39c2f27e19867fa1293612f5b4c9b9e90b01f62/docs/dev/json.rst#L35
https://github.com/metabrainz/listenbrainz-server/blob/684e28863f209baa79c3e7e571b3f56b94f9b62f/docs/dev/api.rst#L119
(I think that github won't link you to the correct line for these files, but you can open them up yourself and go to that line)

alastair (changes requested):
Thanks for this contribution! I made some suggestions to improve the way that we select these reviews from the database

critiquebrainz/db/review.py
```diff
@@ -487,12 +487,18 @@ def get_reviews_list(connection, *, inc_drafts=False, inc_hidden=False, entity_i
                 "created": row.pop("user_created"),
             })
 
+        if review_type == 'rating':
```
alastair, line 490: sorry - looking back at my original comment I realise that I wasn't very clear here.
What I meant to say by moving this parameter into `get_reviews_list`, is that we should perform this filtering in the SQL query above (by adding items to `filters`/`filter_data`), instead of performing a post-filtering in python.

critiquebrainz/ws/review/views.py
```diff
@@ -337,6 +337,7 @@ def review_list_handler():
     :query limit: results limit, min is 0, max is 50, default is 50 **(optional)**
     :query offset: result offset, default is 0 **(optional)**
     :query language: language code (ISO 639-1) **(optional)**
+    :query review_type: ``review`` or ``rating`` **(optional)**
```
alastair, line 340: Thanks for updating the documentation for this parameter. I also added some descriptive text in the change that I suggested. We should add this text to this documentation to explain what the parameter does.
I also suggested some improvements to some other parameter documentation for this function that we should improve.

alastair:
notify @reosarevok on release

alastair:
I had a quick look at the failing tests here. It looks like there are two queries that are executed here. Your modification works on one of them, but not the second.

https://github.com/metabrainz/critiquebrainz/blob/a82c7d7a66d51802dc7765be1658cefaae488f65/critiquebrainz/db/review.py#L385-L389
This query only uses the `review` table, so we're unable to filter on the review text/rating in the revision table. The second query in this method joins to the revision table just fine, so that filter should work.

The reason why we do these two queries is because we need two pieces of information:
1. The _total_ number of reviews written for this entity
2. The current list of reviews to return to the API

Consider the situation where there are 174 reviews and a user makes an API request. The maximum number of reviews that you can get in the API is 50, so the response is going to say `{'count': 174, reviews: [list of first 50 reviews]}`
If we filter to only reviews with ratings, there may only be 62 of them. This means that the response will be `{'count': 62, reviews: [list of first 50 reviews]}`

Because of this, we can't count the reviews just by getting `len(rows)` of the second query. This would be correct if there were <=50 reviews, but not if there were more. This bug was also present in the initial version that you made, so it's good that we caught it.

I think that the only situation to this is to join against the `revision` table in both queries, as we currently do in the second query. We should make sure that we don't duplicate this query - I think it would be fine to extract it out to a variable and substitute it into the query.

critiquebrainz/ws/review/views.py
amCap1712: Done

amCap1712:
I applied the remaining points from feedback. Also, found a bug due to review_type being included in cache key so added it there.

critiquebrainz/ws/review/test/views_test.py
```diff
@@ -72,6 +72,61 @@ def test_review_delete(self):
...
+        expected_review_ids = [str(review_type_all['id']), str(review_only_rating['id'])]
+        self.assertCountEqual(actual_review_ids, expected_review_ids)
+
+        response = self.client.get('/review/', query_string={'review_type': 'review'})
+        self.assert200(response)
+        actual_review_ids = [review['id'] for review in response.json['reviews']]
+        expected_review_ids = [str(review_type_all['id']), str(review_only_review['id'])]
+        self.assertCountEqual(actual_review_ids, expected_review_ids)
+
+    def test_review_large_count(self):
```
alastair, line 114: we should add a comment to explain what this is testing (that `count` is the total number of reviews with this given review_type, which may be different from the number of returned items on this page).
In order to be explicit it'd be good to add 1 rating review here, and check that without a param there are 101 reviews, and with review_type=review there are 100

critiquebrainz/db/review.py
```diff
@@ -511,6 +525,8 @@ def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None, entity_t
         inc_drafts (bool): True if reviews marked as drafts should be included, False if not.
         inc_hidden (bool): True if reviews marked as hidden should be included, False if not.
         exclude (list): List of reviews (their IDs) to exclude from results.
+        review_type (str): Filter reviews. Can either be "review"(for reviews with only text), or "rating" (for
+                           reviews with only rating), or "None" (for all reviews).
```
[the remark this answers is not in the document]
alastair: be careful about the meaning of this description: "(for reviews with only text)" could mean that it doesn't include reviews which have text and a rating (although in our implementation it _does_ return this)

critiquebrainz/ws/review/views.py
```diff
@@ -335,13 +335,14 @@ def review_list_handler():
...
-    :json uuid entity_id: UUID of the release group that is being reviewed
-    :json string entity_type: One of the supported reviewable entities. 'release_group' or 'event' etc. **(optional)**
+    :json uuid entity_id: UUID of the entity that is being reviewed
+    :json string entity_type: One of the supported reviewable entities. :data:`critiquebrainz.db.review.ENTITY_TYPES` **(optional)**
     :query user_id: user's UUID **(optional)**
     :query sort: ``popularity`` or ``published_on`` **(optional)**
     :query limit: results limit, min is 0, max is 50, default is 50 **(optional)**
     :query offset: result offset, default is 0 **(optional)**
     :query language: language code (ISO 639-1) **(optional)**
+    :query review_type: ``review`` or ``rating`` **(optional)**
```
alastair, line 345: this doc item should also include a description of what this parameter does

PR #319, opened 2020-10-17
CB-367: Review page doesn't link to the detail page of the entity. 

AbhinavOhri:
CB-367

### DIFF
--- a/critiquebrainz/frontend/templates/review/modify/artist.html
+++ b/critiquebrainz/frontend/templates/review/modify/artist.html
@@ -1,8 +1,8 @@
 <div class="col-md-12">
   <dl class="dl-horizontal">
     <dt>{{ _('Artist') }}</dt>
-    <dd>     
-      <a href="{{ url_for('artist.entity', mbid=review.entity['id']) }}" target="_blank">{{ entity['name'] }}</a>
+    <dd>
+      <a href="{{ url_for('artist.entity', mbid=entity['id']) }}" target="_blank">{{ entity['name'] }}</a>
     </dd>
     <dt>{{ _('Type') }}</dt>
     <dd>{{ entity['type'] or '-' }}</dd>
--- a/critiquebrainz/frontend/templates/review/modify/artist.html
+++ b/critiquebrainz/frontend/templates/review/modify/artist.html
@@ -1,11 +1,8 @@
-{% if entity is not defined %}
-  {% set entity = entity_id | entity_details(type=entity_type) %}
-{% endif %}
 <div class="col-md-12">
   <dl class="dl-horizontal">
     <dt>{{ _('Artist') }}</dt>
-    <dd>
-      <a href="{{ url_for('artist.entity', mbid=entity_id) }}" target="_blank">{{entity['name'] | default(_('[Unknown artist]')) }}</a>
+    <dd>     
+      <a href="{{ url_for('artist.entity', mbid=review.entity['id']) }}" target="_blank">{{ entity['name'] }}</a>
     </dd>
     <dt>{{ _('Type') }}</dt>
     <dd>{{ entity['type'] or '-' }}</dd>
--- a/critiquebrainz/frontend/templates/review/modify/artist.html
+++ b/critiquebrainz/frontend/templates/review/modify/artist.html
@@ -2,7 +2,7 @@
   <dl class="dl-horizontal">
     <dt>{{ _('Artist') }}</dt>
     <dd>
-      <a href="{{ url_for('artist.entity', mbid=entity['id']) }}" target="_blank">{{ entity['name'] }}</a>
+      <a href="{{ url_for('artist.entity', id=entity['id']) }}" target="_blank">{{ entity['name'] }}</a>
     </dd>
     <dt>{{ _('Type') }}</dt>
     <dd>{{ entity['type'] or '-' }}</dd>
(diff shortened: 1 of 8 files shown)


alastair:
@brainzbot test this please

critiquebrainz/frontend/templates/review/modify/artist.html
```diff
@@ -1,11 +1,11 @@
 {% if entity is not defined %}
-  {% set entity = review.entity_id | entity_details(type=entity_type) %}
```
alastair, line 2: I'm a little bit unsure why we have this check at the beginning of all of these .html files (the `if entity is not defined` check). It seems like these files are imported [from here](https://github.com/metabrainz/critiquebrainz/blob/a60a4731e1ff28327b0ff4ccf871cbe3b8f83d9a/critiquebrainz/frontend/templates/review/modify/base.html#L14), and this file is rendered from `frontend/views/review.py`, and the `entity` context variable is added here.

I recommend that we remove these 3 lines from these template files. Further below in the file when you need the entity_id, can we just use `entity.id`

critiquebrainz/frontend/templates/review/modify/artist.html
```diff
@@ -1,11 +1,11 @@
...
+  {% set entity = entity_id | entity_details(type=entity_type) %}
 {% endif %}
 <div class="col-md-12">
   <dl class="dl-horizontal">
     <dt>{{ _('Artist') }}</dt>
     <dd>
-      {{ entity['name'] | default(_('[Unknown artist]')) }}
+      <a href="{{ url_for('artist.entity', mbid=entity_id) }}" target="_blank">{{entity['name'] | default(_('[Unknown artist]')) }}</a>
```
alastair, line 8: I'm not sure why we have this `default` value in these descriptions. In this file, I believe that we would always have a name, so I recommend that we remove it from all of these files

critiquebrainz/frontend/templates/review/modify/work.html
```diff
@@ -6,6 +6,7 @@
     <dt>{{ _('Work') }}</dt>
     <dd>
       {{ entity['name'] | default(_('[Unknown work]')) }}
+         <a href="{{ url_for('work.entity', id=entity_id) }}" target="_blank">{{entity['name'] | default(_('[Unknown work]')) }}</a>
```
alastair, line 9: It looks like you've added this href but not deleted the text version above

critiquebrainz/frontend/views/review.py
```diff
@@ -278,7 +278,7 @@ def create(entity_type=None, entity_id=None):
         if not form.errors:
             flash.info(gettext("Please provide some text or a rating for this review."))
         return render_template('review/modify/write.html', form=form, entity_type=entity_type, entity=entity,
-                               spotify_mappings=spotify_mappings, soundcloud_url=soundcloud_url)
+                               entity_id=entity_id, spotify_mappings=spotify_mappings, soundcloud_url=soundcloud_url)
```
alastair, line 281: If we use `entity.id` in the templates, we don't need to pass this into the template context

alastair (changes requested):
Thanks for these changes. 
One thing that we like to do when making changes to these projects is to make sure that we leave the code in a better quality than when we started. This could mean things like improving tests, or removing code that's no longer required. I added some suggestions for things that we could do to improve the structure of this code to make this better.

critiquebrainz/frontend/templates/review/modify/artist.html
```diff
@@ -1,11 +1,8 @@
-{% if entity is not defined %}
-  {% set entity = review.entity_id | entity_details(type=entity_type) %}
-{% endif %}
 <div class="col-md-12">
   <dl class="dl-horizontal">
     <dt>{{ _('Artist') }}</dt>
-    <dd>
-      {{ entity['name'] | default(_('[Unknown artist]')) }}
+    <dd>     
+      <a href="{{ url_for('artist.entity', mbid=review.entity['id']) }}" target="_blank">{{ entity['name'] }}</a>
```
alastair, line 5: I think this is an error here, it should be `mbid=entity['id']`, not `review.entity`.

Be careful with the line above, too - You can see in the diff view that you've added some whitespace after the `dd` tag

critiquebrainz/frontend/views/review.py
```diff
@@ -289,7 +289,7 @@ def create(entity_type=None, entity_id=None):
     if not form.errors:
         flash.info(gettext("Please provide some text or a rating for this review."))
     return render_template('review/modify/write.html', form=form, entity_type=entity_type,
-                           entity_title=entity_title, entity=entity)
+                           entity_title=entity_title, entity=entity, entity_id=entity_id)
```
alastair, line 292: you can remove `entity_id=entity_id` from the template context, we don't need this in the template file.

alastair (commented):
This is much better, thanks. Just a few small things to update

critiquebrainz/frontend/templates/review/modify/artist.html
alastair: I found out why this `default` is set, it's in case the item no longer exists in musicbrainz (CB-400). However in this specific situation I don't think there's a problem with removing this. Thanks!

alastair:
Thanks for making these changes. I see one more display issue in one of the templates. Can you double-check all of them that the show the link properly?

AbhinavOhri:
@alastair Tests for this PR are passing offline but not online. Please take a look.

alastair:
@brainzbot test this please

alastair:
@AbhinavOhri I found the failing tests, and fixed them in #322, we can merge those fixes here once we approve that PR.

How are you running the tests? I realised that `./develop.sh test` may give the wrong results sometimes. Try with 

    ./test.sh -d
    ./test.sh -b
    ./test.sh

and see if you get the failures

AbhinavOhri:
https://gist.github.com/AbhinavOhri/020de79a735721b22cee423bbd479abf

amCap1712:
Updated the PR. Did some reformatting and added a missing link.

PR #320, opened 2020-12-02
CB-330 Removed duplicate flash messages

AbhinavOhri:
[CB-330] 'Review has been hidden' message is shown twice when moderator hides a review.

### DIFF
--- a/critiquebrainz/frontend/templates/log/action.html
+++ b/critiquebrainz/frontend/templates/log/action.html
@@ -1,7 +1,7 @@
 {% extends 'base.html' %}
 
 {% if review is defined %}
-  {% set release_group = review.entity_id | entity_details %}
+  {% set release_group = review.entity_id | entity_details(entity_type=review.entity_type) %}
   {% set rg_title = release_group.title | default(_('[Unknown release group]')) %}
   {% set cancel_url = url_for('review.entity', id=review.id) %}
 {% elif user is defined %}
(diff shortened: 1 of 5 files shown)


alastair:
Thanks!
We should have a look at the workflows that can cause these flash messages to appear. I can think of three scenarios

1. A regular user tries to visit a review that has been hidden. Personally I think that we should improve this workflow. If it's the user's review then we can send a message that it's hidden. If not, we should return a generic "Review not found" message
2. An admin user marks a review as hidden, the process makes a flash message and then redirects the admin user to the review again where another message is flashed
3. An admin user visits the review directly using the link. With the change that you've made, a message no longer appears but it's important that it does.

I think that it would be better to remove the other message that is here, instead so that scenario 3. still works: https://github.com/metabrainz/critiquebrainz/blob/2d319ad45e556e5ca3764a6fa1f956e455cc8289/critiquebrainz/frontend/views/review.py#L466

I see that there is one test to check that the message shows when an admin hides a review, but I think that we should also add tests for the scenarios that I mentioned in 1. and 3.

AbhinavOhri:
I have updated the PR with the requested changes.

alastair:
Thanks for making this change!

I also suggested that we make sure that we have tests for each of these three scenarios that I mentioned. Can you check if there are already tests for these cases, and if not add one?

PR #322, opened 2020-12-02
Update tests for unknown_entities_for_missing=True fix

alastair:
In 47a6164bb0 (#304) we added `unknown_entities_for_missing=True` to calls to the BU musicbrainz db methods, but for some reason forgot to update the tests.
Also missing was this flag on Label, add it.


amCap1712 (approved):
Awesome! Thanks for the fix!

PR #323, opened 2020-12-03
Address pylint and flake8 messages

alastair:


alastair:
@amCap1712 I added two more exceptions to pylintrc here. Let me know if you think that they're OK.

no-self-use
https://github.com/metabrainz/critiquebrainz/blob/3bfc296cb07d861ece34130638890870d6170bb9/critiquebrainz/frontend/external/musicbrainz_db/test/test_cache.py#L16-L31

We have no `self.whatever` in this method, but I don't want to change the consistency of these classes just because of that. I'm happy to disable this one


no-else-return
```
if foo == 1:
    return x
elif foo == 2:
    return y
else:
    return z
```
I like the pattern of this, even though it could be rewritten to replace `elif`->`if` and eliminate the final `else`. I think that the pattern of this code makes it more explicit that only one of these items is possible.

alastair:
After discussion with @amCap1712 we decided that we should disable `no-self-use` on a per-file basis in tests only when it's strictly necessary.

PR #326, opened 2020-12-22
CB-397: Average ratings macro has bad pluralisation

dravadhis:


alastair:
@brainzbot add to whitelist

alastair:
Thanks for the contribution to CB!

In CritiqueBrainz, we use a tool called gettext which allows us to easily translate between many different languages. If you don't know about it, you can read more information here: https://en.wikipedia.org/wiki/Gettext

When working with text that has a different pluralisation depending on a count, gettext has a method called `ngettext` specifically for this case. It allows you to specify different text to output if there is only one item (i.e. "1 review") compared to what to write if there are 0 or more than 1 item (0 reviews, 5 reviews), and then gives translators specific examples for them to fill in the same thing in a different language.

Take a look at the flask documentation for translations: https://jinja.palletsprojects.com/en/2.11.x/templates/#i18n-in-templates and specifically look at the ngettext example: `{{ ngettext('%(num)d apple', '%(num)d apples', apples|count) }}`. There is one other example of ngettext in CB that you can search for if you need another example.

We should use ngettext in this case, but make sure that you only use it in the text that needs to be translated. You duplicated some html code which is the same in both cases, so we should leave that outside the check.

dravadhis:
Unable to work on this. Some other work came up. Hence closing.

alastair:
no problem, thanks for the contribution. It looks pretty good now so I'm going to reopen it and get it tested for merging - thanks!

alastair:
I made some changes to this and merged it: https://github.com/metabrainz/critiquebrainz/commit/c19ab98c9b3269676d229fed84151f8fb203237f Thanks!

PR #327, opened 2020-12-29
 Shouldn't have to chose a license just to save drafts

AbhinavOhri:
I created a new function in order to do the task but i am facing an error.
https://gist.github.com/AbhinavOhri/0826401900485e980aac77bd9b169a86

According to me we need to add a separate license for the 'save draft' option or make the license id key optional.

### DIFF
--- a/critiquebrainz/frontend/forms/review.py
+++ b/critiquebrainz/frontend/forms/review.py
@@ -19,6 +19,15 @@ class StateAndLength(validators.Length):
             raise ValidationError(self.message)
 
 
+def optional_license(message):
+    def _license(form, field):
+        if form.state.data == "draft":
+            return
+        else:
+            if not field.raw_data or not field.raw_data[0]: #https://wtforms.readthedocs.io/en/2.3.x/_modules/wtforms/validators/#InputRequired
+                raise ValidationError(message)
+    return _license
+
 # Loading supported languages
 languages = []
 for language_code in supported_languages:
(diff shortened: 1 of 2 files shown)


alastair:
Thanks for looking in to this. You're right that we'll have to do something about what to do when the values are null. I had a look into some options and think that we can do the following:

Remove the not-null check from review.license_id, but add a check that ensures that it's always set if the review has been published:
```
ALTER TABLE review ADD CONSTRAINT license_id_not_null_is_draft_false
    CHECK ((is_draft = 'f' AND license_id IS NOT NULL) OR is_draft = 't');

ALTER TABLE review ALTER COLUMN license_id DROP NOT NULL;
```

Change the `review_license_id_fkey` foreign key (in create_foreign_keys.sql) to chance ON DELETE CASCADE to ON DELETE SET NULL.

These 3 changes should be added to a data migration file (e.g. admin/schema_changes/20.sql), and the changes should also be made to the files in admin/sql.

We also need to make some other code changes:
 * In [review.create](https://github.com/metabrainz/critiquebrainz/blob/2b08a6deafbd3015216071976fa9ca95d4fd3263/critiquebrainz/db/review.py#L306) we need to only set the `license_id` value if it's not None. That is, in the string `INSERT INTO review (fields, here)` the fields will have to change depending on if we want to add license_id or not
 * In [review.get_by_id](https://github.com/metabrainz/critiquebrainz/blob/2b08a6deafbd3015216071976fa9ca95d4fd3263/critiquebrainz/db/review.py#L109) and [review.get_reviews_list](https://github.com/metabrainz/critiquebrainz/blob/2b08a6deafbd3015216071976fa9ca95d4fd3263/critiquebrainz/db/review.py#L442) we need to do a left outer join on the license table because the value could now be null
 * We will need to update the [schema version](https://github.com/metabrainz/critiquebrainz/blob/ff5416b761bb41b1f3aacf4eb715b19fe24ba57b/critiquebrainz/db/__init__.py#L5) used in dumps, because any dumps made after this change will be incompatible with the database structure before-hand

On top of these changes, we should ensure that we have tests in place that try to write a review, and ensure that
* License can be blank if saving as draft
* You can edit a draft that has no license set
* License is required if the review is being published

PR #328, opened 2020-12-30
CB-392-After Saving draft, cannot publish without making some change.

AbhinavOhri:
CB-392
I added an extra condition to check whether it is a draft while checking if the user has made any updates in its review.

### DIFF
--- a/critiquebrainz/frontend/views/review.py
+++ b/critiquebrainz/frontend/views/review.py
@@ -357,12 +357,12 @@ def edit(id):
         # Can't change license if review is published.
         del form.license_choice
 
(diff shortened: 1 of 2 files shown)


alastair:
This looks like a nice and simple change, thanks.

We should add a test that ensures that this functionality works. Specifically, we should have:
* A test that checks that you can load a draft review, press the publish button without making any changes, and have the review published
* A test that loads a draft review, makes no changes, and presses the save as draft button. If this happens, the warning message should still appear. We do this because we don't want to create two identical revisions of a draft review.

PR #329, opened 2020-12-31
CB-399 Title doesn't show on edit page for some reviews

AbhinavOhri:
CB-399


amCap1712:
Did some variable renaming and tested on beta.

critiquebrainz/frontend/views/review.py
```diff
@@ -67,6 +67,16 @@ def browse():
 
 # TODO(psolanki): Refactor this function to remove PyLint warning.
 # pylint: disable=too-many-branches
```
alastair, line 69: is this pylint-disable from the `def entity` function? If injecting this function into the middle of the declaration/function doesn't change the results of pylint then we can remove the comment

PR #330, opened 2021-01-02
Fix bug in initialising CB test database

amCap1712:
The flask app created before creating the database tables checks whether the database is being created for tests and uses the config file accordingly. However, after that we again create the flask app but do not check whether test configuration should be used. Therefore, the SQLALCHEMY_DATABASE_URI gets set incorrectly. This is evident from failure to insert license in the license table just after. I do not see why we need to create the app twice, hence remove the second creation and fix the bug.


alastair:
Before we used docker and docker-compose, we had separate test configs and a separate test database. Now that we run tests isolated in a separate docker-compose project, we can use the same config file and same database connect string, because they refer to separate databases.
Do you want to give a go at removing test-specific config/db code from here?

amCap1712:
Sure, I'll do that.

amCap1712:
Failing tests are due to WTF_CSRF_ENABLED. currently needs to be False in tests and probably True in default config.

PR #331, opened 2021-01-02
CB-401: entity pages for a non-existent MBID return HTTP 200 and show a detail page.

AbhinavOhri:
Fix bug for wrong import of 'NoDataException'.

Changed the value of "unknown-entities_for_missing" from True to False in the "get_artist_by_id" function which passed the appropiate exception when the artist was not found.
The exceptions of this function were caught using "critiquebrainz.frontend.external.musicbrainz_db.exceptions" but it gave the "Internal server error" . We had to use the exceptions from " brainzutils.musicbrainz_db.exceptions" in order to get the "page not found error ".

### DIFF
--- a/critiquebrainz/frontend/external/musicbrainz_db/artist.py
+++ b/critiquebrainz/frontend/external/musicbrainz_db/artist.py
@@ -19,7 +19,7 @@ def get_artist_by_id(mbid):
         artist = db.fetch_multiple_artists(
             [mbid],
             includes=['artist-rels', 'url-rels'],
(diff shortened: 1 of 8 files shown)


alastair:
This fixes the case of Artist, but doesn't address any of the other entities in CB. However, I thought of a reason why this functionality would have been present, and opened CB-408 to discuss the issue further before we continue with this PR.

alastair:
I'm going to close this, as we had a discussion in CB-408 for a more detailed solution to this problem.

PR #336, opened 2021-01-21
Prepare for consul upgrade

amCap1712:
The changes here follow the changes made for the [ListenBrainz setup](https://github.com/metabrainz/listenbrainz-server/pull/1237).

As discussed with @alastair, it is desirable to run cron jobs in a separate container. This may require additional changes during deployment,


Dockerfile
```diff
@@ -60,21 +60,24 @@ RUN useradd --create-home --shell /bin/bash critiquebrainz
 
 # Consul Template service is already set up with the base image.
 # Just need to copy the configuration.
-COPY ./docker/prod/consul-template.conf /etc/consul-template.conf
+COPY ./docker/consul-template.conf /etc/consul-template.conf
```
alastair, line 63: We don't need this comment any more, it's related to the old version. The file doesn't need to be copied to this location anymore either. We should copy it to `/etc/consul-template-uwsgi.conf` for uwsgi, and a separate file `/etc/consul-template-cron.conf` for cron. The cron job needs no `exec{}` block. We just use it to render the config file in the cron container.

Dockerfile
```diff
@@ -60,21 +60,24 @@ RUN useradd --create-home --shell /bin/bash critiquebrainz
...
 
 # cron jobs
 ADD ./docker/prod/cron/jobs /tmp/crontab
 RUN chmod 0644 /tmp/crontab && crontab -u critiquebrainz /tmp/crontab
 RUN rm /tmp/crontab
 RUN touch /var/log/dump_backup.log /var/log/public_dump_create.log /var/log/json_dump_create.log \
     && chown critiquebrainz:critiquebrainz /var/log/dump_backup.log /var/log/public_dump_create.log /var/log/json_dump_create.log
-
-# Make sure the cron service doesn't start automagically
-# http://smarden.org/runit/runsv.8.html
```
alastair, line 77: cron is now a bit more complex, we have to do two things:

 - add down to /etc/service/cron/down
 - add a new service (call it cron-config), which calls `run-consul-template` and generates the config files necessary for cron. This isn't in LB yet.
 - When a cron container starts up, we'll activate both of these services. One to make the config and another to start cron (we don't combine them together because we don't want consul to restart cron if it's in the middle of a task)

docker/consul-template.conf
```diff
@@ -1,9 +1,15 @@
...
     source = "/code/consul_config.py.ctmpl"
     destination = "/code/consul_config.py"
-    command = "sv hup uwsgi"
 }
 template {
     source = "/code/admin/config.sh.ctmpl"
     destination = "/code/admin/config.sh"
 }
+exec {
+    command = "sv hup uwsgi"
```
alastair, line 10: this should be `uwsgi /etc/uwsgi/uwsgi.ini` instead of sv hup (the command that used to be in uwsgi/uwsgi.service)

alastair (commented):
Looks good. A few small changes to make with the location of consul-template.conf and cron setup.

alastair:
Thanks for your work on this! I made a few small changes, and also fixed spelling mistake uswgi -> uwsgi :)

PR #341, opened 2021-02-08
Pin package versions and upgrade brainzutils

amCap1712:
Once metabrainz/brainzutils-python#45 is merged, BU will use open ended versions. To make sure we use desired versions, pin the versions in downstream projects.

The BU PR must be merged before this PR and the requirements.txt must be updated to use the new release of BU.


amCap1712:
@brainzbot retest this please

PR #344, opened 2021-02-12
Invalidate individual cache keys instead of the entire namespace

amCap1712:
`cache.invalidate_namespace` invalidates entire cache namespace. This wastes resources since the entire cache gets invalidated whenever a single review is updated or created. Therefore, we track affected cache entities individually and invalidate them as required.


critiquebrainz/db/review.py
```diff
@@ -522,21 +540,19 @@ def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None, entity_t
                                 language=language, exclude=exclude, sort=sort, limit=limit, offset=offset)
 
 
-def get_popular(limit=None):
+def get_popular_reviews_for_index():
     """Get a list of popular reviews.
 
     Popularity is determined by 'popularity' of a particular review. popularity is a
```
alastair, line 546: not directly related to this change, but this is a weird comment. "popularity is measured by getting the popularity".
We could probably delete this whole first sentence.

critiquebrainz/db/test/statistics_test.py
```diff
@@ -162,11 +162,10 @@ def test_get_top_users_overall(self, cache_set, cache_get):
         # get list of top users
         top_users = db_statistics.get_top_users_overall()
         expected_key = b"top_users_overall_cb_statistics"
-        cache_get.assert_called_once_with(expected_key,
-                                          db_statistics._CACHE_NAMESPACE)
-        cache_set.assert_called_once_with(key=expected_key, val={"users": top_users},
-                                          namespace=db_statistics._CACHE_NAMESPACE,
-                                          time=db_statistics._DEFAULT_CACHE_EXPIRATION)
+        cache_get.assert_called_with(expected_key, db_statistics._CACHE_NAMESPACE)
```
alastair, line 165: why does the namespace occur twice here?
https://github.com/metabrainz/critiquebrainz/blob/75ddff3569036bb7296097d0b61d90684581dd74/critiquebrainz/db/statistics.py#L238-L239
we use gen_key to make a key with "top_users_overall" + "cb_statistics", and then we do cache.get with a specific namespace anyway. Key will end up being "cb_statistics:top_users_overall_cb_statistics". Don't need the gen_key in statistics.py

critiquebrainz/ws/review/test/views_test.py
```diff
@@ -186,3 +190,33 @@ def test_revision_entity_handler(self):
...
         self.assert200(resp)
+
+    def test_cache_tracking(self):
+        entity_id = self.review["entity_id"]
+        track_key = cache.gen_key("ws_cache", entity_id)
+
+        # Test no cache if entity id is not provided
+        self.client.get('/review/', query_string={'sort': 'rating'})
+        cache_keys = cache.get(track_key, namespace="Review")
+        self.assertEqual(None, cache_keys)
```
alastair, line 201: What is this checking? I think it makes sense that there's no cache item that includes entity_id (from a few lines earlier). Does it create any cache items in this case? If it creates no cache items then I think we should instead mock cache set and ensure that it's not called. Otherwise I don't think that it makes too much sense to have this test here

critiquebrainz/ws/review/test/views_test.py
```diff
@@ -186,3 +190,33 @@ def test_revision_entity_handler(self):
...
+
+        expected_cache_keys = [b'list_6b3cd75d-7453-39f3-86c4-1441f360e121_None_popularity_50_0_None',
```
alastair, line 203: can we have at least one test that checks all of the items in this key (I'm looking at the two Nones)

critiquebrainz/ws/review/views.py
```diff
@@ -367,12 +369,17 @@ def review_list_handler():
 
     # TODO(roman): Ideally caching logic should live inside the model. Otherwise it
     # becomes hard to track all this stuff.
-    cache_key = cache.gen_key('list', entity_id, user_id, sort, limit, offset, language)
-    cached_result = cache.get(cache_key, REVIEW_CACHE_NAMESPACE)
+    cache_key = None
+    cached_result = None
+
+    # If entity_id is absent, it is not possible to track cache keys. Hence, skip caching altogether in that case.
```
alastair, line 375: I recall talking to you about this when you made the change, but I can't remember now - why can't we do it if there is no entity_id?
Is it because we create the `cache.gen_key('ws_cache', entity_id)` key to contain a list of all caches related to this entity_id?
What's the behaviour of this view if there is no entity_id? Is it a list of many different reviews? Why can't we have a key `ws_cache_None` (or even special-cased to `ws_cache_listkeys` or similar)  that contains the keys for this  non-id list view?

critiquebrainz/ws/review/views.py
```diff
@@ -384,10 +391,27 @@ def review_list_handler():
...
+            # cache keys of that entity. We can retrieve this list at time during invalidate and invalidate all caches
+            # for that entity_id.
+
+            track_cache_key = cache.gen_key('ws_cache', entity_id)
+            review_cache_keys = cache.get(track_cache_key, REVIEW_CACHE_NAMESPACE)
+            if review_cache_keys is None:
+                review_cache_keys = []
+            review_cache_keys.append(cache_key)
+            logging.error(track_cache_key, review_cache_keys)
+            cache.set(track_cache_key, review_cache_keys, namespace=REVIEW_CACHE_NAMESPACE, time=REVIEW_CACHE_TIMEOUT)
```
alastair, line 414: can we use a redis set here? it'd make it just one line to add the new key to the set.

critiquebrainz/ws/review/views.py
```diff
@@ -384,10 +391,27 @@ def review_list_handler():
...
+            }, namespace=REVIEW_CACHE_NAMESPACE, time=REVIEW_CACHE_TIMEOUT)
+
+            # When reviews for an entity get created/updated, the existing cache for it need to be cleared. The keys
+            # used to cache reviews include parameters like user_id, sort, limit so on which are unavailable later.
+            # Hence, making it impossible to generate the cache keys at time of invalidation. To circumvent this issue,
+            # we track all cache keys for an entity_id. Cache keys of the form ws_cache_{entity_id} point to a list of
+            # cache keys of that entity. We can retrieve this list at time during invalidate and invalidate all caches
+            # for that entity_id.
+
+            track_cache_key = cache.gen_key('ws_cache', entity_id)
```
alastair, line 408: the use of 'track' in this name confused me for a while. It's for tracking, not for musicbrainz tracks, right?
Let's call something like `cache_keys_for_entity_id_key`

critiquebrainz/ws/review/views.py
```diff
@@ -384,10 +391,27 @@ def review_list_handler():
...
+                'reviews': reviews,
+                'count': count,
+            }, namespace=REVIEW_CACHE_NAMESPACE, time=REVIEW_CACHE_TIMEOUT)
+
+            # When reviews for an entity get created/updated, the existing cache for it need to be cleared. The keys
+            # used to cache reviews include parameters like user_id, sort, limit so on which are unavailable later.
+            # Hence, making it impossible to generate the cache keys at time of invalidation. To circumvent this issue,
+            # we track all cache keys for an entity_id. Cache keys of the form ws_cache_{entity_id} point to a list of
+            # cache keys of that entity. We can retrieve this list at time during invalidate and invalidate all caches
+            # for that entity_id.
```
alastair, line 406: It's good to add this comment, but I think we can improve it a bit.

When we cache the results of a request, we include (entity_id, user_id, sort, limit, offset, language) in the cache key.
When entity_id is edited, we need to expire all cache items for this entity.
To do this, we track all of the cache keys for the entity in a separate cache item, ws_cache_{entity_id}.
These keys are retrieved and all keys are expired in `invalidate_ws_entity_cache`.

alastair (commented):
This looks good in terms of its functionality but there are a bunch of places that I think we can make improvements. Let's split up the work.

critiquebrainz/ws/review/views.py
amCap1712: Yes, its for tracking cache keys. I'll rename it.

critiquebrainz/ws/review/views.py
amCap1712: `entity_id` is just another clause in the `WHERE`.
https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/db/review.py#L351-L353

The actual SQL is a few lines below. So the output depends on the rest of filters passed. Yeah, special casing is possible. I do not remember why we decided not to do that.

amCap1712: I guess before adding redis sets this would have been a bit difficult. With redis sets, its easier. We can probably do this now.

critiquebrainz/ws/review/test/views_test.py
amCap1712: This part was for testing that if a query does not have an entity_id, it does not cached. We are now reconsidering what to do in this case.

alastair:
Replaced by #404

PR #354, opened 2021-03-31
Upgrade brainzutils and sentry-sdk

amCap1712:


alastair:
LOG_EMAIL still in custom_config.py

alastair:
CB also uses the js raven library, which we should update too: https://github.com/metabrainz/critiquebrainz/blob/95c39be56a98658086247aeadddd3e161f117a58/critiquebrainz/frontend/templates/errors/base.html#L11-L19
We should also use this time to migrate from the two public and private config items to just one (the private dsn is no longer used in both frontend and backend):
https://github.com/metabrainz/critiquebrainz/blob/95c39be56a98658086247aeadddd3e161f117a58/consul_config.py.ctmpl#L12-L15

alastair (approved):
looks great!

PR #358, opened 2021-04-13
CB-409: Add name/link for artist reviews in user page

reosarevok:
### Fix CB-409

This was seemingly just forgotten when artist reviews were added.


alastair:
This was inadvertently implemented in https://github.com/metabrainz/critiquebrainz/pull/303 / https://github.com/metabrainz/critiquebrainz/commit/4cd1187a6e4a42fc6ac34db3b04be2ce6e015b62 and so is no longer needed!

PR #371, opened 2021-08-18
Fix post review API

jdaok:
This PR fixes a small global variable bug that caused the publish review endpoint to return an error when the review's "is_draft" was set to false.

### DIFF
--- a/critiquebrainz/ws/review/views.py
+++ b/critiquebrainz/ws/review/views.py
@@ -416,6 +416,7 @@ def review_post_handler(user):
     """
 
     def fetch_params():
+        global REVIEW_TEXT_MIN_LENGTH
         is_draft = Parser.bool('json', 'is_draft', optional=True) or False
         if is_draft:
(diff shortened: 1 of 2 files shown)


critiquebrainz/ws/review/views.py
```diff
@@ -416,6 +416,7 @@ def review_post_handler(user):
     """
 
     def fetch_params():
+        global REVIEW_TEXT_MIN_LENGTH
```
amCap1712, line 419: This fix looks incorrect to me. We should not be reassigning a global variable (especially which we want to asssume to be constant) while processing a request as it will be reflected in future requests. For example, a draft review may set to REVIEW_TEXT_MIN_LENGTH, then another review comes which is not a draft, since REVIEW_TEXT_MIN_LENGTH is None the minimum length check will not work as expected.

amCap1712:
I have added a new fix and a test. This should hopefully fix the issue.

PR #375, opened 2021-08-22
Fix PG client in CB - Attempt 2

amCap1712:
Do not know what the issue is with the existing Dockerfile but doing what we do in LB should hopefully fix it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,12 +15,13 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # PostgreSQL client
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -15,12 +15,13 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # PostgreSQL client
-RUN apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8
```
alastair, line 18: it's possible that this keyserver is no longer available:
```
$ host ha.pool.sks-keyservers.net
Host ha.pool.sks-keyservers.net not found: 3(NXDOMAIN)
```

PR #378, opened 2021-08-24
Fix CORS issues for some endpoints

jdaok:
This PR adds proper headers and an OPTIONS method to many of the API endpoints to CORS errors.

### DIFF
--- a/critiquebrainz/ws/review/views.py
+++ b/critiquebrainz/ws/review/views.py
@@ -185,7 +185,9 @@ def review_revision_entity_handler(review_id, rev):
     return jsonify(revision=revision)
 
 
-@review_bp.route('/<uuid:review_id>', methods=['DELETE', 'OPTIONS'])
+# don't need to add OPTIONS here because its already added
+# for this endpoint in review_entity_handler
+@review_bp.route('/<uuid:review_id>', methods=['DELETE'])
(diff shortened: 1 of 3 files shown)


critiquebrainz/ws/review/views.py
```diff
@@ -34,8 +34,8 @@ def get_review_or_404(review_id):
     return review
 
 
-@review_bp.route('/<uuid:review_id>', methods=['GET'])
-@crossdomain()
+@review_bp.route('/<uuid:review_id>', methods=['GET', 'OPTIONS'])
+@crossdomain(headers="Authorization, Content-Type")
```
amCap1712, line 38: The same url here is mapped to GET and OPTIONS here. Further, down to POST and OPTIONS so on. This can be problematic as OPTIONS has multiple bindings. I'll look into this to see what the best way to handle this.

amCap1712:
As mentioned in the commit message, OPTIONS method is needed for CORS prefetch requests. @crossdomain decorator needs OPTIONS in methods otherwise OPTIONS requests to the route will return 405 NOT ALLOWED.

However, we don't need to add OPTIONS to each `methods` list of a particular route i.e. if there is a function to handle GET and another to handle POST for a route, just adding to OPTIONS to the GET function is enough.

PR #380, opened 2021-08-30
Check if access_token is None before accessing it

amCap1712:
Should fix bunch of errors in sentry from accessing unavailable token (probably revoked by user?). 

https://sentry.metabrainz.org/metabrainz/critiquebrainz/issues/145691/

### DIFF
--- a/critiquebrainz/db/oauth_token.py
+++ b/critiquebrainz/db/oauth_token.py
@@ -108,7 +108,7 @@ def list_tokens(*, client_id=None, refresh_token=None, access_token=None, limit=
             OFFSET :offset
         """.format(where_clause=filterstr)), filter_data)
         rows = results.fetchall()
-    return [dict(row) for row in rows] if rows.rowcount else None
+    return [dict(row) for row in rows]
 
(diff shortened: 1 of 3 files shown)


critiquebrainz/db/oauth_token.py
```diff
@@ -108,7 +108,7 @@ def list_tokens(*, client_id=None, refresh_token=None, access_token=None, limit=
             OFFSET :offset
         """.format(where_clause=filterstr)), filter_data)
         rows = results.fetchall()
-    return [dict(row) for row in rows]
+    return [dict(row) for row in rows] if rows.rowcount else None
```
alastair, line 111: Given that this method is called `list_tokens` I think it makes more sense that it returns an empty list in the case that there are no items instead of None.
The check in fetch_token, `return tokens[0] if tokens else None` still makes sense in this case as `[]` will evaluate to False

amCap1712: Fixed.

alastair (approved):

PR #383, opened 2021-09-17
Add bulk get review by id endpoint

jdaok:
This PR adds an API endpoint to fetch multiple reviews by UUIDs in a single SQL query.
example: https://critiquebrainz.org/ws/1/review/b7575c23-13d5-4adc-ac09-2f55a647d3de,e4364ed2-a5db-4427-8456-ea7604b499ef


jdaok:
Hey @amCap1712, please let me know if you think this looks good and I can add the python tests.

critiquebrainz/ws/review/views.py
```diff
@@ -94,6 +102,80 @@ def review_entity_handler(review_id):
...
+
+    :resheader Content-Type: *application/json*
+    """
+    reviews = []
+    review_ids = list(review_ids.split(',')) # retrieve UUID's as list from URL parameter
+    for review_id in review_ids:
+        if not validate_uuid(review_id):
+            raise ParserError(review_id, 'not valid UUID')
+
+    reviews = get_reviews_or_404(review_ids)
```
alastair, line 169: In AcousticBrainz for a bulk lookup we do a few things:
* We return a dictionary of items where the key is the entity id, this allows users to immediately retrieve what they need from the result without iterating through it
* We accept "fuzzy" uuids, that is, they could be written with capital letters instead of lower-case, or could have - missing or in different places
* Because of accepting fuzzy uuids, we provide a mapping between the uuid that a user provided and the "canonical representation" that we use as the dictionary keys, so that clients can take the uuid that they provided and find the key in the results
* If a uuid doesn't exist in the database, we omit it from the results. If none of the items exist in the database then we return a valid response, but with no items (your behaviour here is to only return 404 if _none_ of the reviews exist)

The code that does this is here: https://github.com/metabrainz/acousticbrainz-server/blob/165a65655070893b1eee66748cd561494c9c106a/webserver/views/api/v1/core.py#L335-L350

critiquebrainz/ws/review/views.py
```diff
@@ -94,6 +102,80 @@ def review_entity_handler(review_id):
...
+        }
+
+    :statuscode 200: no error
+    :statuscode 400: invalid UUID format
+    :statuscode 404: review not found
+
+    :resheader Content-Type: *application/json*
+    """
+    reviews = []
+    review_ids = list(review_ids.split(',')) # retrieve UUID's as list from URL parameter
```
alastair, line 164: one bug that I recently fixed in AcousticBrainz is that empty items (i.e. a list that is `uuid,uuid,,uuid`) will try and validate `''` and fail. We can probably ignore these items.

alastair (changes requested):
Thanks for making these changes! I think there are a small handful of improvements that we can make in order to simplify the code and ensure that the endpoint is easy to use.

We should make sure that we have tests for the new methods that we've added.

PR #387, opened 2022-01-23
Permanent fix for certificate verification

akshaaatt:
Apparently, an encrypt cert has expired and we are unable to do the local setup. I think this is fair in terms of local setup and we can look for permanent solutions in the Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,12 @@
 FROM metabrainz/python:3.8-20210115
 
+# remove expired let's encrypt certificate and install new ones
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends ca-certificates \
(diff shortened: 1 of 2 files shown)


amCap1712:
Thanks for fixing this but I suggest we go for the permanent fix directly. You can make the same changes as https://github.com/metabrainz/acousticbrainz-server/commit/b1722e4791b7767a0d9f7e4e2f3d5843c3ad1d9e for the permanent fix. Also, CB has two Dockerfiles (the one used in dev which the PR already modifies and the [other one used for prod](https://github.com/metabrainz/critiquebrainz/blob/master/Dockerfile)) so it would be good to fix both while at it.

akshaaatt:
Thanks @amCap1712! Made the necessary updates.

amCap1712:
Thanks. LGTM

PR #388, opened 2022-01-23
Enhance README.md

akshaaatt:
We need to better include the about section, contribution section, and some important links.


amCap1712 (approved):
Thanks for the improvements!

PR #389, opened 2022-01-25
CB-421: Show user ratings on the profile page

anshg1214:
[CB-421](https://tickets.metabrainz.org/projects/CB/issues/CB-421) Added User ratings on the profile page.


critiquebrainz/frontend/templates/user/base.html
```diff
@@ -30,3 +30,8 @@ <h2 class="pull-left"><img class="avatar" src="{{ user.avatar }}&s=32" /> {{ use
     {% block profile_content %}{% endblock %}
   </div>
 {% endblock %}
+
+{% block scripts %}
+  {{ super() }}
+  <script src="{{ get_static_path('rating.js') }}"></script>
```
amCap1712, line 36: This script is not used in this file and user/info.html inherits from this template and does not use this script either. So I suggest to move this block to reviews.html which is the file actually needing the script.

critiquebrainz/frontend/messages.pot
```diff
@@ -1,14 +1,14 @@
 # Translations template for PROJECT.
-# Copyright (C) 2021 ORGANIZATION
+# Copyright (C) 2022 ORGANIZATION
```
amCap1712, line 2: I don't think there are translation updates to apply as only metadata appears to change. So I suggest to revert changes to the file.

amCap1712 (changes requested):
Hi! Thanks for working on this. The changes look good to me. Once the feedback is addressed, I'll merge the PR.

critiquebrainz/frontend/messages.pot
anshg1214: There was a new column `ratings` added, so I added it for translation. Also, due to the addition of code, line numbers had changed.  So, I updated the file.

critiquebrainz/frontend/templates/user/base.html
anshg1214: Sure, I'll do that.

critiquebrainz/frontend/messages.pot
amCap1712: Ah! This change is fine then.

amCap1712 (approved):
Thanks!

PR #390, opened 2022-01-29
CB-406: Extract pagination logic into common component

anshg1214:
[CB-406](https://tickets.metabrainz.org/projects/CB/issues/CB-406) 

- Extracted pagination into macros
- Improved pagination for easy navigation


critiquebrainz/frontend/templates/macros.html
```diff
@@ -220,3 +220,40 @@ <h4 style="margin-bottom:0;">{{ _('Rate this {}:'.format(entity_type_readable))
...
+          <li {{ "class=active" if current_page == total_pages-1 }}><a href="{{ url_for(page_info, page=(total_pages-1), **parameter) }}">{{ total_pages-1 }}</a></li>
+        {% endif %}
+
+        {% if (current_page < (total_pages - 4)) %}
+          <li class='disabled'><a>{{ '...' }}</a></li>
+        {% endif %}
+
+        <li {{ "class=active" if current_page == total_pages }}><a href="{{ url_for(page_info, page=total_pages, **parameter) }}">{{ total_pages }}</a></li>
+
+        <li class="next {{'disabled' if current_page-1 >= count//limit }}"><a aria-label="Next" href="{{ url_for(page_info, page=page+1, **parameter) }}">{{ _('Next') }} &rarr;</a></li>
```
amCap1712, line 255: count is not a parameter to this macro and some templates do not have this variable defined (artist template for example) so this is causing an error when loading artist page.

critiquebrainz/frontend/templates/macros.html
```diff
@@ -220,3 +220,40 @@ <h4 style="margin-bottom:0;">{{ _('Rate this {}:'.format(entity_type_readable))
...
+        <li class="previous {{'disabled' if current_page <= 1 }}"><a aria-label="Previous" href="{{ url_for(page_info, page=page-1, **parameter) }}">&larr; {{ _('Previous') }}</a></li>
+        
+        <li {{ "class=active" if current_page == 1 }}><a href="{{ url_for(page_info, page=1,  **parameter) }}">{{ 1 }}</a></li>
+
+        {% if current_page > 5 %}
+          <li class="disabled"><a>{{ '...' }}</a></li>
+        {% endif %}
+
+        {% if current_page == 5 %}
+          <li {{ "class=active" if current_page == 2 }}><a href="{{ url_for(page_info, page=2, **parameter) }}">{{ 2 }}</a></li>
```
amCap1712, line 238: current_page == 2 inside an if of current_page == 5 seems wrong.

critiquebrainz/frontend/templates/macros.html
```diff
@@ -220,3 +220,40 @@ <h4 style="margin-bottom:0;">{{ _('Rate this {}:'.format(entity_type_readable))
         -
     {% endif %}
 {% endmacro %}
+
+{% macro display_pagination(current_page, total_pages, page_info, parameter) %}
```
amCap1712, line 224: It looks like `count/limit|round(method='ceil')|int` theme is common in callers of this macro. I suggest you take count and limit as parameters to this macro. This way you can fold the common if check as well into here and also apply the modification at one place.

amCap1712 (changes requested):
Hi! Thanks for working on this and apologies for the late review.

The new pagination code is different from what we used originally. It looks like it allows to paginate among 5 pages at  a time. I guess this way is probably fine too but any particular reason you chose to implement it this way instead of the existing previous and next buttons?

anshg1214:
I chose to paginate 5 pages at a time because I felt that it is much better for easy navigation compared to the existing previous and next buttons. Also, I chose to paginate 5 pages mainly because it provides the best use. A smaller number would not add much utility, and any number greater than 5 would be too much for the user to navigate.

critiquebrainz/frontend/templates/macros.html
```diff
@@ -220,3 +220,41 @@ <h4 style="margin-bottom:0;">{{ _('Rate this {}:'.format(entity_type_readable))
         -
     {% endif %}
 {% endmacro %}
+
+{% macro display_pagination(current_page, count, limit, page_info, parameter) %}
```
amCap1712, line 224: You can change the parameter here to something like `**params` and invoke the macro normally like `{{ display_pagination(page, count, limit, 'review.browse', entity_type=entity_type) }}` which looks a bit cleaner.

critiquebrainz/frontend/templates/macros.html
```diff
@@ -220,3 +220,41 @@ <h4 style="margin-bottom:0;">{{ _('Rate this {}:'.format(entity_type_readable))
...
+        <li class="previous {{'disabled' if current_page <= 1 }}"><a aria-label="Previous" href="{{ url_for(page_info, page=page-1, **parameter) }}">&larr; {{ _('Previous') }}</a></li>
+        
+        <li {{ "class=active" if current_page == 1 }}><a href="{{ url_for(page_info, page=1,  **parameter) }}">{{ 1 }}</a></li>
+
+        {% if current_page > 5 %}
+          <li class="disabled"><a>{{ '...' }}</a></li>
+        {% endif %}
+
+        {% if current_page == 5 %}
+          <li><a href="{{ url_for(page_info, page=2, **parameter) }}">{{ 2 }}</a></li>
```
amCap1712, line 239: I don't think this special case `current_page == 5` and `current_page == (total_page - 4)` (below) are worth it. Let's remove this two special cases. As a result, you'll probably also need to update `>` and `<` to `>=` and `<=` in the `...` cases.

critiquebrainz/frontend/templates/macros.html
```diff
@@ -220,3 +220,41 @@ <h4 style="margin-bottom:0;">{{ _('Rate this {}:'.format(entity_type_readable))
...
+          <li><a href="{{ url_for(page_info, page=(total_pages-1), **parameter) }}">{{ total_pages-1 }}</a></li>
+        {% endif %}
+
+        {% if (current_page < (total_pages - 4)) %}
+          <li class='disabled'><a>{{ '...' }}</a></li>
+        {% endif %}
+
+        <li {{ "class=active" if current_page == total_pages }}><a href="{{ url_for(page_info, page=total_pages, **parameter) }}">{{ total_pages }}</a></li>
+
+        <li class="next {{'disabled' if (current_page-1) >= total_pages }}"><a aria-label="Next" href="{{ url_for(page_info, page=page+1, **parameter) }}">{{ _('Next') }} &rarr;</a></li>
```
amCap1712, line 256: While testing, I saw that if the user navigates to the last page. The next button is still enabled. This condition should probably be using `current_page` instead of `current_page - 1`.

critiquebrainz/frontend/templates/mapping/spotify.html
```diff
@@ -47,18 +47,7 @@ <h2>{{ _('Spotify album mapping') }}</h2>
     {% endfor %}
   </div>
   {% if count > limit %}
```
amCap1712, line 49: Let's move these `count > limit` (this and the ones inside other files modified in this PR) checks to inside the macro.

amCap1712 (changes requested):
Thanks for the updates. A few more changes and we should be good to go!

critiquebrainz/frontend/templates/macros.html
anshg1214: Ok, I'll change it.

critiquebrainz/frontend/templates/macros.html
anshg1214: I tried doing this, but when we have multiple parameters it does not work. Like in case of artists page, we pass 2 parameters, `id` and `release_type`. So I found this thing as the most flexible way.

amCap1712: Uh right, Sorry I had forgotten these are Jinja2 macros and not python functions. **kwargs do work with [Jinja2](https://stackoverflow.com/questions/13944751/args-kwargs-in-jinja2-macros) as use different magic. The current version is better than jinja's magic so this is fine in the current way. Thanks.

amCap1712 (approved):
Thanks!

PR #391, opened 2022-02-06
Remove unused code for Google Analytics and Google+

amCap1712:
- Google Analytics has been disabled for a long time (or was never enabled?) using configuration. Disabled here means that the code to render GA wasn't send to the browser. No reason to keep this unused code.

- Google+ was discontinued a few years so removing the sharing button for it. It does not render anything currently anyway.


alastair:
looks good 👍

PR #394, opened 2022-02-09
CB-347: Allow users to browse reviews by new, old and other criteria

anshg1214:
[CB-347](https://tickets.metabrainz.org/browse/CB-347) 

- Added sorting option on reviews page.
- Created new sort option `published_on_asc` in db.

### DIFF
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -579,7 +579,7 @@ def list_reviews(*, inc_drafts=False, inc_hidden=False, entity_id=None, entity_t
         entity_type (str): Type of the entity that has been reviewed.
         user_id (uuid): ID of the author.
         sort (str): Order of the returned reviews. Can be either "popularity" (order by difference in +/- votes),
-                    or "published_on" (order by publish time), or "random" (order randomly).
+                    "published_on" (order by publish time), "published_on_asc" (order by publish time ascending) or "random" (order randomly).
         limit (int): Maximum number of reviews to return.
         offset (int): Offset that can be used in conjunction with the limit.
         language (str): Language code of reviews.
--- a/critiquebrainz/db/review.py
+++ b/critiquebrainz/db/review.py
@@ -470,9 +470,11 @@ def get_reviews_list(connection, *, inc_drafts=False, inc_hidden=False, entity_i
 
     result = connection.execute(query, filter_data)
     count = result.fetchone()[0]
-    order_by_clause = str()
-    if sort_order == None:
+    order_by_clause = ""
+    if not sort_order:
         sort_order = 'DESC'
+    if sort_order.upper() not in ('ASC', 'DESC'):
+        raise ValueError("sort_order must be ASC or DESC")
 
     if sort == "popularity":
         order_by_clause = """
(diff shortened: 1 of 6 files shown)


critiquebrainz/frontend/views/review.py
```diff
@@ -42,14 +42,16 @@ def get_review_or_404(review_id):
 @review_bp.route('/')
 def browse():
     entity_type = request.args.get('entity_type', default=None)
+    sort = request.args.get('sort', default='popularity')
```
amCap1712, line 45: The earlier default (from Line 52) seems to be published_on. Let's keep that.

amCap1712 (changes requested):
The current changes look good although I have not yet tested locally. But these changes need to be made for the web service as well. Also, tests need to be written for the db method testing the sort, the frontend endpoint/template and the to be added api method.

amCap1712 (commented):
Thanks for the changes! 

The added tests look good and in line with the existing ones. However, looking at it now, the existing tests seem incomplete (testing sort options with just 1 review seems not a useful test). Anyway fine to leave as is for now. I'll make a separate ticket to improve the tests in future.

I'll test on beta and if everything goes well, will merge.

alastair:
I was just looking at this and thought that it might be nice to have a `sort_order` parameter which affects the order of all sort values, so that we don't have to special-case the published_on_asc sort field - I'll take a look at this this week and work with @anshg1214 on it

anshg1214:
Yes it's a good idea. We can parse it in the API as well. I'll work on it.

critiquebrainz/db/review.py
```diff
@@ -471,15 +471,17 @@ def get_reviews_list(connection, *, inc_drafts=False, inc_hidden=False, entity_i
     result = connection.execute(query, filter_data)
     count = result.fetchone()[0]
     order_by_clause = str()
+    if sort_order == None:
```
alastair: small thing here - typically in python if you want to compare something against none you can use `if not sort_order`, or if you really want to know that it's None and not an empty string/empty list you can use `if sort_order is None`

critiquebrainz/frontend/templates/review/browse.html
```diff
@@ -5,24 +5,36 @@
 
 {% block content %}
 <h2>{{ _('Reviews') }}</h2>
-<p><em class="text-muted">{{ _('Ordered by creation date') }}</em></p>
+<p><em class="text-muted">{{ _('Ordered by') }} {{ sort_options[sort, sort_order] }}</em></p>
```
alastair, line 8: We should have a check somewhere in the view to ensure that these two variables sort and sort_order are known and present in sort_options, otherwise if they're invalid this item will crash

alastair:
@anshg1214 thanks for this PR. I added a few small comments, but I've also fixed these items myself.

I also added a new sort order (popularity ascending)
I made sure that all of the text that we show is marked with `gettext()` so that it can be translated. I saw that we were duplicating some text in two places (the description of the sort order, and also in the dropdown), so I removed the duplication.

I found that there was an existing bug that didn't validate that the entity type or sort type were valid values when on the "review list" page, I added some checks for that and the new sort order field.

@amCap1712 it'd be great if you could have a quick look at my recent changes too

critiquebrainz/db/review.py
anshg1214: Understood

PR #396, opened 2022-02-21
Switch to http-hosted location for CB dumps

alastair:
As most browsers now don't have a built-in ftp client

### DIFF
--- a/critiquebrainz/frontend/templates/footer.html
+++ b/critiquebrainz/frontend/templates/footer.html
@@ -22,7 +22,7 @@
       <ul>
         <li><a href="https://critiquebrainz.readthedocs.io/api.html">{{ _('Web API') }}</a></li>
         <li><a href="https://github.com/metabrainz/critiquebrainz">{{ _('Source code') }}</a></li>
-        <li><a href="ftp://ftp.musicbrainz.org/pub/musicbrainz/critiquebrainz/">{{ _('Data dumps') }}</a></li>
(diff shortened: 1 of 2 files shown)


critiquebrainz/frontend/templates/footer.html
```diff
@@ -22,7 +22,7 @@
       <ul>
         <li><a href="https://critiquebrainz.readthedocs.io/api.html">{{ _('Web API') }}</a></li>
         <li><a href="https://github.com/metabrainz/critiquebrainz">{{ _('Source code') }}</a></li>
-        <li><a href="ftp://ftp.musicbrainz.org/pub/musicbrainz/critiquebrainz/">{{ _('Data dumps') }}</a></li>
+        <li><a href="http://ftp.musicbrainz.org/pub/musicbrainz/critiquebrainz/">{{ _('Data dumps') }}</a></li>
```
amCap1712, line 25: Nit: Maybe `http://data.metabrainz.org/pub/musicbrainz/critiquebrainz/` instead it is the main server? 

Although the ftp one also pointed to the mirror so its probably fine either way.

amCap1712:
Failing tests seem to be due to markupsafe getting (similar to itsdangerous) a newer release which is incompatible with Flask 1.*

critiquebrainz/frontend/templates/footer.html
alastair: yeah, I just updated the protocol and left the domain. I note that on https://listenbrainz.org/data/ we call this host the "main download site (USA)", so I guess for some definition it's also a main site. I'm happy to update it if you think it's important

alastair:
cool, let's look at upgrading or pinning markupsafe in a new PR

PR #398, opened 2022-02-28
CB-401, CB-408: Return 404 for entities that don't exist in the MB database

alastair:
This is a prerequisite to upgrading to BU 2.0 - we removed the concept of "unknown entities", and so had to make a decision about how to present these in CB.

Based on discussion in CB-401 and CB-408, we decided to explicitly return 404 on all entity pages that don't exist, and also on review pages for entities that don't exist.

In a future PR, we'll add an admin interface to see these items, and add the ability to change the entity id that the review is targeting, or delete the review.

In https://github.com/metabrainz/critiquebrainz/commit/754f5a71c62d22ddc960d94293349f6cf5a650ed I reverted a change that removed the test config file, because I realised that the test db name is different in tests and development.
Because create_app actually connects to the database in the config file, it's not possible to create the app and then change the config value and then connect to the database. Easier to revert the change.

I switched from using the metabrainz/musicbrainz-test-database:beta postgres image to metabrainz/brainzutils-mb-sample-database:schema-25-2021-04-04.0. This is much faster to start up (I was waiting ~10-15 seconds for the initial setup and schema creation every time I ran tests, now it's basically instant), and now the database has real musicbrainz data in it. This means that we were able to get rid of a bunch of mocks, which have always been a bit flaky for us. This involved changing a few entity mbids in tests to ones which were in the database.

Also fixed a bug in get_revision_number that showed up during testing, and pinned some dependency versions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM metabrainz/python:3.8-20210115
 
+ENV PYTHONUNBUFFERED 1
+
 # remove expired let's encrypt certificate and install new ones
 RUN apt-get update \
     && apt-get install -y --no-install-recommends ca-certificates \
--- a/critiquebrainz/data/testing.py
+++ b/critiquebrainz/data/testing.py
@@ -9,9 +9,7 @@ from critiquebrainz.frontend import create_app
 class DataTestCase(TestCase):
 
     def create_app(self):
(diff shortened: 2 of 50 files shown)


amCap1712 (approved):
LGTM

PR #400, opened 2022-03-03
[CB-420]: Fix comment input

MonkeyDo:
CB-420 surfaced that currently validation of comment input isn't working.
In short, we generate a hidden input textarea with the `required` attribute.
The textarea being both hidden and required causes some issues with native browser validation hints.

By using the Length validator instead of InputRequired, we conserve the intended mechanism without the textarea having a `required` prop.
Trying to post a comment with no content results in the expected error message:
image


alastair:
thank you!

PR #405, opened 2022-03-16
Finish upgrade to brainzutils 2.4

alastair:
This is quite a messy PR, but there turned out to be a bunch more things to change than expected.

In short:
* https://github.com/metabrainz/critiquebrainz/pull/404 made some good changes to improve caching, but this requires code which turned into BU 2.0, so this PR has to be based on that one
* We made 3 major changes in the BU 2.0 code that returns MB data from the db - no longer return dummy data if the entity doesn't exist, update method names to _by_mbid, and return a `mbid` field instead of `id` in data.
* The best way that I was able to find places in templates to update here was to set `app.jinja_options['undefined'] = jinja2.StrictUndefined`, which raises an exception when the template engine can't find a variable. 
* This resulted in me finding a whole bunch of places where we expected data to be returned but didn't actually have it (probably due to the original change from mb API to mb database access?) https://github.com/metabrainz/brainzutils-python/pull/81 fixed this, and some other changes
* When things legitimately don't exist in a template (such as `error`, only present if an error is set, or related fields on an MB object that only sometimes exist such as `life-span`), I used the jinja syntax `{% if artist['life-span'] is defined %}` instead of the shortcut `if artist['lifespan']`. However, I only set `StrictUndefined` during development, so as not to cause errors in CB if I've missed anything. I think I've visited all pages and updated/fixed variables, but I'm not 100% sure. Many of the if statements are due to cases where data might or might not exist, so to be completely sure, we would need tests for each variation of an entity. For now I'm just hoping that I've got everything

It became clear during testing that there were a bunch of buggy endpoints which we haven't noticed because no one seems to be writing reviews for some specific entities. I fixed as many of these as possible but there mgiht still be some problems.

I'm tempted to just merge this and deploy it, and optionally set `jinja2.StrictUndefined` and keep an eye on sentry alerts so that we can fix them... thoughts?


amCap1712 (approved):
LGTM. I suggest we put on beta with StrictUndefined and try it out for a while, ask users to do the same.

PR #407, opened 2022-03-23
GH actions: deploy Docker images from PRs

MonkeyDo:
The current Github Actions workflow we use to deploy development Docker images allows us to select a branch or tag from this repo, but not PRs that are created from a forked repo.

With these modifications we can target the ref of the pull request even if it was created from a forked repo.

See also  https://github.com/metabrainz/listenbrainz-server/pull/1928


alastair:
Thanks!

PR #416, opened 2022-04-06
Add missing check for moderator CB id

amCap1712:
AFAIU this situation should only happen if a moderator does not have a CB account. Since, we have StrictDefined mode on we need the is defined check first. Found in local development.


alastair (approved):
this is fine, but it probably doesn't need the final `and mod[]`... If the item is missing if there is no cb_id, but present and filled in if there is, then just the defined check should be enough.

PR #417, opened 2022-04-06
CB-390: Shouldn't have to choose a license just to save drafts

amCap1712:
Supersedes #327


alastair:
This doesn't make sense any more in the context of https://github.com/metabrainz/critiquebrainz/pull/474, closing

PR #421, opened 2022-04-11
Upgrade python version and dependencies

alastair:
Upgrade to python 3.10 and flask 2.0, along with some other python dependencies.

Still some pending items to fix:
* Some tests that use flask-testing's assertRedirect are broken, it seems due to a change in flask between v1 and v2 along with some assumptions in flask-testing which are now broken. flask-testing is quite old and unmaintained, as indicated by a few open issues. We can fix these tests manually, but might be something to look into in more detail to see if it's useful for us
* I've run tests, but haven't actually loaded the app to see if everything works as expected
* new version of pycountry caused translations to stop working
* Tests seem to take longer to run than they used to - this is just a feeling. Some tests (such as critiquebrainz/frontend/views/test/test_review.py::ReviewViewsTestCase) take a second or so to run each test.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - \
 
 RUN pip install --upgrade pip==21.0.1
 
-RUN pip install --no-cache-dir uWSGI==2.0.18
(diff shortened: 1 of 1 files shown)


amCap1712 (approved):
LGTM, Thanks!

PR #423, opened 2022-04-22
CB-434: Show events on place page 

anshg1214:
CB-434

Show the events linked with the place on the place's page.

Pending tasks: Upgrade to latest version of BU


critiquebrainz/frontend/views/test/test_place.py
```diff
@@ -1,11 +1,54 @@
+from unittest import mock
+
 from critiquebrainz.frontend.testing import FrontendTestCase
 
+def return_events(*, place_id, event_types=None, limit=None, offset=None):
```
alastair, line 5: I prefer to use real data from the database instead of a mock in this kind of test, because we've run into problems before where the returned data from the library changes and we forget to update the mocks.

Now that the CB tests use the MB sample database (like the BU tests), we are able to get real data - can you see if you can replace this test with one that uses real data for a place?

anshg1214: Done

PR #424, opened 2022-04-26
Only allow http/s protocols in an oauth application's callback URL

alastair:
The callback URL is used in an `<a href>` on the oauth permissions page in
the "deny" button, making it possible for someone to craft a URL that
could redirect a user to a malicious page. Prevent this by requring the
callback URL to refer to properly formatted URL.
The homepage url is not checked, but this uses target="_blank", which
causes browsers to block it when opening the link.


amCap1712 (approved):
I think we should validate homepage uri anyways (fine to do it in a later PR) but otherwise LGTM, Thanks!

alastair:
Thanks for the report!

PR #425, opened 2022-04-26
Show reports for entities other than release groups

alastair:
There were bugs on the reports page that caused errors when a review of a non-releasegroup entity was reporterd


amCap1712 (approved):

amCap1712:
LGTM, Thanks!

alastair:
Yeah, there are a lot of places where we found errors in templates due to things that weren't updated when new entities were added. I'm not sure if it's a great use of time to generate tests for every single combination of data (e.g. event with a main artist, event with a date, with a place, without some of those, ...) but it does seem like there are some changes that we could make to make it a bit easier to test that the correct data is loaded when we expect it. Perhaps something to look into later.

PR #426, opened 2022-04-26
Add macro to show life span

anshg1214:


alastair:
We had a bit of a discussion about this in IRC. Because we use life-span in many places, it might be nice to create a macro that allows us to show them in any necessary place. Such a macro could do the following:

* If an entity has no life-span, display nothing
* If it has a start but no end, just show the start
* If it has an end, show a - and the end
* allow a flag to determine if you show the full date (possibly yyyy-mm-dd), or just the year (truncate to 4 characters)

Then every place that we show a life-span, use this macro

PR #427, opened 2022-05-21
CB-410: Revisions of reviews are created when drafts are made, and they remain visible

anshg1214:
[CB-410](https://tickets.metabrainz.org/projects/CB/issues/CB-410)
The revisions are saved until the review is in draft, but the older draft revisions are deleted whenever a user publishes that review.


amCap1712:
At a quick look, these changes need to be made for the api as well?

critiquebrainz/frontend/views/test/test_review.py
```diff
@@ -374,3 +375,32 @@ def test_hide_redirect(self):
             review["entity_type"], review["entity_id"]))
         redirect_url = urlparse(response.location)
         self.assertEqual(redirect_url.path, url_for('review.entity', id=review['id']))
+
+
+    def test_publish_draft_review(self):
```
amCap1712, line 380: Since we are deleting data, this test should be more comprehensive IMO. Probably fetch the actual revisions before publish, assert, then fetch again after publish to confirm the correct revisions were deleted.

anshg1214: Sounds Good! I've improved the tests.

PR #428, opened 2022-05-25
Upgrade to MB schema 27 and new version of mbdata and sample DB

alastair:
One entity which was in schema 25 sample db is no longer present in
the new one, update the entity id


amCap1712 (approved):
LGTM, Thanks!

PR #429, opened 2022-05-26
Update search results view

alastair:
* Always show a disambiguation if it exists, and make it `<small>`
* Always make the first column of search results 50% wide. Search results are 4 or 5 columns wide, the other fields fit in the other 50%
* Add a link to the Place on the search results for an event
* Fix some checks for undefined data in templates

Before
Screen Shot 2022-05-26 at 5 14 29 PM

After
Screen Shot 2022-05-26 at 5 14 49 PM

@MonkeyDo tagging you for css check - I decided to just add a new class for it instead of repeating a `style` attribute, we don't have any react or embedded css, so just adding it to a section in the main less file.

### DIFF
--- a/critiquebrainz/frontend/static/styles/main.less
+++ b/critiquebrainz/frontend/static/styles/main.less
@@ -102,6 +102,11 @@ ul.sharing {
   margin-top: 4px;
 }
 
+// Search page
+#search-results thead tr th:first-child {
+  width: 50%;
+}
+
 // Review listings
 .cover-art-container {
   position: relative;
--- a/critiquebrainz/frontend/templates/event/entity.html
+++ b/critiquebrainz/frontend/templates/event/entity.html
@@ -30,7 +30,7 @@
 
   {% if event['place-rels'] is defined %}
     {% set place = event['place-rels'][0]['place'] %}
-    {% if place['coordinates'] %}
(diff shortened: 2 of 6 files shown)


amCap1712 (approved):
LGTM, Thanks!

critiquebrainz/frontend/static/styles/main.less
```diff
@@ -102,6 +102,11 @@ ul.sharing {
   margin-top: 4px;
 }
 
+// Search page
+.search-result-title {
+  width: 50%;
+}
```
MonkeyDo, line 108: At some point I wonder if my brain is just trying to justify all the css it contains, but you could also do this:

```
#search-results thead tr th:first-child {
  width: 50%;
}
```

and add a single `id=`search-results"` to your `<table>` element to avoid repeating the class in each `if` block

MonkeyDo (commented):
Looks good! A suggestion for extra fancy points:

critiquebrainz/frontend/static/styles/main.less
alastair: at the very least I like the idea of adding in an id to anchor the css a bit more to a specific page - I was thinking about doing that.

PR #430, opened 2022-05-26
Show both forward and backward place-place rels (Parts and Part of)

alastair:
On a Place page we would show the parent place that this place was a part of (backward place-place rel), but we wouldn't show the child places of this place.

Update parsing code and display to add this information


amCap1712 (approved):
LGTM, Thanks!

But probably best to test on beta first.

PR #431, opened 2022-05-26
CB-429: API: Default sort by published_on desc if no sort params provided

alastair:
By default the reviews list endpoint had no sort, which meant that we weren't sending any `ORDER BY` clause to the database. Update to always include a default sort if it's not set.


amCap1712 (approved):
LGTM, Thanks!

(Looking at the code, it seems if Parser class had an option for specifying default value would be great. But that seems too much work for the small amount of gains here so if/or is fine).

PR #433, opened 2022-06-04
Add BookBrainz Database 

anshg1214:
This PR adds BookBrainz Database to the CB database.

### DIFF
--- a/critiquebrainz/frontend/__init__.py
+++ b/critiquebrainz/frontend/__init__.py
@@ -75,6 +75,10 @@ def create_app(debug=None, config_path=None):
 
     add_robots(app)
 
+    # BookBrainz Database
+    import critiquebrainz.frontend.external.bookbrainz_db as bookbrainz_db 
+    bookbrainz_db.init_db_engine(app.config.get("BB_DATABASE_URI"))
+    
     # MusicBrainz Database
     from brainzutils import musicbrainz_db
     musicbrainz_db.init_db_engine(app.config.get("MB_DATABASE_URI"))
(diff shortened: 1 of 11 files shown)


docker/pg_custom/create-bb-db.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-EOSQL
+    CREATE USER critiquebrainz WITH SUPERUSER PASSWORD 'critiquebrainz';
+    CREATE DATABASE critiquebrainz;
+    GRANT ALL PRIVILEGES ON DATABASE critiquebrainz TO critiquebrainz;
```
alastair, line 7: can you split this into 2 files, one for CB and one for BB?

alastair (approved):
this looks great, thanks! One small change that we should make to the postgres startup scripts

critiquebrainz/frontend/__init__.py
```diff
@@ -75,6 +75,10 @@ def create_app(debug=None, config_path=None):
 
     add_robots(app)
 
+    # BookBrainz Database
+    import critiquebrainz.frontend.external.bookbrainz_db as bookbrainz_db 
+    bookbrainz_db.init_db_engine(app.config.get("BB_DATABASE_URI"))
```
alastair, line 80: For now it would be nice to be able to run CB without a bookbrainz database. To do this, we can check if `app.config.get("BB_DATABASE_URI")` is None, and if so don't run init_db_engine.

PR #434, opened 2022-06-06
Fetch edition group from BB database

anshg1214:
This PR fetches edition group from the BB database.


critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    if bbids == []:
+        return {}
+
+    key = cache.gen_key('edition-groups', bbids)
+    edition_groups = cache.get(key)
+    if not edition_groups:
+        with db.bb_engine.connect() as connection:
+            result = connection.execute(sqlalchemy.text("""
+                SELECT
```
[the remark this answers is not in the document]
alastair: @anshg1214 pointed out to me that the views don't have all necessary columns, so he used a direct query instead. We should check with @MonkeyDo to see if we can add these columns to the views

critiquebrainz/frontend/external/__init__.py
```diff
@@ -6,3 +6,80 @@
...
+       isn't in the MusicBrainz database. Used in development with a sample database."""
+    entity = get_entity_by_id(entity_id, entity_type)
+    if entity is None and current_app.config["DEBUG"]:
+        current_app.logger.info("returning dummy entity in development mode")
+        return get_dummy_item(entity_id, entity_type)
+    return entity
+
+
+def get_dummy_item(entity_id, entity_type):
+    """Get something that looks just enough like a MusicBrainz entity to display in a CB template"""
```
alastair, line 79: If we are in DEBUG mode, and we try and get a review for a BB item which isn't in the BB database, we will return this item. Are the fields correct to use in the BB pages?
I know that this is very unlikely to happen, as we're currently using a full database for BB tests and local development, however I'd like to ensure that we make this work now rather than encounter an error later.

Perhaps the change could be as simple as `if entity_type.startswith("bb_"): return get_dummy_bb_item....` in the development methods above

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -0,0 +1,77 @@
...
+from critiquebrainz.frontend.external.bookbrainz_db.relationships import fetch_relationships
+
+
+def get_edition_group_by_bbid(bbid: uuid.UUID) -> dict:
+    """
+    Get info related to an edition group using its BookBrainz ID.
+    Args:
+        bbid : BBID of the edition group.
+    Returns:
+        A dictionary containing the basic information related to the edition group.
```
alastair, line 17: I know that the musicbrainz methods don't include this, but can we list the fields that this method returns? That's just the list of items that we SELECT in the sql query below.

alastair: In documentation it's also good to include what is returned in the case of an error. In this case, "Returns None if the BBID doesn't exist in the database"

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -0,0 +1,77 @@
...
+    Get info related to multiple edition groups using their BookBrainz IDs. 
+    Args:
+        bbids (list): List of BBID of edition groups.
+    Returns:
+        A dictionary containing info of multiple edition groups keyed by their BBID.
+    """
+    if bbids == []:
+        return {}
+
+    key = cache.gen_key('edition-groups', bbids)
```
alastair, line 36: to help us in the future, let's prefix all bookbrainz cache uses with `bb_` or similar (especially the 'identifier' one in identifiers.py, that word is common enough that it could conflict with something else in the future)

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -0,0 +1,77 @@
...
+    Get info related to an edition group using its BookBrainz ID.
+    Args:
+        bbid : BBID of the edition group.
+    Returns:
+        A dictionary containing the basic information related to the edition group.
+    """
+    edition_group = fetch_multiple_edition_groups([bbid])
+    if not edition_group:
+        return None
+    return edition_group[bbid]
```
alastair, line 22: there's a very sneaky bug here. 
If you pass in a bbid formatted slightly differently (e.g. "343ca0d1-cf7e-4a69-8701-97733d81c3B0" - note that the last b is upper-case), then postgres will happily accept it, but return `edition_group.bbid` in lower-case, and this dictionary key access will fail.

In AcousticBrainz, we check user-provided data at the API level by turning it into a UUID object and then back into a string: https://github.com/metabrainz/acousticbrainz-server/blob/ed51bcc5afaf07f78fd0cdf48fc28db844c3ca6b/webserver/views/api/v1/core.py#L196-L200
however it's much more common to do this at the user-level, and assume that the data is clean by the time it gets down to a low-level data method like this one. I see that the musicbrainz methods fail in the same way.
I'll take a deeper look into this and let you know if we should change it.

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -0,0 +1,77 @@
...
+from brainzutils import cache
+from typing import List
+import sqlalchemy
+import critiquebrainz.frontend.external.bookbrainz_db as db 
+from critiquebrainz.frontend.external.bookbrainz_db import DEFAULT_CACHE_EXPIRATION
+from critiquebrainz.frontend.external.bookbrainz_db.identifiers import fetch_identifiers
+from critiquebrainz.frontend.external.bookbrainz_db.relationships import fetch_relationships
+
+
+def get_edition_group_by_bbid(bbid: uuid.UUID) -> dict:
```
alastair, line 11: I see that you have a type hint here saying that this method takes a UUID object, but if I try it I get an error:
```py
In [9]: edition_group.get_edition_group_by_bbid(uuid.UUID("343ca0d1-cf7e-4a69-8701-97733d81c3b0"))
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Input In [9], in <cell line: 1>()
----> 1 edition_group.get_edition_group_by_bbid(uuid.UUID("343ca0d1-cf7e-4a69-8701-97733d81c3b0"))

File /code/critiquebrainz/frontend/external/bookbrainz_db/edition_group.py:22, in get_edition_group_by_bbid(bbid)
     20 if not edition_group:
     21     return None
---> 22 return edition_group[bbid]

KeyError: UUID('343ca0d1-cf7e-4a69-8701-97733d81c3b0')
```

This is similar to the previous issue of upper-case letters in a UUID string.

In this case we should probably convert it to `str(bbid)` in the case that it's a UUID object.

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    if bbids == []:
+        return {}
+
+    key = cache.gen_key('edition-groups', bbids)
+    edition_groups = cache.get(key)
+    if not edition_groups:
+        with db.bb_engine.connect() as connection:
+            result = connection.execute(sqlalchemy.text("""
+                SELECT
```
alastair, line 41: Take a look a the (in-progress) MusicBrainz SQL formatting guidelines, and see if you can update this query to follow that format: https://gist.github.com/reosarevok/2d8a2fa6310cc0144cad043a9c566717 

eventually we would like to follow this format with all of our queries in all apps, this is a good place to start! Ask me or @reosarevok if you have any questions with the guidelines.

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -0,0 +1,77 @@
...
+                    edition_group.name,
+                    sort_name,
+                    edition_group_type,
+                    disambiguation,
+                    identifier_set_id,
+                    relationship_set_id,
+                    JSON_AGG( acn order by "position" ASC) as artist_credits
+                FROM edition_group 
+                LEFT JOIN author_credit_name acn on acn.author_credit_id = edition_group.author_credit_id 
+                WHERE bbid in :bbids and master = 't'
```
alastair, line 52: I tried this method with an invalid-formatted bbid and the query failed, raising an sqlalchemy exception. This happens with the musicbrainz methods too. I'll look into this and see if we should update these methods to catch this exception and return a CB-specific data access error.

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -0,0 +1,77 @@
...
+                edition_group['bbid'] = str(edition_group['bbid'])
+                edition_group['identifiers'] = fetch_identifiers(edition_group['identifier_set_id'])
+                edition_group['rels'] = fetch_relationships( edition_group['relationship_set_id'], ['Edition'])
+                results[edition_group['bbid']] = edition_group
+            
+            edition_groups = results
+            cache.set(key, edition_groups, DEFAULT_CACHE_EXPIRATION)
+    
+    if not edition_groups:
+        return None
```
alastair, line 76: In methods that return a dictionary or list, I prefer to return an empty version of that object instead of None in this case, as it can still be tested as a false object with `if not`, but it "looks like" the correct object type. For example, it allows you to do something like

```py
data = get_all_the_things()
newdata = [process(d) for d in data]
```

if `get_all_the_things` returns a list normally, but `None` in some cases, you can't do this without first checking if it's None or not.

In this case, better to return an empty `{}` in this method

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -0,0 +1,77 @@
...
+            edition_groups = result.fetchall()
+            results = {}
+            for edition_group in edition_groups:
+                edition_group = dict(edition_group)
+                edition_group['bbid'] = str(edition_group['bbid'])
+                edition_group['identifiers'] = fetch_identifiers(edition_group['identifier_set_id'])
+                edition_group['rels'] = fetch_relationships( edition_group['relationship_set_id'], ['Edition'])
+                results[edition_group['bbid']] = edition_group
+            
+            edition_groups = results
```
alastair, line 72: very small thing here, I would skip this line, and `return results` at the end of the method. There's no reason to overwrite this variable, and it makes it a bit confusing as to what data it includes (the list from the database query, or this differently formatted return data)

critiquebrainz/frontend/external/bookbrainz_db/identifiers.py
```diff
@@ -0,0 +1,123 @@
...
+    Fetch identifiers from the database.
+    Args:
+        identifier_set_id (int): Identifier set ID.
+    Returns:
+        List of identifiers.
+    """
+    if not identifier_set_id:
+        return None
+    
+    key = cache.gen_key('identifier', identifier_set_id)
```
alastair, line 18: use bb_ prefix

critiquebrainz/frontend/external/bookbrainz_db/identifiers.py
```diff
@@ -0,0 +1,123 @@
+from typing import List
+from brainzutils import cache
+import sqlalchemy
+import critiquebrainz.frontend.external.bookbrainz_db as db
+from critiquebrainz.frontend.external.bookbrainz_db import DEFAULT_CACHE_EXPIRATION
+
+def fetch_identifiers(identifier_set_id: int) -> List:
```
alastair, line 7: can we have a name that's more descriptive than `identifiers`? maybe they're bookbrainz_external_site_identifiers?

critiquebrainz/frontend/external/bookbrainz_db/identifiers.py
```diff
@@ -0,0 +1,123 @@
...
+            result = connection.execute(sqlalchemy.text("""
+                SELECT
+                    iden.type_id as type_id,
+                    idtype.label as label,
+                    iden.value as value
+                FROM bookbrainz.identifier_set__identifier idens
+                LEFT JOIN bookbrainz.identifier iden on idens.identifier_id = iden.id
+                LEFT JOIN bookbrainz.identifier_type idtype on iden.type_id = idtype.id
+                WHERE idens.set_id = :identifier_set_id;
+                """), ({'identifier_set_id': identifier_set_id}))
```
alastair, line 31: double-check this syntax. I think the `()` is unnecessary, and probably has no effect

critiquebrainz/frontend/external/bookbrainz_db/identifiers.py
```diff
@@ -0,0 +1,123 @@
...
+        return None
+    return identifiers
+
+
+def process_identifiers(identifiers: List) -> List:
+    """Process identifiers and include urls."""
+    external_urls = []
+
+    url_map = {
+        1: "https://musicbrainz.org/release/",
```
alastair, line 47: This map is OK for now, as it copies [data that is already in the BB website](https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/components/pages/entities/identifiers-links.js). However, I don't like having data that can potentially change in two different locations (BB and CB). In the case that a new identifier is added to BB, we also need to remember to add it here, and I guarantee that we will forget to do so.

In BB, we already have the `bookbrainz.identifier_type` table, which includes these identifiers _and already has a column for a recognition regex_. I'd like to open a discussion with @MonkeyDo to add a new column to this table to also have the "construct url" text, so that we can share this data between BB and CB by using the database.

Can you please open a ticket in Jira for the CB project for an enhancement to read these identifiers from the database?

critiquebrainz/frontend/external/bookbrainz_db/identifiers.py
```diff
@@ -0,0 +1,123 @@
...
+        34: "imdb-16.png",
+    }
+    
+    for identifier in identifiers:
+        value = identifier['value']
+        type_id = identifier['type_id']
+        if type_id == 13: 
+            url = url_map[13] + value.replace(" ", "") # Remove spaces first (see BB-499)
+        else:
+            url = url_map[type_id] + value
```
alastair, line 110: are we sure that all URLs are going to have this structure (`http://website.com/[value]`)? I see that the [bookbrainz version of this method](https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/components/pages/entities/identifiers-links.js) explicitly substitutes the `${value}` variable.
In python we can also use string formatting:

```py
url_map = {
    1: "https://musicbrainz.org/release/{value}",
...
url = url_map[type_id].format(value=value)
```
This will help future-proof this code in case we end up with a URL that doesn't have the value at the end of the URL

alastair: In the case that new `type_id` is added to BB and not added here, this will fail. Can we add a check here to avoid this error?

critiquebrainz/frontend/external/bookbrainz_db/identifiers.py
```diff
@@ -0,0 +1,123 @@
...
+        
+        if type_id in icon_map:
+            icon = icon_map[type_id]
+        else:
+            icon = None
```
alastair, line 115: for your information, this could also be written as 
```py
icon = icon_map.get(type_id, None)
```
I tend to use this syntax to save space

critiquebrainz/frontend/external/bookbrainz_db/identifiers.py
alastair, line 110: Possible alternate way of writing this:
```suggestion
        if type_id == 13: 
            value = value.replace(" ", "") # Remove spaces first (see BB-499)
        if type_id in url_map:
            url = url_map[type_id] + value
        else:
            url = None
```

critiquebrainz/frontend/external/bookbrainz_db/relationships.py
```diff
@@ -0,0 +1,74 @@
...
+            SELECT
+                label
+            FROM bookbrainz.relationship_type
+        """))
+        relationships = [relationship[0] for relationship in result.fetchall()]
+        relationships_mapping = {relationship.lower(): relationship for relationship in relationships}
+
+        for relation_type in relation_types:
+            if relation_type not in relationships_mapping:
+                raise bb_exceptions.InvalidTypeError("Bad release_types: {rtype} is not supported".format(rtype=relation_type))
```
alastair, line 32: release_type?

critiquebrainz/frontend/external/bookbrainz_db/relationships.py
```diff
@@ -0,0 +1,74 @@
...
+                FROM bookbrainz.relationship_set__relationship rels
+                LEFT JOIN bookbrainz.relationship rel on rels.relationship_id = rel.id
+                LEFT JOIN bookbrainz.relationship_type reltype on rel.type_id = reltype.id
+                WHERE rels.set_id = :relationship_set_id
+                AND reltype.label in :relation_types
+            """), {'relationship_set_id': relationship_set_id, 'relation_types': tuple(relation_types)})
+            relationships = result.fetchall()
+            relationships = [dict(relationship) for relationship in relationships]
+            for relationship in relationships:
+                relationship['source_bbid'] = str(relationship['source_bbid'])
```
alastair, line 68: if you want to make a UUID field from postgres into a string, you can also use
```
SELECT rel.source_bbid::text as source_bbid
```
to get postgres to turn it into a string for you!

critiquebrainz/frontend/external/bookbrainz_db/identifiers.py
alastair: @MonkeyDo do you have a better idea?

critiquebrainz/frontend/external/bookbrainz_db/test/edition_group_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_fetch_multiple_edition_groups(self):
+        edition_groups = edition_group.fetch_multiple_edition_groups([self.bbid2, self.bbid3])
+        self.assertEqual(len(edition_groups), 2)
+        self.assertEqual(edition_groups[self.bbid2]["bbid"], self.bbid2)
+        self.assertEqual(edition_groups[self.bbid2]["name"], "Stephen King Goes to the Movies")
+        self.assertEqual(edition_groups[self.bbid2]["edition_group_type"], "Book")
+        self.assertEqual(edition_groups[self.bbid3]["bbid"], self.bbid3)
+        self.assertEqual(edition_groups[self.bbid3]["name"], "Harry Potter and the Deathly Hallows")
+        self.assertEqual(edition_groups[self.bbid3]["edition_group_type"], "Book")
```
alastair, line 31: can we add a test for some relationships too?

this might get kind of complex, because we're using a real database that changes every week and tests might just suddenly start failing. perhaps we can ask @MonkeyDo to make us a "fixed" sample database every time BB makes a schema change so that we know that data won't change underneath us too often.

alastair (changes requested):
This is really great work, thank you!

I added a number of small comments to improve some code layout, and especially around error handling.

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
alastair: For now let's leave this as-is

critiquebrainz/frontend/external/bookbrainz_db/identifiers.py
```diff
@@ -0,0 +1,89 @@
...
+    
+    for identifier in identifiers:
+        value = identifier['value']
+        type_id = identifier['type_id']
+        url_template = identifier['url_template']
+
+        if type_id == 13: 
+            value = value.replace(" ", "") # Remove spaces first (see BB-499)
+
+        url = url_template.format(value=value)
```
alastair, line 79: this is great!

MonkeyDo: Noice

alastair (approved):
🎉

PR #435, opened 2022-06-07
Provide a user-friendly error if oauth params are incorrect 

alastair:
An invalid client id, mismatching redirect uri, or bad scope resulted in
an unhandled HTTP500 error, as the exception raised is only handled in
the API app, not the website app.
Catch and re-raise this exception so that it is properly handled.

Update db.oauth_client.create to actually return a dictionary as
documented, this is needed for the test.


amCap1712 (approved):
LGTM, Thanks@

PR #437, opened 2022-06-11
Nav Search Enhancement

akshaaatt:
- Require the user to enter something in the text box before presenting any results.


alastair (commented):
looks good, thanks! I also applied this to the search page, in case the user makes a search and then tries to delete the search term

PR #444, opened 2022-07-11
Allow searching of Edition Group

anshg1214:
This PR adds the feature to allow user to search for Edition Group

### DIFF
--- a/critiquebrainz/frontend/external/musicbrainz.py
+++ b/critiquebrainz/frontend/external/musicbrainz.py
@@ -7,8 +7,6 @@ Package musicbrainzngs is available at https://pypi.python.org/pypi/musicbrainzn
 More information about the MusicBrainz webservice can be found at http://wiki.musicbrainz.org/XML_Web_Service.
 """
 import musicbrainzngs
-import requests
-from critiquebrainz.frontend.external.bookbrainz_db.edition_group import fetch_multiple_edition_groups
 
 DEFAULT_CACHE_EXPIRATION = 12 * 60 * 60  # seconds (12 hours)
 THREAD_POOL_PROCESSES = 10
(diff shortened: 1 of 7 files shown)


alastair (changes requested):

critiquebrainz/frontend/external/musicbrainz.py
```diff
@@ -20,6 +22,17 @@ def init(app_name, app_version, hostname=None):
         musicbrainzngs.set_hostname(hostname)
 
 
+def search_editon_group(query='', limit=None, offset=None):
+    baseURL = 'https://bookbrainz.org/search/search'
+    params = {'q': query, 'type': 'EditionGroup', 'size': limit, 'from': offset}
+    data = requests.get(baseURL, params=params).json()
```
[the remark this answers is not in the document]
anshg1214: Done

critiquebrainz/frontend/external/musicbrainz.py
```diff
@@ -20,6 +22,19 @@ def init(app_name, app_version, hostname=None):
         musicbrainzngs.set_hostname(hostname)
 
 
+def search_editon_group(query='', limit=None, offset=None):
```
alastair, line 25: let's add a new `bookbrainz.py` for this, so that we can add the upcoming search methods to it.

critiquebrainz/frontend/views/search.py
```diff
@@ -24,6 +25,15 @@ def search_wrapper(query, type, offset=None):
             count, results = musicbrainz.search_labels(query, limit=RESULTS_LIMIT, offset=offset)
         elif type == "recording":
             count, results = musicbrainz.search_recordings(query, limit=RESULTS_LIMIT, offset=offset)
+        elif type == "bb_edition_group":
+            try:
+                count, results = musicbrainz.search_editon_group(query, limit=RESULTS_LIMIT, offset=offset)
+            except HTTPError:
+                raise BadRequest('Request failed while searching for edition groups.')
```
alastair, line 32: This is good, but we should use an error code that indicates a server error (starting with 5) instead of BadRequest, which indicates a client error. Let's use `ServiceUnavailable`.

I think that we can return a single error message here in case of all of these errors. so let's just do

`exept (HTTPError, JSONDecodeError, Timeout):`

and raise the same error message

alastair (changes requested):

alastair (approved):

PR #451, opened 2022-07-27
Update queries for author credits

anshg1214:


critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -60,7 +60,10 @@ def fetch_multiple_edition_groups(bbids: List[uuid.UUID]) -> dict:
                     disambiguation,
                     identifier_set_id,
                     relationship_set_id,
-                    json_agg( acn ORDER BY "position" ASC ) as author_credits
+                    COALESCE( json_agg( acn ORDER BY "position" ASC )
```
alastair, line 63: it's great that you were able to find coalesce!
I remember that we added some manual data for tests for relationships in BB entities. Would it also be possible to add some data with a null artist credit to ensure that this item isn't returned in the query?

anshg1214: Yes, it is possible. I'll add tests for both cases if author credits exist and if they don't.

admin/sql/test/bb-test-data.sql
```diff
@@ -9,3 +11,35 @@ INSERT INTO relationship_set (id) VALUES (99999999);
...
+            1,
+            0);
+
+
+INSERT INTO author_credit_name (author_credit_id, position, author_bbid, name, join_phrase )
+    VALUES (99999999,
+            0,
+            'e5c4e68b-bfce-4c77-9ca2-0f0a2d4d09f0',
+            'Test Author',
+            'Test Join Phrase');
```
alastair, line 37: for your information, normally a join phrase only exists if there are more than 1 name in the credit. It's so that you can do "Test Author **and** Someone Else" where __and__ is the join phrase. However, this is fine

alastair (approved):
looks good, thanks for the additional tests

PR #452, opened 2022-07-29
Allow CB to review Literary Works

anshg1214:
This PR allows users to browse and review BookBrainz Literary Works.

NOTE: This PR needs to be merged first.


critiquebrainz/frontend/external/bookbrainz_db/literary_work.py
```diff
@@ -0,0 +1,111 @@
...
+
+    bbids = [str(uuid.UUID(bbid)) for bbid in bbids]
+
+    if limit is None:
+        limit = len(bbids)
+
+    if work_type is None:
+        work_type_filter_string = ''
+    elif work_type in WORK_TYPE_FILTER_OPTIONS:
+        work_type_filter_string = 'AND work_type = \'%s\'' % work_type
```
alastair, line 58: even though we do some validation of this user input data, we should always use sql query parameters for anything in a WHERE statement.
We can do this by setting up the query parameters dict ahead of time:
```py
query_params = {}
elif work_type in WORK_TYPE_FILTER_OPTIONS:
     work_type_filter_string = 'AND work_type = :work_type'
     query_params['work_type'] = work_type
...
connection.execute(..., query_params)
```
You can then add additional items to query_params as needed (bbids, limit, offset, etc)

critiquebrainz/frontend/external/bookbrainz_db/literary_work.py
```diff
@@ -0,0 +1,111 @@
...
+    Returns:
+        A dictionary containing info of multiple literary works keyed by their BBID.
+    """
+    if bbids == []:
+        return {}
+
+    bbids = [str(uuid.UUID(bbid)) for bbid in bbids]
+
+    if limit is None:
+        limit = len(bbids)
```
alastair, line 53: You can pass `LIMIT NULL` to an sql query to set no limit. So passing `None` is fine here.

critiquebrainz/frontend/external/bookbrainz_db/literary_work.py
```diff
@@ -0,0 +1,111 @@
...
+        return None
+    return literary_work[bbid]
+
+
+def fetch_multiple_literary_works(bbids: List[uuid.UUID], work_type=None, limit=None, offset=0) -> dict:
+    """
+    Get info related to multiple literary works using their BookBrainz IDs. 
+    Args:
+        bbids (list): List of BBID of literary works.
+        work_type (str): Filter the results by work type.
```
alastair, line 41: add some documentation about what values are valid, and what a value of 'other' means

critiquebrainz/db/review.py
```diff
@@ -29,7 +29,23 @@
...
+    "release_group",
+    "work",
+    "artist",
+    "label",
+    "recording"
+]
+
+BOOKBRAINZ_ENTITY_TYPES = [
+    "bb_edition_group",
+    "bb_literary_work"
```
alastair, line 48: can we set `ENTITY_TYPES = MUSICBRAINZ_ENTITY_TYPES + BOOKBRAINZ_ENTITY_TYPES` here? (I thought I saw some other code where you did this, but I can no longer find it)

critiquebrainz/db/review.py
```diff
@@ -417,8 +433,15 @@ def get_reviews_list(connection, *, inc_drafts=False, inc_hidden=False, entity_i
         filter_data["entity_id"] = entity_id
 
     if entity_type is not None:
-        filters.append("entity_type = :entity_type")
-        filter_data["entity_type"] = entity_type
+        if entity_type == 'musicbrainz':
```
alastair, line 436: This is a great additional feature, but it should be documented everwhere it can be used - both in the docstring for this db method, and also in the API documentation if this can be used directly from the API

critiquebrainz/frontend/views/bb_literary_work.py
```diff
@@ -0,0 +1,62 @@
...
+bb_literary_work_bp = Blueprint('bb_literary_work', __name__)
+
+
+@bb_literary_work_bp.route('/<uuid:id>')
+def entity(id):
+    id = str(id)
+    literary_work = bb_literary_work.get_literary_work_by_bbid(id)
+
+    if literary_work is None:
+        raise NotFound(gettext("Sorry, we couldn't find a literary work with that BookBrainz ID."))
```
alastair, line 19: while we use 'literary work' as our identifier for this type, perhaps here we should just say "Sorry, we couldn't find a BookBrainz work with this ID"

critiquebrainz/frontend/external/bookbrainz.py
```diff
@@ -1,14 +1,27 @@
...
+
+
+def search_literary_work(query='', limit=None, offset=None):
+    params = {'q': query, 'type': 'Work', 'size': limit, 'from': offset}
+    data = requests.get(BASE_URL, params=params, timeout=5)
+    data.raise_for_status()
+    data = data.json()
+    count = data['total']
+    results = data['results']
+    bbids = [result["bbid"] for result in results]
```
alastair, line 25: All of this is the same as the editiongroup version, except for the `type` parameter. How about we break this out into a helper function `do_bookbrainz_search(entity_type, query, limit, offset)`?

alastair (changes requested):
This looks good, and works for me locally!
Just a few small recommendations for code cleanups. Make sure you also apply these to the other open PRs too

critiquebrainz/db/review.py
```diff
@@ -29,7 +29,23 @@
...
+    "event",
+    "place",
+    "release_group",
+    "work",
+    "artist",
+    "label",
+    "recording"
+]
+
+BOOKBRAINZ_ENTITY_TYPES = [
```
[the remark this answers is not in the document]
anshg1214: Yes, indeed!

critiquebrainz/frontend/external/bookbrainz_db/literary_work.py
anshg1214: Thats a great idea. I'll make the changes

critiquebrainz/frontend/external/bookbrainz_db/literary_work.py
anshg1214: Sure

critiquebrainz/db/review.py
anshg1214: I agree. I'll add it in the documentation

critiquebrainz/frontend/external/bookbrainz.py
anshg1214: I've now converted it to a single function. For this I have mapped the BB entity type name to CB entity type name and created a common function to fetch the entity data which does the job.

alastair (approved):
🎉

PR #455, opened 2022-08-04
Allow CB to review Series

anshg1214:
This PR allows users to browse and review BookBrainz Series.


alastair:
@anshg1214 I see some test failures here - perhaps they're passing in your environment because you have some old `.pyc` files around? Can you try and delete them and see if you can reproduce the failures?

PR #458, opened 2022-08-12
CB-444: API error messages are music-specific

anshg1214:
CB-444

### DIFF
--- a/critiquebrainz/ws/review/views.py
+++ b/critiquebrainz/ws/review/views.py
@@ -430,7 +430,6 @@ def review_list_handler():
 
         reviews = [db_review.to_dict(p) for p in reviews]
 
-
         if include_metadata == 'true':
             entities = [(str(review["entity_id"]), review["entity_type"]) for review in reviews]
             entities_info = mbstore.get_multiple_entities(entities)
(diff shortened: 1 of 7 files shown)


critiquebrainz/ws/review/views.py
```diff
@@ -508,7 +519,7 @@ def fetch_params():
         if language and language not in supported_languages:
             raise InvalidRequest(desc='Unsupported language')
         if db_review.list_reviews(user_id=user.id, entity_id=entity_id)[1]:
-            raise InvalidRequest(desc='You have already published a review for this album')
+            raise InvalidRequest(desc='You have already published a review for this {entity_name}'.format(entity_name=entity_name_map[entity_type]))
```
alastair, line 522: One possible issue that I see with this change is that if we add a new entity type and forget to add to the map, we'll raise an exception here.
I wonder if it would be just as good to have as the error message "You have already published a review for this item", what do you think?

I'm not sure how this error message works, but it looks like it might not be translated. Can you check to see if it is correctly changed if the language is changed?

alastair: @anshg1214 pointed out to me that this is a webservice lookup, and so isn't translated.

However, we see that we have a similar mapping already at https://github.com/metabrainz/critiquebrainz/blob/1944ad057e17f1283b6a4b5fd029fa7dfff5a642/critiquebrainz/frontend/views/user.py#L53

Perhaps it makes more sense to combine both of these mappings into a single place. It might make sense to put it at https://github.com/metabrainz/critiquebrainz/blob/1944ad057e17f1283b6a4b5fd029fa7dfff5a642/critiquebrainz/db/review.py#L24 so that there is just one file to edit when a new entity is added.

anshg1214: Absolutely! I've made the changes.

alastair (approved):

alastair:
@anshg1214 I believe I fixed the babel error that occurs when trying to render a `lazy_gettext` in the API, can you check?

anshg1214:
It is working now. Thanks!

PR #459, opened 2022-08-17
Display Relationships on Entity Pages

anshg1214:
This PR 
1. Displays the Works linked to the Edition Group
2. Displays translated works on the Work's page


critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -92,3 +92,47 @@ def fetch_multiple_edition_groups(bbids: List[uuid.UUID]) -> dict:
     if not results:
         return {}
     return results
+
+
+def fetch_work_for_edition_group(bbid: uuid.UUID):
```
alastair, line 97: we should have a test for this method. I'm not sure if it makes sense to create our own custom relationship like we do in some other tests so that we know that they exist and won't change, or if we should just use an existing relationship. I think that if we choose a book that was published some time ago (e.g. 30-40 years) then the chances of a new edition coming out and breaking the tests is quite low, so probably an acceptable tradeoff.

critiquebrainz/frontend/external/bookbrainz_db/literary_work.py
```diff
@@ -107,7 +107,7 @@ def fetch_multiple_literary_works(bbids: List[uuid.UUID], work_type=None, limit=
             for literary_work in literary_works:
                 literary_work = dict(literary_work)
                 literary_work['identifiers'] = fetch_bb_external_identifiers(literary_work['identifier_set_id'])
-                literary_work['rels'] = fetch_relationships(literary_work['relationship_set_id'], ['Edition'])
+                literary_work['rels'] = fetch_relationships(literary_work['relationship_set_id'], ['Translation'])
```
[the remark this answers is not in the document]
alastair: does this fill in _only_ translation rels into the variable `literary_work['rels']`? In that case, do you think it makes sense to call it `literary_work['translation_rels']` or something similar instead?

critiquebrainz/frontend/templates/bb_edition_group/entity.html
```diff
@@ -60,6 +60,51 @@ <h4 style="margin-bottom:0;">{{ _('Reviews') }}</h4>
         {% endif %}
       </ul>
     {% endif %}
+
+    {% if works is defined and works %}
+      <h4>{{ _('Literary Works') }}</h4>
```
alastair, line 65: maybe "Works in this edition group"

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
alastair: does this method return more than one work? if so, `fetch_works_for_edition_group`

critiquebrainz/frontend/templates/bb_edition_group/entity.html
```diff
@@ -60,6 +60,51 @@ <h4 style="margin-bottom:0;">{{ _('Reviews') }}</h4>
...
+      <table class="table table-condensed table-hover">
+        <thead>
+          <tr>
+            <th>{{ _('Name') }}</th>
+            <th>{{ _('Languages') }}</th>
+            <th>{{ _('Type') }}</th>
+          </tr>
+        </thead>
+        <tbody>
+          {% for work in works %}
```
alastair, line 75: do works in an edition have an ordering? I don't see an `ORDER BY` clause in the python method.

critiquebrainz/frontend/templates/bb_literary_work/entity.html
```diff
@@ -60,6 +60,52 @@ <h4 style="margin-bottom:0;">{{ _('Reviews') }}</h4>
         {% endif %}
       </ul>
     {% endif %}
+
+    {% if work_rels_info is defined and work_rels_info %}
+      <h4>{{ _('Translations') }}</h4>
```
alastair, line 65: "Translations of this work"

critiquebrainz/frontend/templates/bb_literary_work/entity.html
```diff
@@ -60,6 +60,52 @@ <h4 style="margin-bottom:0;">{{ _('Reviews') }}</h4>
...
+      <table class="table table-condensed table-hover">
+        <thead>
+          <tr>
+            <th>{{ _('Name') }}</th>
+            <th>{{ _('Languages') }}</th>
+            <th>{{ _('Type') }}</th>
+          </tr>
+        </thead>
+        <tbody>
+          {% for work in work_rels_info %}
```
alastair, line 75: it would be nice to give this variable a clearer name to say that it's translation relationships

critiquebrainz/frontend/templates/bb_literary_work/entity.html
```diff
@@ -60,6 +60,52 @@ <h4 style="margin-bottom:0;">{{ _('Reviews') }}</h4>
         {% endif %}
       </ul>
     {% endif %}
+
+    {% if work_rels_info is defined and work_rels_info %}
```
alastair, line 64: because we always pass this variable into the template context in the view, we don't need the `is defined` check. Let's just use `{% if work_rels_info %}` (so it will omit this block if the variable is `[]` but show it otherwise)

critiquebrainz/frontend/external/bookbrainz_db/literary_work.py
```diff
@@ -107,7 +107,7 @@ def fetch_multiple_literary_works(bbids: List[uuid.UUID], work_type=None, limit=
             for literary_work in literary_works:
                 literary_work = dict(literary_work)
                 literary_work['identifiers'] = fetch_bb_external_identifiers(literary_work['identifier_set_id'])
-                literary_work['rels'] = fetch_relationships(literary_work['relationship_set_id'], ['Edition'])
```
alastair, line 110: Do we also want to add an additional section to this page, "Edition Groups in which this work appears"? @MonkeyDo

alastair (changes requested):

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
```diff
@@ -92,3 +92,47 @@ def fetch_multiple_edition_groups(bbids: List[uuid.UUID]) -> dict:
...
+               FROM edition_group
+         INNER JOIN edition ON edition.edition_group_bbid = edition_group.bbid
+         INNER JOIN relationship_set__relationship rels on rels.set_id = edition.relationship_set_id
+          LEFT JOIN relationship rel on rels.relationship_id = rel.id
+              WHERE edition_group.bbid = :bbid
+                AND edition_group.master = 't'
+                AND edition_group.data_id IS NOT NULL
+                AND edition.master = 't'
+                AND edition.data_id IS NOT NULL
+                AND rel.type_id = 10
```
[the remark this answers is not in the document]
anshg1214: This is a good idea. I'll make the changes

critiquebrainz/frontend/external/bookbrainz_db/edition_group.py
anshg1214: I've added the tests.

critiquebrainz/frontend/external/bookbrainz_db/literary_work.py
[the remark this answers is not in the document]
anshg1214: Done. Now `type_id` is being used for fetching relationships.

critiquebrainz/frontend/templates/bb_edition_group/entity.html
anshg1214: No, Works for Edition does not have an ordering.

critiquebrainz/frontend/external/bookbrainz_db/literary_work.py
anshg1214: I've added this section

alastair (approved):

PR #461, opened 2022-08-24
CB-438: add rel=nofollow to links in reviews

alastair:
Because we allow users to write links in reviews and comments, we should mark them as rel=nofollow so that any spam links that might come through don't negatively affect CB's search ranking

I didn't include [noreferrer](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noreferrer) (or noopener) in this case, I'm not sure if it's strictly necessary as well, any thoughts @amCap1712?


amCap1712 (approved):
Let's noopener to links as well but otherwise LGTM, Thanks!

PR #462, opened 2022-08-24
CB-439 (I): Only include text reviews on front page recent reviews

alastair:
CB-439

Reviews can be either text, a rating, or both.
In my experience, looking at the recent reviews list on the homepage is annoying if you click through to some new ones and see that they're only ratings.
In order to make the list of reviews more useful, only show ones that have actual text reviews.


amCap1712 (approved):
LGTM, Thanks!

PR #466, opened 2022-09-24
Update musicbrainz_id on login

amCap1712:
We already do this in LB, implementing the same for CB. However, CB also has a display name field as well. If the display name has been changed to something else than the old musicbrainz id, update it as well otherwise let it remain set to the custom name.


alastair (approved):
thanks!

PR #468, opened 2022-09-26
 Update WTForms

amCap1712:
This is needed for https://github.com/metabrainz/critiquebrainz/pull/417.

Unselected radio field results in the value of the field being 'None' not None. This is fixed in newer versions. In newer versions, a few imports have changed and email_validators needs to be added a dependency separately to use EmailField.

Newer version of WTForms do not cache form.errors so each call to errors returns a new dict. Form level validation should instead be performed by appending errors to form.form_errors as shown in this commit. https://github.com/wtforms/wtforms/commit/22636b55eda9300b549c8bbaae6f9ae31595d445


alastair (approved):
👍

PR #474, opened 2022-10-08
CB-394: Do not allow CC BY-NC-SA license for future reviews

anshg1214:
CB-394

This PR:
1. Disables CC BY-NC-SA license for future reviews.
2. Upgrades to CC BY-SA 4.0 license. So all new reviews will be published under CC BY-SA 4.0 license.

### DIFF
--- a/critiquebrainz/frontend/forms/review.py
+++ b/critiquebrainz/frontend/forms/review.py
@@ -32,18 +32,11 @@ class ReviewEditForm(FlaskForm):
         StateAndLength(min=MIN_REVIEW_LENGTH, max=MAX_REVIEW_LENGTH,
                        message=lazy_gettext("Text length needs to be between %(min)d and %(max)d characters.",
                                             min=MIN_REVIEW_LENGTH, max=MAX_REVIEW_LENGTH))])
-    license_choice = RadioField(
-        choices=[
-            ('CC BY-SA 3.0', lazy_gettext('Allow commercial use of this review(<a href="https://creativecommons.org/licenses/by-sa/3.0/" target="_blank">CC BY-SA 3.0 license</a>)')),  # noqa: E501
-            ('CC BY-NC-SA 3.0', lazy_gettext('Do not allow commercial use of this review, unless approved by MetaBrainz Foundation (<a href="https://creativecommons.org/licenses/by-nc-sa/3.0/" target="_blank">CC BY-NC-SA 3.0 license</a>)')),  # noqa: E501
-        ],
-        validators=[validators.InputRequired(message=lazy_gettext("You need to choose a license"))])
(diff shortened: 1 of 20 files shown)


critiquebrainz/frontend/forms/review.py
```diff
@@ -32,18 +32,11 @@ class ReviewEditForm(FlaskForm):
...
-            ('CC BY-NC-SA 3.0', lazy_gettext('Do not allow commercial use of this review, unless approved by MetaBrainz Foundation (<a href="https://creativecommons.org/licenses/by-nc-sa/3.0/" target="_blank">CC BY-NC-SA 3.0 license</a>)')),  # noqa: E501
-        ],
-        validators=[validators.InputRequired(message=lazy_gettext("You need to choose a license"))])
     remember_license = BooleanField(lazy_gettext("Remember this license choice for further preference"))
     language = SelectField(lazy_gettext("You need to accept the license agreement!"), choices=languages)
     rating = IntegerField(lazy_gettext("Rating"), widget=Input(input_type='number'), validators=[validators.Optional()])
 
-    def __init__(self, default_license_id='CC BY-SA 3.0', default_language='en', **kwargs):
-        kwargs.setdefault('license_choice', default_license_id)
+    def __init__(self, default_license_id='CC BY-SA 4.0', default_language='en', **kwargs):
```
amCap1712, line 39: Since license id is gone, is default_license_id still needed?

anshg1214: Ahh, I missed this. We don't need it now. I'll fix it.

critiquebrainz/frontend/views/review.py
```diff
@@ -382,9 +378,6 @@ def edit(id):
         "entity_type": review["entity_type"],
         "review": review,
     }
-    if not review["is_draft"]:
```
amCap1712, line 385: IIUC, this PR currently doesn't show users the option to upgrade from CC SA v3 to CC SA v4. I don't remember whether we decided the move should be opt-in or mandatory but in any case a dialog should be shown to the user.  Also, even when doing that need to confirm with @alastair whether CC NC can be upgraded to CC SA or not and accordingly a check may be needed.

Till that is implemented in this or a future PR, we should keep this if check.

amCap1712 (commented):
Looks good but need to test on beta before merging.

critiquebrainz/frontend/views/review.py
alastair: We decided to not worry about asking to upgrade. We'll leave old ones as the current license. If we find that people are asking to upgrade their licenses, the we can add the feature

MonkeyDo:
@anshg1214 @amCap1712 Is this something we should push forward?

PR #476, opened 2022-10-24
Use confidential=False when showing data in template

alastair:
A user's musicbrainz user, or user_ref isn't private data, and show should always show when rendering a user object.
For security, don't pass email address to the template


amCap1712 (approved):
LGTM, Thanks!

PR #482, opened 2023-03-03
Update dependencies to resolve vulnerabilities

amCap1712:


amCap1712:
The failing test fails in the master as well so probably something changed with the data being imported. For later investigation.

PR #487, opened 2023-05-11
Fix failing tests and Migrate to Node 20

anshg1214:
This PR:
1. Fixes Failing tests
2. Upgrades to Node 20
3. Fixes the node `SCRIPT DEPRECATION WARNING` https://github.com/metabrainz/critiquebrainz/actions/runs/6082007492/job/16498874717#step:8:758

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,9 +29,12 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # Node
-RUN curl -sL https://deb.nodesource.com/setup_16.x | bash - \
-   && apt-get install -y nodejs \
-   && rm -rf /var/lib/apt/lists/*
+RUN mkdir -p /etc/apt/keyrings
+RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
+ENV NODE_MAJOR=20
+RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list
+RUN apt-get update \
+    && apt-get install nodejs -y
 
 RUN pip install --upgrade pip==21.0.1
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,12 +29,13 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # Node
-RUN mkdir -p /etc/apt/keyrings
-RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
-ENV NODE_MAJOR=20
-RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list
-RUN apt-get update \
-    && apt-get install nodejs -y
+ARG NODE_MAJOR=20
+RUN mkdir -p /etc/apt/keyrings \
+    && curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg \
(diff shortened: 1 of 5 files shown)


critiquebrainz/frontend/external/bookbrainz_db/test/literary_work_test.py
```diff
@@ -33,4 +33,4 @@ def test_fetch_edition_groups_for_works(self):
         self.assertEqual(edition_group_bbids_1[0], "02ae4cfc-6412-4693-93b1-e24dce5e31f9")
 
         edition_group_bbids_2 = literary_work.fetch_edition_groups_for_works(self.bbid3)
-        self.assertEqual(len(edition_group_bbids_2), 0)
+        self.assertEqual(len(edition_group_bbids_2), 1)
```
yvanzo, line 36: :thinking: It seems that this test was passing at the time it has originally been added at https://github.com/metabrainz/critiquebrainz/pull/459/commits/5faec6ba8df30c9b091f475dab211677fb6ba718#diff-f5a1981beb9671ab22e01b3d6fcd412420ce491dfc52668cbb777ef29d1c75b0

1. It is not clear why it broke since then.
2. It seems that changing this value to `1` makes it a quasi-duplicate of the first assertion on `len(edition_group_bbids_1)`, so the test might no longer be covering the case it was supposed to cover in the first place.

yvanzo (commented):
It makes the tests pass but for the second fix it is not clear why it is needed.

critiquebrainz/frontend/external/bookbrainz_db/test/literary_work_test.py
anshg1214: For testing functionalities for depending on BookBrainz's Database, since we don't have any test dump for BB. So we use the latest BB Dump. Since this data has changed in the dump, therefore this test failed.

yvanzo: Thank you, it answers to the 1st point. For the 2nd point, I suggest to replace the current work with [a work that has currently no edition](https://bookbrainz.org/work/14b910e7-c003-491b-923c-eeadd272d29b) so as to keep testing that this function returns `0` in such case.

`self.bbid4 = "14b910e7-c003-491b-923c-eeadd272d29b"`

yvanzo: Thanks for addressing this in https://github.com/metabrainz/critiquebrainz/pull/487/commits/bee002dec3bacb9efe86c66316ed81ebea252326

Dockerfile
```diff
@@ -29,9 +29,12 @@ RUN apt-get update \
...
 # Node
-RUN curl -sL https://deb.nodesource.com/setup_16.x | bash - \
-   && apt-get install -y nodejs \
-   && rm -rf /var/lib/apt/lists/*
+RUN mkdir -p /etc/apt/keyrings
+RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
+ENV NODE_MAJOR=20
+RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list
+RUN apt-get update \
+    && apt-get install nodejs -y
```
yvanzo, line 37: 1. Pass `NODE_MAJOR` as a build argument with `ARG`, it is not need in container environment.
2. Keep those commands in a single `RUN` instruction as having intermediate images seems to be superfluous otherwise.
3. Keep removing the package lists as these are not needed afterwards.

```suggestion
ARG NODE_MAJOR=20
RUN mkdir -p /etc/apt/keyrings \
    && curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg \
    && echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list \
    && apt-get update \
    && apt-get install -y nodejs \
    && rm -rf /var/lib/apt/lists/*
```

yvanzo (commented):
Thanks, I didn’t even notice that CB was using Node.js. It could have been a separate pull request but here we go. I just made 3 suggestions.

Please also rebase on `master` to avoid unneeded merge commits.

yvanzo (approved):
Thank you @anshg1214!

yvanzo:
Quickly tested in beta too.

PR #489, opened 2023-08-30
CB-459: Add LLM guidelines

Aerozol:
Updated the guideline section to include community agreed upon LLM/’AI’ usage guidelines.

### DIFF
--- a/critiquebrainz/frontend/templates/index/about.html
+++ b/critiquebrainz/frontend/templates/index/about.html
@@ -43,7 +43,7 @@
 </p>
 <p>
   {{ _('Your submissions must be your original work. Do not submit content that you do not hold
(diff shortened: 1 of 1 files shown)


yvanzo (approved):
Technically sound to be merged. Might it be worth keeping “plagiarism”?

Aerozol:
Plagiarism has been clarified to ‘[do not submit content that] is not your own’. Does that seem okay @yvanzo?
I had a play with putting ‘plagiarism’ back in as well, but couldn’t find an elegant way to do it.

yvanzo:
Maybe by putting it back in parenthesis after the replacement text “not your own” to make it clear that it is in the broadest meaning?

PR #490, opened 2023-08-31
Fix single quotes

yvanzo:
# Problem

An ISE (500) was returned instead of the page “About” with the pull requests #488 and #489 not released yet.

# Solution

* Replace curved single quotes `‘’` that were used instead of straight one `'` as string delimiter

# Extra

* Replace straight single quotes `'` that were used instead of typographic apostrophe `’` in strings
* Replace straight single quotes `'` that were used instead of curved quotation marks `“”` in strings

### DIFF
--- a/critiquebrainz/db/comment.py
+++ b/critiquebrainz/db/comment.py
@@ -102,7 +102,7 @@ def get_by_id(comment_id):
 
         comment = result.mappings().first()
         if not comment:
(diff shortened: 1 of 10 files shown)


yvanzo:
Manually tested in `beta` the pages that have been changed only.
CI test failures are unrelated, there is another pull request opened about those.

PR #495, opened 2023-10-20
Update CI unit tests results workflow

MonkeyDo:
I noticed this test workflow has been consistently failing* even thought the testing step succeeds. The action that is supposed to publish the test results is failing due to permissions issues.

I'm now adding  the required permissions to allow the workflow to write checks and comments on pull requests, as well as updating to a newer version of the action.

* see failures like so: https://github.com/metabrainz/critiquebrainz/actions/runs/6559728119/job/17815889047?pr=494#step:9:31


amCap1712 (approved):
I had not seen the PR earlier and removed the failing step altogether because we didn't use it much anyway. LGTM, Thanks.

PR #502, opened 2023-11-02
Change socket to http in uwsgi.ini

MonkeyDo:
We are trying to resolve issues with uWSGI running in the CB docker containers, which fail with the error `bind(): Address already in use [core/socket.c line 769]`
See this discussion or some background: https://chatlogs.metabrainz.org/libera/metabrainz/2023-11-02/?msg=5243055&page=2

This mimics how we configure the artwork redirect container: https://github.com/metabrainz/artwork-redirect/blob/c632ecfda1c5d5037e20a341f992b35ea7cc5ab8/docker/prod/uwsgi.ini#L3

### DIFF
--- a/docker/uwsgi/uwsgi.ini
+++ b/docker/uwsgi/uwsgi.ini
@@ -1,6 +1,6 @@
 [uwsgi]
 master = true
-http = 0.0.0.0:13032
+socket = 0.0.0.0:13032
 module = manage
 callable = application
 chdir = /code/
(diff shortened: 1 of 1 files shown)


zas (approved):

MonkeyDo:
I am building a docker image from this branch so i can deploy it to the beta CB container.
If everything is working fine and in the coming days/weeks the container is still stable while the untouched prod container fails, then we can merge this PR

zas (changes requested):
Actually this doesn't work without changes to gateways config.
Plus it might not be the actual issue.

amCap1712 (approved):

MonkeyDo:
Thanks @amCap1712 !

PR #510, opened 2024-04-26
Add support for MeB.org OAuth provider

amCap1712:
Use the newly developed MeB.org OAuth provider for handling login. Also, add support for writing reviews, voting and retrieving profile if the appropriate scopes are granted to tokens issued by the MeB.org provider. To distinguish the origin of the tokens, the fact that MeB.org issued tokens start with `meba_` can be used.


amCap1712:
Once BB and LB have migrated to using new tokens exclusively, the old OAuth provider in CB can be removed.

PR #515, opened 2024-06-20
Update link to chatrooms in footer

MonkeyDo:
Point to the MB wiki page for all things communications
image


MonkeyDo:
@phw could I have your thumbs up here too please?
I modified the link to point to MB docs page instead of wiki, as in the LB PR.

phw (approved):

PR #534, opened 2026-03-11
fix(db): Fix multiple bugs in list_comments() filter and count logic

invo-coder19:
1. Typo: 'commend.is_hidden' was used instead of 'comment.is_hidden', causing the SQL filter to always fail with an 'unknown table' error in strict SQL mode, or silently match nothing.

2. Inverted inc_hidden logic: The filter was only applied when inc_hidden=True (i.e., the caller explicitly wants hidden comments), but the intent of the parameter is the opposite — hidden comments should be EXCLUDED by default and only included when inc_hidden=True. This caused hidden comments to leak into all default queries.

3. Wrong total count: The function returned len(rows) which is capped at the 'limit' value (default 20). Callers using the count for pagination would compute incorrect page numbers. A separate COUNT(DISTINCT comment.id) query is now run with the same WHERE clause to return the true total.

### DIFF
--- a/critiquebrainz/db/comment.py
+++ b/critiquebrainz/db/comment.py
@@ -135,7 +135,7 @@ def list_comments(*, review_id=None, user_id=None, limit=20, offset=0, inc_hidde
     Returns:
         (comments, comment_count)
         where comments is list of dicts, each dict representing a comment
-              comment_count is an integer representing number of comments returned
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
Hi! Thanks for opening this PR and fixing those issues, the changes look good.
I took the liberty of adding some regression tests so issues on that front get caught in the future.

PR #542, opened 2026-07-22
Migrate OAuth to MetaBrainz.org

MonkeyDo:
Adds authentication using OAuth2 on MetaBrainz instead of MusicBrainz.
Adds webhooks for user crreation/update/deletion

See also https://github.com/metabrainz/listenbrainz-server/pull/3395 for reference


MonkeyDo:
@amCap1712 As far as I can tell, the webhooks were already using the correct props names in the payload, contrarily to LB (username/name mismatch, fixed in https://github.com/metabrainz/listenbrainz-server/pull/3395/changes/d5ef95a2ea11ea009491c1b4d6a8f3f07903c407)

amCap1712 (approved):

PR #555, opened 2026-09-10
Improve BookBrainz entity lookup query performance

mwiencek:
Depends on https://github.com/metabrainz/bookbrainz-site/pull/1370

Yesterday I noticed an influx of these queries due to an increase in CB traffic. Averaging about 200 ms each.

These improvements were largely LLM-assisted outside of the code comments and commit messages, as it probably would have taken days for me to learn the BookBrainz schema and debug the PostgreSQL planner.

I did verify that the old and new queries are equivalent (by removing the parameters and using `EXCEPT ALL` in both directions) and used `print` debugging to check the cache issue in the second commit. I also checked the `EXPLAIN` output of the new queries to make sure they actually use all of the new indexes in the BookBrainz PR.

These now execute in 1-2 ms locally.


mwiencek:
Not sure what's up with the test failures, but the same ones occur in the recent dependabot PRs, so I guess unrelated.

MonkeyDo (approved):
Queries work fine, and in combination with the indices in the BB PR I can confirm it speeds up the queries tremendously

MonkeyDo:
> Not sure what's up with the test failures, but the same ones occur in the recent dependabot PRs, so I guess unrelated.

Unrelated indeed.
Failing tests are related to https://github.com/metabrainz/bookbrainz-site/commit/583cdd356192feb2effe415b1c8f09846a898d01 where I replaced a UUID extension with native postgres `gen_random_uuid`, which is only available from PG v14 onwards. Forgot that CB runs PG v12, will open a PR to fix that.

Edit: Opened in #556 with CI tests passing

PR #556, opened 2026-09-14
Upgrade to Postgres 18

MonkeyDo:
We had mismatched Postgres versions in dev and in CI, plus uses of the uuid-ossp extension for UUID that is not needed anymore with Postgres v14+

Upgrades PG versions to v18, removes uuid-ossp extension and replaces its uuid_generate_v4 with native gen_random_uuid.

See same work done on BookBrainz in
https://github.com/metabrainz/bookbrainz-site/commit/583cdd356192feb2effe415b1c8f09846a898d01

After this change in BB, CB tests were failing with:
> ERROR:  function gen_random_uuid() does not exist
LINE 2:     bbid uuid DEFAULT gen_random_uuid() NOT NULL,
                              ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
ERROR:  relation "bookbrainz.entity" does not exist
ERROR:  relation "bookbrainz.entity" does not exist
LINE 30:      LEFT JOIN bookbrainz.entity e ON ((e.bbid = ar.bbid)))
                        ^
ERROR:  relation "bookbrainz.author" does not exist

Confirmed running tests.sh does not fail with the uuid-related errors. Passing CI suite will be the second confirmation

I will prepare a follow-up PR that creates a minimal BookBrainz database for use with tests, instead of downloading the database dumps as we currently do, to avoid this sort of issues in the future.


mwiencek (approved):
Tests pass and the changes look sound. 👍

